PR #154, opened 2015-01-06
Add ability to output data which is sorted on an input column

vaibhavsehgal1:


core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -72,13 +83,22 @@
...
+
     @Inject
     @Provided
     UserPreferences userPreferences;
 
+    private File _tempFile;
+
+    @Initialize
+    public void initTempFile() throws Exception {
+        _tempFile = File.createTempFile("csv_file_analyzer", ".csv");
```
kaspersorensen: I guess that if "columnToBeSortedOn" is null (sorting is turned off), we should not write to a temp file. We should write to the final file directly. So maybe let's rename the "_tempFile" variable and call it something a la "_targetFile". And initialize it according to the null check.

core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -107,16 +127,94 @@ public OutputWriter createOutputWriter() {
...
             headers[i] = columns[i].getName();
         }
-        return CsvOutputWriterFactory.getWriter(file.getPath(), headers, separatorChar, quoteChar, escapeChar,
-                includeHeader, columns);
+        return CsvOutputWriterFactory.getWriter(_tempFile.getPath(), headers, separatorChar, quoteChar, escapeChar, includeHeader, columns);
     }
 
     @Override
     protected WriteDataResult getResultInternal(int rowCount) {
+        final CsvConfiguration csvConfiguration = new CsvConfiguration(CsvConfiguration.DEFAULT_COLUMN_NAME_LINE, FileHelper.DEFAULT_ENCODING,
```
kaspersorensen: Should also take into account whether or not we need to do sorting. If not, a lot of this can be left out.

core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -107,16 +127,94 @@ public OutputWriter createOutputWriter() {
...
+                separatorChar, quoteChar, escapeChar, false, true);
+        final CsvDataContext tempDataContext = new CsvDataContext(_tempFile, csvConfiguration);
+        final Table table = tempDataContext.getDefaultSchema().getTable(0);
+
+        final Comparator<? super Row> comparator = new Comparator<Row>() {
+            @SuppressWarnings("unchecked")
+            @Override
+            public int compare(Row row1, Row row2) {
+                Comparable<Object> value1 = (Comparable<Object>) row1.getValue(columnToBeSortedOn.getPhysicalColumn().getColumnNumber());
+                Comparable<Object> value2 = (Comparable<Object>) row2.getValue(columnToBeSortedOn.getPhysicalColumn().getColumnNumber());
```
kaspersorensen: calling Row.getValue with the columnToBeSortedOn is not going to work... Maybe it has worked by coincidence, but it's certainly not a given. The reason is that the rows may not even contain this column (that's dependent on the @Configured InputColumn[] that drives the inserts) and it's index in the file may not be the same as it's original column number.

I'd suggest figuring out what the index is at the time of inserting the records into the temp file. Maybe first check if columnToBeSortedOn is included in the array of InputColumns that are being written. If not, it needs to be added somehow. If it is, then keep it's index so that it can be used here.

kaspersorensen:
Finished with my review, please see the inline/code comments.

vaibhavsehgal1:
Incorporated review comments.

core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -103,20 +135,135 @@ public String getSuggestedLabel() {
...
+            final CsvDataContext tempDataContext = new CsvDataContext(_targetFile, csvConfiguration);
+            final Table table = tempDataContext.getDefaultSchema().getTable(0);
+
+            final Comparator<? super Row> comparator = new Comparator<Row>() {
+                @SuppressWarnings("unchecked")
+                @Override 
+                public int compare(Row row1, Row row2) {
+                    Comparable<Object> value1 = (Comparable<Object>) row1.getValue(indexOfColumnToBeSortedOn);
+                    Comparable<Object> value2 = (Comparable<Object>) row2.getValue(indexOfColumnToBeSortedOn);
+                    int comparableResult = value1.compareTo(value2) ;
```
kaspersorensen: Very minor: There's a risk of NPE here. Might want to use CompareUtils to avoid that.

core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -103,20 +135,135 @@ public String getSuggestedLabel() {
...
+            if(indexOfColumnToBeSortedOn == -1) {
+                this.isColumnToBeSortedOnPresentInInput  = false ;
+                indexOfColumnToBeSortedOn = columns.length ;
+                headers.add(columnToBeSortedOn.getName()) ;
+                InputColumn<?>[] newColumns = new InputColumn<?>[columns.length + 1] ;
+                for(int i = 0; i < columns.length; i++){
+                    newColumns[i] = columns[i] ;
+                }
+                newColumns[columns.length]  = columnToBeSortedOn ;
+                columns = newColumns ;
```
kaspersorensen: Very minor: The "columns" field should ideally not change. It is set by the job configuration and should be considered immutable.

core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -103,20 +135,135 @@ public String getSuggestedLabel() {
...
+                        return comparableResult ;
+                    } else {
+                        // The values of the data at the row, and column to be sorted on are 
+                        // exactly the same. Now look at other values of all the columns to 
+                        // find if the two rows are same.
+                        int numberOfSelectItems = row1.getSelectItems().length ;
+                        for(int i = 0; i < numberOfSelectItems; i++) {
+                            Comparable<Object> rowValue1 = (Comparable<Object>) row1.getValue(i);
+                            Comparable<Object> rowValue2 = (Comparable<Object>) row2.getValue(i);
+                            if(rowValue1.compareTo(rowValue2) == 0) {
```
kaspersorensen: Same NPE risk here

core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -103,20 +135,135 @@ public String getSuggestedLabel() {
...
+                    final String line = csvWriter.buildLine(columnNames);
+                    writer.write(line);
+                    writer.append('\n');
+                }
+
+                @Override
+                protected void writeRow(Writer writer, Row row, int count) throws IOException {
+                    for(int i = 0; i < count; i++) {
+                        List<Object> valuesList = new ArrayList<Object>(Arrays.asList(row.getValues())) ;
+                            if(!isColumnToBeSortedOnPresentInInput) {
```
kaspersorensen: Indentation looks strange here.

core/src/main/java/org/eobjects/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -103,20 +135,135 @@ public String getSuggestedLabel() {
...
+                        // exactly the same. Now look at other values of all the columns to 
+                        // find if the two rows are same.
+                        int numberOfSelectItems = row1.getSelectItems().length ;
+                        for(int i = 0; i < numberOfSelectItems; i++) {
+                            Comparable<Object> rowValue1 = (Comparable<Object>) row1.getValue(i);
+                            Comparable<Object> rowValue2 = (Comparable<Object>) row2.getValue(i);
+                            if(rowValue1.compareTo(rowValue2) == 0) {
+                                continue;
+                            } else {
+                                return rowValue1.compareTo(rowValue2) ;
```
kaspersorensen: And same NPE risk here :)

kaspersorensen:
OK did another review. Looks very close to done now. Can we also add a unittest that does sorting on a small file in src/test/resources or so?

vaibhavsehgal1:
Wrote code to cater to new review comments. I can write a unit test, but there are no unit test for CreateCsvFileAnalyzer as a whole, so I will need help on it.

kaspersorensen:
OK I will merge and then do a manual smoke test ... a unit test is still quite valuable IMO but we can leave it for a consecutive contribution. As far as "how" - simply make a normal junit class that invokes the methods on the class. For the InputColumn and InputRow objects we have mock implementations that are useful for testing, e.g. MockInputRow and MockInputColumn.

PR #171, opened 2015-01-22
Fixes #166

### LINKED ISSUE #166: JobGraph: Show max 1 dialog per component
Do not allow multiple dialogs open for configuring the same component. If the user double-clicks or selected "Configure" in the right-click menu, and a dialog is already open, just put it to focus.

kaspersorensen:
This fixes #166


LosD:
Nice and simple. Merging.

PR #172, opened 2015-01-23
Feature/140 cassandra dialogs

### LINKED ISSUE #140: Add datastore dialogs for Cassandra and ElasticSearch datastores

tomaszguzialek:
Closes #140 

ElasticSearch icon might be outdated. It is not used on their official website, where they have only the textual logo. However, this one is not suitable for using here.


kaspersorensen:
I am missing some nicer icons here:

image

kaspersorensen:
I can save my ElasticSearch datastore without providing an index name:

image

I think similarly I can save Cassandra without keyspace name and so on.

kaspersorensen:
Small remark, but in this list ...

image

...I'd like this ordering:

* MongoDB
* CouchDB
* ElasticSearch
* Cassandra
* HBase

kaspersorensen:
Looks quite good. I made some cosmetic adjustments and reported a few issues above. Can you please take a look at those?

I think it easily reveals some nasty exceptions coming from MetaModel when trying to connect to e.g. an unexisting ES index. But I'll report those on the MM JIRA instead.

tomaszguzialek:
I introduced validating the form in the AbstractDatastoreDialog class. Looking forward to feedback for this code.

Icons replaced in existing datastores sorted according to the wish in adding datastore panel.

desktop/ui/src/main/java/org/datacleaner/windows/AbstractDatastoreDialog.java
```diff
@@ -49,6 +52,27 @@
     private final JButton _saveButton;
     private final JButton _cancelButton;
     private final UserPreferences _userPreferences;
+    
+    protected final JXTextField _datastoreNameTextField;
```
kaspersorensen: I think this is too broad. Keep in mind that AbstractDatastoreDialog is a super class of ALL datastore dialogs. So if you want to centralize the datastore name widget (which is doable) then it needs adaption in 10-15 other subclasses too :-)

tomaszguzialek: That's true. Do you think it is a good idea to increase the scope and rework all the subclasses to use inherited name field and its validity check? I can see the file-based dialogs have nice checks, but the rest (Salesforce, SugarCRM, Mongo etc.) do not block the Save button and show a dialog after clicking it.

kaspersorensen:
Don't worry about the Travis CI build failure - that unittest is somehow sometimes failing, but it's unrelated.

desktop/ui/src/main/java/org/datacleaner/windows/AbstractDatastoreDialog.java
kaspersorensen: I will let that be up to you. I can live with both ways of doing it ("quick'n'dirty" or "more work") :-) Up to you!

tomaszguzialek:
I moved quite a lot of code from AbstractFileBasedDatastoreDialog to AbstractDatastoreDialog. Now file-based and others should be more consistent. Also, others have status bar now and improved error handling. 

Looking forward to code feedback as I wanted to make it nice and clean, but I have a feeling it could be even better ;).

kaspersorensen:
I think this looks very good. I have two minor remarks - they can be handled directly in the branch or after merging (meaning that I like the PR and accept it as is, but have another wish :-))

 * There's still a handful of datastore dialogs that don't have the status bar etc. We might want to go the last mile to standardize how these dialogs work and then e.g. make AbstractDatastoreDialog.getDialogContent() finalso that it cannot be overridden (but maybe allow one or more fine-grained hooks to decorate the content).
 * I think we should not call it "ElasticSearch database", but rather "ElasticSearch index" (banner/title of dialog as well as tooltip on the dialog's icon).

tomaszguzialek:
ElasticSearch database renamed to ElasticSearch index. The rest of the dialogs have been reworked too. The only thing left not coherent is JDBC dialog - validation of datastore name and driver field is in place, but no validation for the tabbed pane with generic and specific settings. Because of that, the status bar is not rendered in this dialog not to mislead the user with "Datastore ready" status set too eagerly.

kaspersorensen:
If I edit an existing CSV datastore I get this exception:

```
Caused by: java.lang.NullPointerException
	at org.datacleaner.windows.CsvDatastoreDialog.validateForm(CsvDatastoreDialog.java:297)
	at org.datacleaner.windows.AbstractFileBasedDatastoreDialog.validateAndUpdate(AbstractFileBasedDatastoreDialog.java:188)
	at org.datacleaner.windows.AbstractDatastoreDialog$3.onChange(AbstractDatastoreDialog.java:115)
```

Looks like in some cases we might not have created all class members before calling validate. I guess you should ensure that in all cases we set the class members before any such initialization takes place.

kaspersorensen:
btw - I like the help icon in SFDC datastore :-)

tomaszguzialek:
NPE fixed.

PR #173, opened 2015-01-23
Makes a popup button component

LosD:
For now, only used for toolbar menu button, but could potentially be used for all
our popups.

At some point, we probably want to use a ToggleButtonUI for this, but
that needs an DarkUI override, so that'll have to wait.


kaspersorensen:
The code looks good to me, but the new file is missing a license header and thus breaks the build! :-)

LosD:
Whoopsie...
Done. Waiting for Travis to accept :)

PR #183, opened 2015-01-23
Feature/138 analyzer result future renderer

### LINKED ISSUE #138: Add loading panel for AnalyzerResultFuture in DC desktop
AnalyzerResultFuture was introduced as a way for analyzers to return a result that is still processing.

See https://github.com/datacleaner/AnalyzerBeans/issues/64 for details.

In the UI of DataCleaner we should have a way to treat this. We should show a loading indicator in the result panel and start listening for the underlying result to be ready. Once it's ready we should re-render the result panel and put the underlying result's rendered swing component in there.

tomaszguzialek:
Closes #138

Works with the main method in AnalyzerResultFutureSwingRenderer, but not sure if it needs to be plugged in somewhere else to work for all the analyzers.


components/basic-analyzers/src/main/java/org/datacleaner/beans/MockAnalyzer.java
```diff
@@ -0,0 +1,72 @@
...
+import org.datacleaner.api.InputColumn;
+import org.datacleaner.api.InputRow;
+import org.datacleaner.result.AnalyzerResultFuture;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Named("MockAnalyzer")
+@Description("For testing purposes. Sleeps for 5 seconds.")
+@Concurrent(true)
+public class MockAnalyzer implements Analyzer<AnalyzerResultFuture<AnalyzerResult>> {
```
kaspersorensen: Actually we don't want this class in src/main/java ... we could accept it in src/test/java but otherwise it's gonna enter the main distro of DC and that doesn't work! :-)

Note: Your main method would then also need to be in src/test/java

components/basic-analyzers/src/main/java/org/datacleaner/beans/MockAnalyzerResult.java
```diff
@@ -0,0 +1,34 @@
...
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.beans;
+
+import org.datacleaner.api.AnalyzerResult;
+import org.datacleaner.api.Metric;
+
+public class MockAnalyzerResult implements AnalyzerResult {
```
kaspersorensen: Same point about this class's location - should be test code.

desktop/ui/src/main/java/org/datacleaner/widgets/result/AnalyzerResultFutureSwingRenderer.java
```diff
@@ -0,0 +1,122 @@
...
+        renderable.addListener(new AnalyzerResultFuture.Listener<AnalyzerResult>() {
+
+            @Override
+            public void onSuccess(AnalyzerResult result) {
+                Renderer<? super AnalyzerResult, ? extends JComponent> renderer = _rendererFactory.getRenderer(result, SwingRenderingFormat.class);
+                if (renderer != null) {
+                    JComponent jComponent = renderer.render(result);
+                    resultPanel.add(jComponent);
+                } else {
+                    logger.error("No renderer found for: " + result);
```
kaspersorensen: I guess the user would not see this ... So what should he see? I guess we should have the same behaviour as when ResultWindow cannot find a renderer.

desktop/ui/src/main/java/org/datacleaner/widgets/result/AnalyzerResultFutureSwingRenderer.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+                
+                resultPanel.remove(busyLabel);
+                resultPanel.updateUI();
+                
+            }
+
+            @Override
+            public void onError(RuntimeException error) {
+                logger.error("Error occured while retrieving AnalyzerResult from AnalyzerResultFuture" + error);
```
kaspersorensen: And here also.

desktop/ui/src/main/java/org/datacleaner/widgets/result/MockAnalyzerResultSwingRenderer.java
```diff
@@ -0,0 +1,52 @@
...
+import org.datacleaner.api.Renderer;
+import org.datacleaner.api.RendererBean;
+import org.datacleaner.api.RendererPrecedence;
+import org.datacleaner.beans.MockAnalyzerResult;
+import org.datacleaner.panels.DCPanel;
+import org.datacleaner.result.renderer.SwingRenderingFormat;
+import org.datacleaner.widgets.DCLabel;
+
+@RendererBean(SwingRenderingFormat.class)
+public class MockAnalyzerResultSwingRenderer implements Renderer<MockAnalyzerResult, JComponent> {
```
kaspersorensen: Also move this to test code

tomaszguzialek:
Improved error handling for the end user and moved classes to test scope.

desktop/ui/src/main/java/org/datacleaner/widgets/result/AnalyzerResultFutureSwingRenderer.java
```diff
@@ -0,0 +1,90 @@
...
+    RendererFactory _rendererFactory;
+    
+    @Override
+    public RendererPrecedence getPrecedence(AnalyzerResultFuture<? extends AnalyzerResult> renderable) {
+        return RendererPrecedence.LOW;
+    }
+
+    @Override
+    public JComponent render(AnalyzerResultFuture<? extends AnalyzerResult> renderable) {
+        final JXBusyLabel busyLabel = new JXBusyLabel();
```
kaspersorensen: We are normally using LoadingIcon, so let's continue that for now. But we could consider changing LoadingIcon to extend JXBusyLabel if it it better/cooler/whatever.

desktop/ui/src/main/java/org/datacleaner/widgets/result/AnalyzerResultFutureSwingRenderer.java
```diff
@@ -0,0 +1,90 @@
...
+                    logger.debug("renderer.render({})", result);
+                    final JComponent component = renderer.render(result);
+                    resultPanel.add(component);
+                } else {
+                    final String message = "No renderer found for result type " + result.getClass().getName();
+                    logger.error(message);
+                    throw new IllegalStateException(message);
+                }
+                
+                resultPanel.remove(busyLabel);
```
kaspersorensen: Put this in finally block ... if an error is thrown we don't want to see the spinner any more.

kaspersorensen:
Except for the two points above, it looks ready to merge.

tomaszguzialek:
Fixed the try-finally block and replaced JXBusyLabel with LoadingIcon. Gonna merge when Travis builds it successfully.

kaspersorensen:
Travis is happy :-) Merging ...

PR #187, opened 2015-01-23
Keeping amount of dialogs reasonable.

LosD:
This will close dialogs along with nodes and jobs, and adds a way to close all dialogs in the window menu.

It also extends the previous "pull existing component configuration dialog to front" functionality to table configuration dialogs.


LosD:
FYI: I've been a little bit bad: I found a bug in the new popup button method while merging (the playing around with dialogs make it very obvious, and I fixed it as part of doing the merge.

kaspersorensen:
I like it.

PR #188, opened 2015-01-23
Do not refresh  panel when checking size.

### LINKED ISSUE #186: JobGraph.paint() runs constantly in new graph UI
In my machine (2.9 GHz core i7) DataCleaner uses ~15-20% CPU constantly when the graph UI is shown, even with no components are anything added. The issue seems to be somewhere in JobGraph.paint(), which, according to JVisualVM is active 60% of the time that DC is running.

LosD:
The code used getPanel(), which refreshed and redrew the entrie panel.
As the result was only used for sizing, there should be no need.

Fixes #186


kaspersorensen:
Great, what a good catch.

PR #194, opened 2015-01-26
#169 added an option for overwriting the previous file

### LINKED ISSUE #169: Improve error message if "Create Excel spreadsheet" or "Create CSV file" is in append-mode to file without necesary fields
Improve error message if "Create Excel spreadsheet" or "Create CSV file" is in append-mode to file without necesary fields.

No such column in table: ADDRESS_LINE_1, available columns are: [Column[name=FULL_ADDRESS,columnNumber=0,type=STRING,nullable=true,nativeType=null,columnSize=null], Column[name=UNIQUE_ID (lookup),columnNumber=1,type=STRING,nullable=true,nativeType=null,columnSize=null]]
 
When applying one of these writers to an existing file, it will be attempted to append instead of create new file/table. Thus when performing the inserts, MetaModel will report errors like these:

```
java.lang.IllegalArgumentException: No such column in table: ADDRESS_LINE_1, available columns are: [Column[name=FULL_ADDRESS,columnNumber=0,type=STRING,nullable=true,nativeType=null,columnSize=null], Column[name=UNIQUE_ID (lookup),columnNumber=1,type=STRING,nullable=true,nativeType=null,columnSize=null]]
     at org.apache.metamodel.data.AbstractRowBuilder.value(AbstractRowBuilder.java:133)
     at org.apache.metamodel.data.AbstractRowBuilder.value(AbstractRowBuilder.java:119)
     at org.eobjects.datacleaner.output.AbstractMetaModelOutputWriter$1.run(AbstractMetaModelOutputWriter.java:94)
     at org.apache.metamodel.excel.ExcelDataContext.executeUpdate(ExcelDataContext.java:261)
     at ...
```

Instead we should give the user a more sensible message, explaining that the file exists but the columns in it does not match.

ClaudiaPHI:
Fixes #169

### DIFF
--- a/components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
+++ b/components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
@@ -50,6 +50,7 @@ import org.datacleaner.api.HasLabelAdvice;
 import org.datacleaner.api.Initialize;
 import org.datacleaner.api.InputColumn;
 import org.datacleaner.api.Provided;
+import org.datacleaner.api.Validate;
 import org.datacleaner.beans.writers.WriteDataResult;
 import org.datacleaner.beans.writers.WriteDataResultImpl;
 import org.datacleaner.components.categories.WriteSuperCategory;
@@ -88,27 +89,44 @@ public class CreateCsvFileAnalyzer extends AbstractOutputWriterAnalyzer implemen
     boolean includeHeader = true;
 
     @Configured(order = 6, required = false)
-    InputColumn<?> columnToBeSortedOn ;
+    InputColumn<?> columnToBeSortedOn;
+
+    @Configured
+    boolean overwriteFile;
 
     @Inject
     @Provided
     UserPreferences userPreferences;
 
     private File _targetFile;
-    
-    private int indexOfColumnToBeSortedOn = -1 ;
 
-    private boolean isColumnToBeSortedOnPresentInInput = true ; 
+    private int indexOfColumnToBeSortedOn = -1;
 
+    private boolean isColumnToBeSortedOnPresentInInput = true;
 
     @Initialize
     public void initTempFile() throws Exception {
         if (_targetFile == null) {
-            if(columnToBeSortedOn != null) {
+            if (columnToBeSortedOn != null) {
                 _targetFile = File.createTempFile("csv_file_analyzer", ".csv");
             } else {
-                _targetFile = file ;
+                _targetFile = file;
             }
+        }
+    }
+
+    @Override
+    public String getSuggestedLabel() {
+        if (file == null) {
+            return null;
+        }
+        return file.getName();
+    }
+
+    @Validate
+    public void validate() {
+        if (!file.exists() && overwriteFile == true) {
+            throw new IllegalStateException("The file can not be overwritten because the file does not exist");
         }
     }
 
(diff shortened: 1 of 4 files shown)


components/writers/src/main/java/org/datacleaner/extension/output/CreateExcelSpreadsheetAnalyzer.java
```diff
@@ -71,11 +74,14 @@ public void validate() {
         if (sheetName.indexOf(".") != -1) {
             throw new IllegalStateException("Sheet name cannot contain dots (.)");
         }
+
+        if (file.exists() && !overwriteFile) {
+            throw new IllegalStateException("The file already exits and the columns selected do not match");
```
tomaszguzialek: Typo: The file already EXITS :)

I would also be more explicit in the message: "The file already exists and the columns selected do not match the existing records"

components/writers/src/main/java/org/datacleaner/extension/output/CreateExcelSpreadsheetAnalyzer.java
```diff
@@ -88,10 +94,22 @@ public void configureForTransformedData(AnalysisJobBuilder ajb, TransformerDescr
...
     @Override
     public OutputWriter createOutputWriter() {
+
+        if (file.exists() && overwriteFile) {
+            final String pathname = file.getPath();
+            final boolean delete = file.delete();
+            if (delete) {
+                file = new File(pathname);
+            } else {
+                throw new IllegalStateException("The previous file could not be deleted");
```
tomaszguzialek: Again small thing, but "the existing file" sounds better IMHO.

tomaszguzialek:
What about CSV files? I can see only Excel in the code, but the issue mentions both Excel and CSV.

ClaudiaPHI:
I need to add a test.

ClaudiaPHI:
The Csv files writing seems to work. There was a problem only with Excel files

kaspersorensen:
Well - yes and no. The CSV writing did not throw any exception, but it simply overrides any existing file without any warning of flag. Probably we should introduce the same flag in the CSV writer, so that it's consistent and does not accidentally overwrite data.

ClaudiaPHI:
Ok. I am on it.

components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -88,31 +89,48 @@
...
+    public String getSuggestedLabel() {
+        if (file == null) {
+            return null;
+        }
+        return file.getName();
+    }
+
+    @Validate
+    public void validate() {
+        if (!file.exists() && overwriteFile == true) {
```
kaspersorensen: I think we should not care about this situation (neither here nor in Excel) - if the file does not exist then we are not running any risks of overwriting existing data. So we should probably also name the configured property like that - instead of "overwriteFile" it should be "overwriteFileIfExists".

components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -129,40 +147,37 @@ public void configureForTransformedData(AnalysisJobBuilder ajb, TransformerDescr
...
-            return null;
-        }
-        return file.getName();
-    }
-
-    @Override
     public OutputWriter createOutputWriter() {
-        List<String> headers = new ArrayList<String>() ;
+        
+        if (file.exists() && overwriteFile==false){
```
kaspersorensen: This condition could be moved up to a Validate method insted of the createOutputWriter() method (which is invoked at initiatlize time, whereas validation time is much more frequent/early).

components/writers/src/main/java/org/datacleaner/extension/output/CreateExcelSpreadsheetAnalyzer.java
```diff
@@ -71,11 +74,14 @@ public void validate() {
         if (sheetName.indexOf(".") != -1) {
             throw new IllegalStateException("Sheet name cannot contain dots (.)");
         }
+
+        if (file.exists() && !overwriteFile) {
```
kaspersorensen: Well I guess this should not be done here either. But beyond that this is also a bad place to do it - the getSuggestedLabel() method does not have anything to do with validating the component and if it throws an exception here it may have totally different consequences than what you are trying to archieve!

kaspersorensen: This also means that this class is still missing a validate method which will fail in the scenario where overwriteFile is false and the file exists.

components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
ClaudiaPHI: Yeah, but validate it is not called in Save us utilities. 
image

ClaudiaPHI:
Ready for review again

kaspersorensen:
Looks good! And thanks for the hard work with all the review remarks :-)

PR #218, opened 2015-02-05
Feature/202 window size state

### LINKED ISSUE #202: Make desktop UI start up and remember maximized state
By now the window state of DataCleaner will more likely be maximized than not. But we currently do not remember this, we only remember the window size. So if the window was maximized, we restore it as non-maximized but with an almost-maximized window size. Rather we should remember if the main window was maximized and if so (or even by default) show it maximized.

ClaudiaPHI:
Closes #202 and #1533. 

The window is almost maximized. I do not think it can be better maximized(or bigger).


desktop/ui/src/main/java/org/datacleaner/util/UserPreferencesUtils.java
```diff
@@ -0,0 +1,73 @@
...
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.util;
+
+import java.awt.Dimension;
+import java.util.Map;
+
+import org.datacleaner.user.UserPreferences;
+
+public class UserPreferencesUtils {
```
kaspersorensen: Good helper class to have, but I think the class name is not very descriptive. Maybe relate it to the window size?

desktop/ui/src/main/java/org/datacleaner/util/UserPreferencesUtils.java
```diff
@@ -0,0 +1,73 @@
...
+        return new Dimension(Integer.parseInt(widthStr), Integer.parseInt(heightStr));
+    }
+
+    public void setUserPreferredSize(Dimension size) {
+        Map<String, String> properties = _userPreferences.getAdditionalProperties();
+        properties.put(getWidthPropertyKey(), "" + size.width);
+        properties.put(getHeightPropertyKey(), "" + size.height);
+        _userPreferences.save();
+    }
+
```
kaspersorensen: I'm missing a "isMaximized()" boolean

desktop/ui/src/main/java/org/datacleaner/widgets/DCPersistentSizedPanel.java
```diff
@@ -20,69 +20,33 @@
...
-		if (widthStr == null) {
-			widthStr = "" + _defaultWidth;
-		}
-		String heightStr = properties.get(getHeightPropertyKey());
-		if (heightStr == null) {
-			heightStr = "" + _defaultHeight;
-		}
+        _userPreferenceUtils = userPreferenceUtils;
+        final Dimension preferredSizeFromUserPreferences = _userPreferenceUtils.getUserPreferredSize();
+        setPreferredSize(preferredSizeFromUserPreferences);
```
kaspersorensen: Beyond just setting the size we should check if the window was maximized. If it was then we should maximize it (different from setting size).

kaspersorensen:
Done with my review. As far as I can see this PR only addresses one out of two aspects:

 - [x] Keeping size below screen size
 - [ ] Remembering the maximized state of the window

For me it is OK if we only save the maximized state on the AnalysisJobBuilderWindow since it might easily become window specific and because the DCPanel is not a window, just a panel, so it does not have a maximized state, only a size?

ClaudiaPHI:
Unfortunately cannot  make an maximized state to look nice. It renders it bigger than it should.

LosD:
How are you trying to maximize it, and when? It should be after it becomes visible, and using ```myFrame.setExtendedState(myFrame.getExtendedState() | JFrame.MAXIMIZED_BOTH);``` as far as Google tells me (which may or may not be true :))

LosD:
Actually, it seems to be your issue: "_Without setVisible(true) first, the frame is not sized correctly. I tried it both before and after setExtendedState(), and setVisible() must be first._"
http://stackoverflow.com/questions/479523/java-swing-maximize-window/479561#479561

ClaudiaPHI:
:+1: Thanks a lot Dennis. You rock.

LosD:
It worked? Cool! :)

desktop/api/src/main/java/org/datacleaner/util/UserWindowSizePreferencesUtils.java
```diff
@@ -0,0 +1,88 @@
...
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.util;
+
+import java.awt.Dimension;
+import java.util.Map;
+
+import org.datacleaner.user.UserPreferences;
+
+public class UserWindowSizePreferencesUtils {
```
kaspersorensen: Sorry to keep bothering about this point, but I still feel the name of the class is kinda weird. It is somehow referring to the UserPreferences object but I think that is a technical detail (which is revealed by the constructor argument already). We have a similar class in JobGraphPreferencesPanel. So maybe we could name this WindowSizePreferences? Then the "User" and "Utils" part is stripped, but those are also for me the least descriptive parts of the name.

desktop/api/src/main/java/org/datacleaner/util/UserWindowSizePreferencesUtils.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private final UserPreferences _userPreferences;
+    private final String _identifier;
+    private final int _defaultWidth;
+    private final int _defaultHeight;
+
+    public UserWindowSizePreferencesUtils(final UserPreferences userPreferences, final String identifier,
```
kaspersorensen: Can we add a javadoc. I don't know what "identifier" is - I imagine it's a name/ID of the window?

desktop/ui/src/main/java/org/datacleaner/util/UserPreferencesUtils.java
kaspersorensen: This class is still here? Seems redundant with the previous one.

desktop/api/src/main/java/org/datacleaner/windows/AbstractWindow.java
```diff
@@ -49,18 +49,22 @@ public AbstractWindow(WindowContext windowContext) {
         setDefaultCloseOperation(JFrame.DO_NOTHING_ON_CLOSE);
         addWindowListener(this);
         getContentPane().setBackground(WidgetUtils.BG_COLOR_BRIGHT);
+
     }
 
+    protected abstract boolean isWindowMaximized();
```
kaspersorensen: Instead of having to overwrite this in N places with "return false", why not just have a default implementation that returns false and then override it the one place where it is not?

kaspersorensen: and naming wise I think it should say "isInitiallyMaximized" because otherwise it gives the impression that it will always tell you the state of the window (which may change, but the "return false" implementation never changes).

kaspersorensen:
Done with my review :-) Looks to be good function, but naming is something to improve a bit.

desktop/api/src/main/java/org/datacleaner/windows/DataSetWindow.java
```diff
@@ -99,11 +99,6 @@ public String getWindowTitle() {
     }
 
     @Override
-    protected boolean isWindowResizable() {
```
kaspersorensen: now DataSet windows are no longer resizable? Why?

desktop/api/src/main/java/org/datacleaner/windows/AbstractWindow.java
```diff
@@ -114,11 +119,12 @@ public Dimension autoSetSize(Component content) {
         if (isCentered()) {
             centerOnScreen();
         }
-
         return preferredSize;
     }
 
-    protected abstract boolean isWindowResizable();
+    protected boolean isWindowResizable(){
```
kaspersorensen: Why this change?

desktop/api/src/main/java/org/datacleaner/util/WindowSizePreferences.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.datacleaner.user.UserPreferences;
+
+public class WindowSizePreferences {
+
+    private final UserPreferences _userPreferences;
+    private final String _identifier;
+    private final int _defaultWidth;
+    private final int _defaultHeight;
+/**
```
kaspersorensen: Needs spacing/formatting.

desktop/ui/src/main/java/org/datacleaner/util/UserPreferencesUtils.java
kaspersorensen: What is this class still doing?

kaspersorensen:
Sorry, still not there. Please check all the comments, also the one about UserPreferencesUtils which I think was missed before?

desktop/ui/src/main/java/org/datacleaner/util/UserPreferencesUtils.java
ClaudiaPHI: I really do not know. I renamed it, but I think it created a copy instead.

kaspersorensen:
From coding perspective this looks good now. But I also tried it out and I have still two wishes. Now I will try and be polite ;-)

 * By default the main window should be maximized UNLESS there are preferences that say it is not. If you delete your userpreferences.dat file the main window still shows as non-maximized.
 * If I saved the window as maximized and then click the de-maximize icon then the window size is gigantic. Could we make it so that we do not save the window size when the window is maximized - otherwise it will just save my screen resolution as the window size and that's not a practical size to have saved.

ClaudiaPHI:
Done :-)

kaspersorensen:
Cool burger :-)

PR #220, opened 2015-02-05
Feature/185 unsaved jobs

### LINKED ISSUE #185: Prompt for confirmation when closing an unsaved job
Currently DataCleaner will happily close a job that hasn't been saved, making the user lose all modifications if s/he hit the close button by accident or didn't know there was unsaved modifications.

It would be nice if DC prompted the user for confirmation in these cases, i.e. with a standard "Save? Yes/No/Cancel" dialog.

tomaszguzialek:
Fixes #185 

Could be more elegant, but that seemed the easiest. Shout if too dirty ;).


kaspersorensen:
I am not sure this is a good way. I agree that it seems easiest tho' :-) I guess another way would have to include keeping a timestamp of the latest save time and then a timestamp of the latest change (could become difficult to track)...

What do you think @LosD and @ClaudiaPHI ?

LosD:
I had more or less the same though, just with setting a simple dirty flag somewhere. The question is if we have a single place to add this, or if we need to go through all lot of places and add a listener.

kaspersorensen:
We don't have a listener type that catches all job-changing events. So I guess it would be a lot of listeners and a design that would be vulnerable to changes/additions of the job building model.

kaspersorensen:
How heavy does this function seem? I mean, when you run it on a normal laptop is there then a delay that is experienced by the user?

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -558,6 +582,33 @@ protected boolean onWindowClosing() {
         return windowClosing;
     }
 
+    private boolean isJobUnsaved(FileObject lastSavedJobFile, AnalysisJobBuilder analysisJobBuilder) {
+        if (lastSavedJobFile == null) {
```
kaspersorensen: here we can add a few more precautionary optimizations. Also check:

lastSavedJobFile.exists()
lastSavedJobFile.length() > 0

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -558,6 +582,33 @@ protected boolean onWindowClosing() {
...
+        InputStream lastSavedOutputStream = null;
+        ByteArrayOutputStream currentOutputStream = null;
+        try {
+            lastSavedOutputStream = getJobFile().getContent().getInputStream();
+            JaxbJobWriter writer = new JaxbJobWriter(_configuration);
+            currentOutputStream = new ByteArrayOutputStream();
+            writer.write(_analysisJobBuilder.toAnalysisJob(false), currentOutputStream);
+            String currentJob = new String(currentOutputStream.toByteArray());
+            String currentJobNoMetadata = currentJob.replaceAll("\n", "").replaceAll("<job-metadata>.*</job-metadata>", "");
+            String lastSavedJob = IOUtils.toString(lastSavedOutputStream);
```
kaspersorensen: We have a similar helper thingie in FileHelper.readFileAsString(...). Not that it matters much, but I guess we already use FileHelper in a lot of places so it might be better?

ClaudiaPHI:
I will take a look too

tomaszguzialek:
I did not experience any delay between clicking close button and confirmation dialog. I tested it with a job with 3 components.

tomaszguzialek:
Added additional checks and make use of the utility method.

kaspersorensen:
OK I am by now okay with taking this contribution. But it can no longer be auto merged ;-)

PR #227, opened 2015-02-06
Fixes an issue where two mouse listeners (JobGraphLinkPainter and

kaspersorensen:
JobGraphMouseListener) both act on a right-click and causing an
exception because graph was refreshed inbetween actions.


tomaszguzialek:
I wanted to reproduce the problem on the master branch. I have such a job:
graph

I tried to right-click and link Customers with Tokenizer. No exception seen in the console, however it won't let me connect.

How can I reproduce the bug?

kaspersorensen:
Actually I found out that it doesn't relate to filters at all.

To reproduce:

1) Add a source table and any component (don't connect it yet).
2) Right-click the source table and select "Add link".
3) Drag the link to the component, but instead of left-clicking, do a right-click. BOOM.

tomaszguzialek:
Reproduced successfully now and the fix works as expected. Merging.

PR #230, opened 2015-02-06
Issue #229: CLI future aware

### LINKED ISSUE #229: Make command line job execution aware of AnalyzerFutureResult
Command line execution of jobs does not respect that analyzers can return futures instead of ready results.

tomaszguzialek:
Fixes #229

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/result/renderer/ToStringFutureTextRenderer.java
+++ b/engine/core/src/main/java/org/datacleaner/result/renderer/ToStringFutureTextRenderer.java
@@ -1,0 +1,77 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
+ * copy, or redistribute it subject to the terms and conditions of the GNU
+ * Lesser General Public License, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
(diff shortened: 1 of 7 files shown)


engine/core/src/main/java/org/datacleaner/result/renderer/ToStringFutureTextRenderer.java
```diff
@@ -0,0 +1,77 @@
...
+import org.datacleaner.api.AnalyzerResult;
+import org.datacleaner.api.AnalyzerResultFuture;
+import org.datacleaner.api.Renderer;
+import org.datacleaner.api.RendererBean;
+import org.datacleaner.api.RendererPrecedence;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A renderer for analyzers returning {@link Future}. Waits for the Future to
```
kaspersorensen: The Future link here is incorrect. Should be to AnalyzerResultFuture :-)

engine/core/src/main/java/org/datacleaner/result/renderer/ToStringFutureTextRenderer.java
```diff
@@ -0,0 +1,77 @@
...
+                return resultString.toString();
+            } else {
+                final String message = "No renderer found for result type " + result.getClass().getName();
+                logger.error(message);
+                throw new IllegalStateException(message);
+            }
+        } catch (RuntimeException error) {
+            String message = "Unable to fetch result";
+            logger.error(message, error);
+            throw new IllegalStateException(message);
```
kaspersorensen: Add the error as a cause also?

tomaszguzialek:
Fixed the issues mentioned.

PR #231, opened 2015-02-09
Feature/new welcome screen

### LINKED ISSUE #221: No scrollbars on (old) welcome screen
On the welcome screen (what will become the database manager later), there are no scrollbars added when the window is too small to handle the amount of datastores.

kaspersorensen:
New welcome screen and related panels.

Makes #221 obsolete, so...

Fixes #221


kaspersorensen:
There is actually two sides to the review of this PR.

 * The regular code review (code style, blunders, weird things, questions etc.)
 * How does the suggested UI change work for end users.

To evaluate the second point I am right now uploading a build to sourceforge's unstable build store ... Will distribute it to a few people who might provide their feedback.

kaspersorensen:
Downloadable here: http://sourceforge.net/projects/datacleaner/files/datacleaner%20%28unstable%29/4.0%20alpha/DataCleaner-4.0-alpha9.zip/download

kaspersorensen:
One critical issue: The "SelectDatastorePanel" does not update it's datastore list if it changes. This is needed before we can merge this branch I guess...

kaspersorensen:
image

kaspersorensen:
Latest suggestion/state-of-affairs:

(select datastore)
image

(datastore management)
image

Note that there is an intentional overlap with content and the background image. We can tweak how much, for now I just played with it a bit and I would love your remarks.

LosD:
The OO.org icon looks a little ugly in the dark background of the tooltip.

engine/core/src/main/java/org/datacleaner/util/sort/SortMergeWriter.java
```diff
@@ -45,8 +46,6 @@
  * Sorter, deduplicator and writer that uses temporary files as storage to
  * support high volume sorted data.
  * 
- * Note: This class is NOT thread-safe.
```
LosD: Now it is :)

(what happened here? It doesn't seem related at all, was it just something the new UI uncovered?)

LosD:
It looks good, and the code seems as good as Swing lets us make it, but as I'm pretty invested here, I may not be the best to do that part of the review. :)

engine/core/src/main/java/org/datacleaner/util/sort/SortMergeWriter.java
kaspersorensen: You're right ... Another silly thing happened (feedback from our tester Jan Arie) and I accidentally committed the fix on this branch ... DOH.

kaspersorensen:
Doing it ... merrrrrrge.

PR #232, opened 2015-02-09
Feature/222 errors on canvas

### LINKED ISSUE #222: Show job validation error on canvas background
We have a nice "getting started" set of messages on the background of the canvas. This dissappears when all the "getting started" steps are finished. But if a component is still not fully configured, it's invalidation message will appear in the status label. It would be nice that it would instead appear in the canvas background.

Note: In terms of precedence the other "getting started" messages should come first. It is quite usual that there is a status label such as "No analyzers in job" which we do not want to show on the canvas instead of the "getting started" messages.

ClaudiaPHI:
Fixes #222


desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
```diff
@@ -278,9 +284,31 @@ public void paint(Graphics g) {
                             + "So add one or more such components.";
                     imagePath = "images/window/canvas-bg-plus.png";
                 } else {
-                    title = null;
+                    title=null;
```
kaspersorensen: please remember to obey the standard code formatting.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
```diff
@@ -302,6 +330,13 @@ public void paint(Graphics g) {
                 }
 
                 if (subTitle != null) {
+                    
+                    if (error){
+                        g.setColor(WidgetUtils.ADDITIONAL_COLOR_RED_BRIGHT);
```
kaspersorensen: I think making it red will make it way too alarming. This is a quite big message already and it will also feature a "error" icon in the status bar and a icon decoration on the component itself.... Should be enough :-)

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
```diff
@@ -278,9 +284,31 @@ public void paint(Graphics g) {
...
+                    } catch (Exception ex) {
+                        logger.debug("Job not correctly configured", ex);
+                        final String errorMessage;
+                        if (ex instanceof UnconfiguredConfiguredPropertyException) {
+                            UnconfiguredConfiguredPropertyException unconfiguredConfiguredPropertyException = (UnconfiguredConfiguredPropertyException) ex;
+                            ConfiguredPropertyDescriptor configuredProperty = unconfiguredConfiguredPropertyException
+                                    .getConfiguredProperty();
+                            ComponentBuilder componentBuilder = unconfiguredConfiguredPropertyException.getComponentBuilder();
+                            errorMessage = "Property '" + configuredProperty.getName() + "' in "
+                                    + LabelUtils.getLabel(componentBuilder) + " is not set!";
```
kaspersorensen: Too technical for this space. We want to make it very friendly in the canvas' big text. So.
For the main title use a standard (but friendly, not "robot language") message about some component not being in a good state.
Keep the details to the subtitle.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
```diff
@@ -278,9 +284,31 @@ public void paint(Graphics g) {
...
+                        } else {
+                            errorMessage = ex.getMessage();
+                        }
+                        subTitle="Job error status: " + errorMessage;
```
kaspersorensen: This is really what I mean by "robot language". Don't have the "Job error status: " prefix here. Just keep the error message (as subtitle) and make sure the main title is human language.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
LosD: https://www.youtube.com/watch?v=O2uFnQKlXK8 :D

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
LosD: While I agree the message is not friendly enough (Especially exclamation marks is a bad idea), I actually think that we _need_ to tell what to look for, especially when the dialogs don't highlight what the actual problem is.

LosD: How about "Unfortunately, the property x on the component y is required, please set it to continue"

Of course, it may be a problem when there are more components of the same type, or several problems, though the latter is not critical (there will just be a new message when they fixed the first problem)

kaspersorensen: I like it, but would like to refine it a bit more :-)

How about "Please set X on Y to continue".

My point being that "property" and "component" are quite technical words and might not resonate with the user. The user will just see a graph node called "Y" and when he double clicks he will see a label for "X". The fact that we call these components and properties are "nice to know" but if we can make meaningful messages to the user without using those words, then I think it's an advantage.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
ClaudiaPHI: https://www.youtube.com/watch?v=L_AZQv6AYis

kaspersorensen:
Eh... Is it supposed to look like this? I am missing a symbol/image and both a main title and a subtitle...

image

ClaudiaPHI:
Better 

image

ClaudiaPHI:
New version: 

image

image

PR #236, opened 2015-02-11
Remove (AnyComponentRequirement[]) from component labels #201

### LINKED ISSUE #201: Remove (AnyComponentRequirement[]) from component labels
In the UI of DataCleaner we now see "AnyComponentRequirement[]" in the labels of the components in the result window. I think this is coming from LabelUtils. We should not show this specific component requirement in text.

ClaudiaPHI:
Fixes #201 I hope ... I couldn't reproduce in order to check it


kaspersorensen:
LGTM

PR #247, opened 2015-02-13
Feature/245 overwrite if exists sheets

### LINKED ISSUE #245: Change scope of Excel "overwrite file if exists" flag to be sheet-oriented
We did actually previously support having multiple Excel writers at the same time to the same file. But now that does not work because of the new "overwrite file if exists" flag. The reason is that multiple writers could write to separate sheets (in the same file).

So we need to change the scope of the flag to be sheet-based and not file-based.

ClaudiaPHI:
Fixes #245


components/writers/src/main/java/org/datacleaner/extension/output/CreateExcelSpreadsheetAnalyzer.java
```diff
@@ -75,8 +82,18 @@ public void validate() {
             throw new IllegalStateException("Sheet name cannot contain dots (.)");
         }
 
-        if (file.exists() && !overwriteFileIfExists) {
-            throw new IllegalStateException("The file already exists and the columns selected do not match. Please configure the job to overwrite the existing file.");
+        if (file.exists()) {
+            Datastore datastore = new ExcelDatastore(file.getName(), new FileResource(file), file.getAbsolutePath());
+            final Schema[] schemas = datastore.openConnection().getDataContext().getSchemas();
```
kaspersorensen: the openConnection() call returns a DatastoreConnection. This needs to be closed in a finally block once you're finished with the operation :-)

kaspersorensen: Oh or the new smarty try-with-resources style which I see you're using later.

components/writers/src/main/java/org/datacleaner/extension/output/CreateExcelSpreadsheetAnalyzer.java
```diff
@@ -94,14 +111,31 @@ public void configureForTransformedData(AnalysisJobBuilder ajb, TransformerDescr
...
-            } else {
-                throw new IllegalStateException("The existing file could not be deleted");
+        if (file.exists()) {
+            ExcelDatastore datastore = new ExcelDatastore(file.getName(), new FileResource(file),
+                    file.getAbsolutePath());
+            try (final UpdateableDatastoreConnection connection = datastore.openConnection()) {
+                final DataContext dataContext = connection.getDataContext();
+                final Schema[] schemas = dataContext.getSchemas();
+                if (schemas.length >= 1) {
+                    final String[] tableNames = schemas[1].getTableNames();
```
kaspersorensen: Better way to get the proper schema is to say dataContext.getDefaultSchema() ... taking schemas[1] is actually not guaranteed to work since schemas are sorted alphabetically and then you rely on the filename (which is used as schema name) to be sorted after "information_schema" :-)

components/writers/src/main/java/org/datacleaner/extension/output/CreateExcelSpreadsheetAnalyzer.java
```diff
@@ -94,14 +111,31 @@ public void configureForTransformedData(AnalysisJobBuilder ajb, TransformerDescr
...
+                    for (int i = 0; i < tableNames.length; i++) {
+                        if (tableNames[i].equals(sheetName)) {
+                            if (overwriteSheetIfExists) {
+                                final Table tableSheet = dataContext.getTableByQualifiedLabel(sheetName);
+                                final UpdateableDataContext updateableDataContext = connection
+                                        .getUpdateableDataContext();
+                                updateableDataContext.executeUpdate(new UpdateScript() {
```
kaspersorensen: Tip: Instead of the update script you can also use an atomic update here:

...executeUpdate(new DropTable(tableSheet));

kaspersorensen:
Looks good, just a few minor thingies.

ClaudiaPHI:
Can I merge it ?

kaspersorensen:
Yup :)

PR #254, opened 2015-02-16
Feature/249 enum mapping

### LINKED ISSUE #249: Error while mapping enum to columns
When one tries to map an enum to columns using "Select All" option we get an error. 

image

kaspersorensen:
Fixes #249


ClaudiaPHI:
Fine with me .

PR #255, opened 2015-02-16
Issue #242: Fixed.

### LINKED ISSUE #242: Filter outcome "Write to" menu creates 2x the necesary components!
Pretty strange - when selecting to write to eg. an Excel or CSV file, 2x the components are created:

image

kaspersorensen:
Fixes #242


ClaudiaPHI:
ok. If it works it works :-)

PR #258, opened 2015-02-16
Fixes #244 (excel sheet naming)

### LINKED ISSUE #244: Generated sheet name in "Create Excel spreadsheet" might contain dots
... But dots are not allowed in sheet names :-) Make sure that our generation of sheet names will always produce a valid string.

kaspersorensen:


kaspersorensen:
Apparently there's a unittest failure ... fixing

PR #268, opened 2015-02-17
Removed unnecessary WebappLifecycleListener to fix deployment on JBoss/Wildfly

michaelaerni:
The WebappLifecycleListener has no Methods which can be accessed by any Java Application Server. Listener Classes without any declared Listener Interfaces cannot be declared in a web.xml. While Tomcat ignores this, Wildfly and JBoss AS both won't deploy the war. Because the WebappLifecycleListener does not implement any compatible Listener Interface, removing it should not have any impact.

### DIFF
--- a/monitor/ui/src/main/webapp/WEB-INF/web.xml
+++ b/monitor/ui/src/main/webapp/WEB-INF/web.xml
@@ -40,9 +40,6 @@
 	<listener>
 		<listener-class>com.sun.faces.config.ConfigureListener</listener-class>
 	</listener>
-	<listener>
-		<listener-class>com.sun.faces.application.WebappLifecycleListener</listener-class>
-	</listener>
 
 	<!-- Spring listener -->
 	<listener>


kaspersorensen:
Hmm as far as I understand this class does serve a purpose. It does not implement any interface no, but it does have PreDestroy (and maybe more) annotated methods.

I read elsewhere that the issue is not so much that the class is registered. The issue is that the WEB-INF/lib directory contains also jsf-api.jar .... JBoss already has jsf-api bundled in the container, but Tomcat doesn't.

michaelaerni:
The WebappLifecycleListener itself does not have any annotated methods as seen in the source: http://search.maven.org/remotecontent?filepath=org/glassfish/javax.faces/2.1.9/javax.faces-2.1.9-sources.jar therefore it cannot be used by the application server anyways.

Its purpose is mainly to handle other methods with the PreDestroy annotation. If you take a look at the ConfigureListener class (which is also declared as a listener) you can see that it internally uses the WebappLifecycleListener to perform its actions.

kaspersorensen:
You're right. Thank you for clarifying and thank you for the fix! :-)

PR #269, opened 2015-02-17
Feature/246 sorting capabilities

### LINKED ISSUE #246: Add sorting capability to "Create Excel spreadsheet" component
... Just like we now have it for "Create CSV file" component.

ClaudiaPHI:
Fixes #246


kaspersorensen:
This looks good functionally. Can we add a few code comments - especially in the createOutputWriter() method which may (temporarily) write to a CSV file ... That stuff triggers a few "What's going on?!" questions in my mind when I read it ... But now I understand that it is only temporary storage. Would be nice to make that clear with comments and maybe also encapsulate the part of the method that does this into a private method called something a la "createTemporaryCsvFileWriter()"

ClaudiaPHI:
ok doki

ClaudiaPHI:
Can I merge it ?

kaspersorensen:
Yes

kaspersorensen:
And so can I, so I will :)

ClaudiaPHI:
evil boss :-P

PR #273, opened 2015-02-18
Fixes #204 by using the resource reference instead of filename.

### LINKED ISSUE #204: Better example jobs that utilize the graph UI's power
Right now the example jobs are quite simple because they had to be easy to understand in the old UI.

But given the new UI we can make some much more interesting example jobs.

Stuff for consideration:

 * Filtering on various things (maybe null check or so) and putting invalid records into "buckets".
 * Applying multiple analyzers to show many dimensions of the same field(s).
 * Transformations that involve reference data or search/replace or so.
 * Use of the "Fuse flows" transformation to merge two filtered flows back into one.
 * Make an example job that does job title standardization on customers file

### LINKED ISSUE #239: Datastores created via Dropzone cannot be edited - "File does not exist"
image

kaspersorensen:
Fixes #239 
Does not fix #204


LosD:
Hmmm... Is that the right "Fixes"? #204 is a story about making better examples, this looks like a bugfix?

kaspersorensen:
Sorry, yes wrong story ... Will update!

ClaudiaPHI:
Fine with me but a little bit confused #204

PR #282, opened 2015-02-19
Feature/281 datacleaner home system property

### LINKED ISSUE #281: DATACLEANER_HOME cannot be set though system property
DataCleanerHome class should respect DATACLEANER_HOME being set in system property, not only in an environmental variable.

Setting the environmental variable is sometimes not possible or cumbersome, e.g. Eclipse run configuration or application bundle for Macs.

tomaszguzialek:
Fixes #281

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
+++ b/desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
@@ -109,7 +109,7 @@ public final class DataCleanerHome {
                 logger.info("Running in standard mode. Attempting to build DATACLEANER_HOME in {}", candidate);
             } else {
                 // Workaround: isWritable is not reliable for a non-existent directory. Just create it and check again.
-                if (!candidate.exists()) {
+                if ((candidate != null) && (!candidate.exists())) {
                     candidate.createFolder();
                 }
                 if (isWriteable(candidate)) {
@@ -171,6 +171,10 @@ public final class DataCleanerHome {
     }
 
     private static boolean isWriteable(FileObject candidate) throws FileSystemException {
+        if (candidate == null) {
+            return false;
+        }
+        
         if (!candidate.isWriteable()) {
             return false;
         }
--- a/desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
+++ b/desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
@@ -132,10 +132,12 @@ public final class DataCleanerHome {
                         logger.info("Folder {} created successfully. Attempting to build DATACLEANER_HOME here.",
                                 candidate);
                     } catch (FileSystemException e) {
+                        logger.info("Unable to create folder {}. No write permission in that location.", candidate);
                         candidate = initializeDataCleanerHomeFallback();
                     }
+                } else {
+                    candidate = initializeDataCleanerHomeFallback();
                 }
-
             }
 
         }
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
```diff
@@ -84,7 +84,13 @@ private static FileObject findDataCleanerHome() throws FileSystemException {
         if (!StringUtils.isNullOrEmpty(env)) {
             candidate = manager.resolveFile(env);
             logger.info("Resolved env. variable DATACLEANER_HOME ({}) to: {}", env, candidate);
-        }
+        } else {
+        	final String sysProp = System.getProperty("DATACLEANER_HOME");
+        	if (!StringUtils.isNullOrEmpty(sysProp)) {
+        		candidate = manager.resolveFile(sysProp);
```
kaspersorensen: tabs are used here for indentation. We normally use 4x space.

desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
```diff
@@ -97,9 +103,11 @@ private static FileObject findDataCleanerHome() throws FileSystemException {
             logger.info("Running in WebStart mode. Attempting to build DATACLEANER_HOME in user.home: {} -> {}", path,
                     candidate);
         } else {
-            // in normal mode, the default folder will be in the working
-            // directory
-            candidate = manager.resolveFile(".");
+            if (candidate == null) {
```
kaspersorensen: this if-check should not be used. We have the isUsable(...) for making sure that the candidate is good.

tomaszguzialek:
Fixed indendation and reused isUsable().

desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
```diff
@@ -97,9 +103,11 @@ private static FileObject findDataCleanerHome() throws FileSystemException {
             logger.info("Running in WebStart mode. Attempting to build DATACLEANER_HOME in user.home: {} -> {}", path,
                     candidate);
         } else {
-            // in normal mode, the default folder will be in the working
-            // directory
-            candidate = manager.resolveFile(".");
+            if (!isUsable(candidate)) {
```
kaspersorensen: already determined by the surrounding if :)

tomaszguzialek:
Now DataCleanerHome should support non-existent (yet) paths. isUsable() is a method that checks also for existence of the conf.xml file inside, so a non-existent directory never passed this test and it immediately fell back to the working directory instead. Now it is given a chance to create it before using the fallback.

Had to include a couple of lines more as isWritable() returns false on a non-existent directory, but actually it can be created... So I am creating if forcefully and checking if isWritable() is going to pass this time.

tomaszguzialek:
Ergh, build failed. Let me see.

kaspersorensen:
I think the code is still a bit intertwined and strange... It could be reorganized to avoid difficult-to-debug situations. Also the logging statements are the same even though it may originate from different scenarios. This is a particular difficult thing to "get right" because we have seen many error reports when people e.g. invoke the executable from a different directory and so on. Lots of corner cases, so we need to make sure the code is ultra clear.

tomaszguzialek:
Yeah, I wanted to fit in the existing code, if possible. I will rework the whole method then.

tomaszguzialek:
Divided findDataCleanerHome method into two. findDataCleanerHome is responsible for resolving environmental variable and/or a system property DATACLEANER_HOME and checking if the conf.xml is already there. If not, initializeDataCleanerHome comes into the game and builds the DataCleaner directory.

desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
```diff
@@ -97,17 +119,37 @@ private static FileObject findDataCleanerHome() throws FileSystemException {
...
+                // Workaround: isWritable is not reliable for a non-existent
+                // directory. Just create it and check again.
+                logger.debug(
+                        "DATACLEANER_HOME ({}) appears to be not writable. Applying a workaround: creating this directory and checking again.",
+                        candidate);
+                if ((candidate != null) && (!candidate.exists())) {
+                    candidate.createFolder();
+                }
+                if (isWriteable(candidate)) {
+                    logger.info("Attempting to build DATACLEANER_HOME in {}", candidate);
```
tomaszguzialek: This log statement is still identical to line 127, because it is the same scenario - we just need to check it twice because of the workaround. Adding a log statement for the workaround presence with DEBUG level too.

ClaudiaPHI:
@LosD Can you review this ? I think I will pass.

LosD:
Owkey dokey! :)

desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
```diff
@@ -80,16 +80,38 @@ private static FileObject findDataCleanerHome() throws FileSystemException {
...
+        } else {
+            String sysProp = System.getProperty("DATACLEANER_HOME");
+            if (!StringUtils.isNullOrEmpty(sysProp)) {
+                if (sysProp.startsWith("~")) {
+                    String userHomePath = System.getProperty("user.home");
+                    sysProp = sysProp.replace("~", userHomePath);
+                }
+                candidate = manager.resolveFile(sysProp);
+                logger.info("Resolved system property DATACLEANER_HOME ({}) to: {}", sysProp, candidate);
+            }
```
LosD: The duplication here seems rather unnecessary.

desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
```diff
@@ -80,16 +80,38 @@ private static FileObject findDataCleanerHome() throws FileSystemException {
...
         }
 
         if (isUsable(candidate)) {
+            // Found a directory with conf.xml already there
             return candidate;
+        } else {
+            return initializeDataCleanerHome(candidate);
         }
```
LosD: Very minor, but if you moved the isUsable() check into initializeDataCleanerHome (which would just return candidate if it was already initialized), you'd save a bit of code.

LosD: Actually, it seems to already be there. There's a bit of directory checking first, which can take a few milliseconds, but I guess this only runs once during upstart, so an extra check first seems a bit overkill to me

desktop/ui/src/main/java/org/datacleaner/user/DataCleanerHome.java
```diff
@@ -97,17 +119,37 @@ private static FileObject findDataCleanerHome() throws FileSystemException {
...
-                logger.info("Running in standard mode. Attempting to build DATACLEANER_HOME in '.' -> {}", candidate);
+                logger.info("Attempting to build DATACLEANER_HOME in {}", candidate);
             } else {
-                final String path = getUserHomeCandidatePath();
-                candidate = manager.resolveFile(path);
-                logger.info(
-                        "Application directory is not writeable. Attempting to build DATACLEANER_HOME in user.home: {} -> {}",
-                        path, candidate);
+                // Workaround: isWritable is not reliable for a non-existent
+                // directory. Just create it and check again.
```
LosD: If you can create it, wont you always be able to write to it?

LosD:
I'm really not too fond of the whole isWriteable() structure, but I guess each directory is different enough to not make a proper dedup (heh!) refactoring viable.

Besides that, there's only a few minor things.

LosD:
I don't think this code will ever be beautiful, but I think we have crossed the "good enough" line. Merging!

PR #290, opened 2015-02-20
Feature/common metadata

kaspersorensen:
Please review this enum. It's supposed to form the basis of a simple column classification mechanism to be reused in wizards, components etc. No concrete use of it yet in DC itself, but I see several downstream usages - extensions, DC EE etc.

### DIFF
--- a/api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
+++ b/api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
@@ -1,0 +1,160 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
```diff
@@ -0,0 +1,237 @@
...
+    // PHONE oriented items
+
+    PHONE_PHONENUMBER("Phone number", "Phone", "Mobile", "Mobile phone", "Cellphone", "Telefon", "Mobil", "Tel", "Mob"),
+
+    PHONE_FAX("Fax number", "Fax", "Faxnummer"),
+
+    // EMAIL oriented items
+
+    EMAIL_ADDRESS("Email", "Mail", "Email address", "Mail address", "@"),
+
```
ClaudiaPHI: Can we also add "e-mail" and/or "E-mail" ? 
http://en.wikipedia.org/wiki/Email

ClaudiaPHI:
Fine with me.

PR #292, opened 2015-02-20
Right click on canvas issue

### LINKED ISSUE #206: Right click is lazy in graph UI
•	The right click on the graph ui is lazy, it appears late. Sometimes the menu shows up with one click, otherwise I have to click multiple times until the menu shows up again .

ClaudiaPHI:
Fixes #206 
Made the component react on mousePressed event and changed the graph to react on mouse released event.


LosD:
How much will you kill me if I ask you to also move the canvasRightClicked to Released (mouseReleased() in this case)? It is a bit strange that when it is the tool menu, it pops up as soon as the right mouse button is pressed, but when it is the component menu, it pops up when it is released.

PR #294, opened 2015-02-22
Icon updates

kaspersorensen:
What do you think? Please do a quick sanity check and if you agree these are generally better icons, please merge.

In addition to updating the icons, a bunch of reorganization has been done to make icon resolving easier and better:

 * Specific icons moved to the specific component modules
 * IconUtils has now more use-oriented ICON_SIZE constants (e.g. ICON_SIZE_MENU_ITEM etc.)
 * If there are no specific icon for a component, no icon is shown in the menus. On the canvas they have updated generic icons.

image


ClaudiaPHI:
Good with me. But it is a little strange that we do not have icons in the menu

tomaszguzialek:
Having no icons is still better IMO than a wall of generic icons. At least the ones that do have an icon stand out more in the menu and are easier to find. I will merge it.

kaspersorensen:
I'll try and update/make some more as we go :)

PR #296, opened 2015-02-23
added a prefered size for the ComponentConfigurationDialog

### LINKED ISSUE #261: The configuration dialogs are too big sometimes
The configuration dialogs are bigger than the screen sometimes. 

image

ClaudiaPHI:
Fixes #261


LosD:
Seems good to me, and seems to work well.

Some components looks a little silly with plenty of space and the optional properties closed, but that is so minor that calculating if there would be space is not worth it.

Merging!

kaspersorensen:
Since the contents of the dialogs is variable, aren't we just now running the risk of a new "The configuration dialogs are too small sometimes" issue? Would be cool to have the old dynamic sizing UNLESS it was very close to the screen height in which case we could subtract a bit...?

LosD:
Hmmmm, the dynamic dialog sizing newer worked very well. Better no dynamics that works perfectly every time, than dynamics that may or may not get the size right.

PR #297, opened 2015-02-23
Dropzone - use existing datastore

### LINKED ISSUE #240: Check (and use) existing datastore if file is selected in Dropzone
When selecting a file in the "Select datastore" dropzone, we should reuse any existing datastore for the file, if possible.

ClaudiaPHI:
Fixes #240  Check if the datastore catalog already contains the file. If so select the existing datastore. Otherwise create a new one

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/Dropzone.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/Dropzone.java
@@ -65,8 +65,8 @@ public class Dropzone extends DCPanel {
     private final DatastoreSelectedListener _datastoreSelectListener;
     private final UserPreferences _userPreferences;
 
-    public Dropzone(final DatastoreCatalog datastoreCatalog,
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/Dropzone.java
```diff
@@ -115,8 +115,13 @@ protected void showFileChooser() {
             final File file = fileChooser.getSelectedFile();
 
             if (file.exists()) {
-                final Datastore datastore = DatastoreCreationUtil.createAndAddUniqueDatastoreFromFile(
-                        _datastoreCatalog, file);
+                final Datastore datastore;
+                final String filename = file.getName();
+                if (_datastoreCatalog.containsDatastore(filename)) {
```
tomaszguzialek: This checks if the name of the datastore is same as the filename. What if I have this file in my datastores already, but I gave it a custom name instead of a default filename-based name? Shouldn't we check for the whole path to make sure it is the same file?

PR #298, opened 2015-02-23
Toolbox in the tree

### LINKED ISSUE #215: Component catalog should show up in the tree to the left
Multiple user interviews have revealed that people would like to see the component catalog in the tree to the left side of the canvas. This is because then they don't need to dig into the menus to find them - the tree is shown continuously and always available. Furthermore it will lend itself to drag'n'drop behaviour when adding components, which would be nice and natural.

LosD:
This will fix #215.

This also disabled the animation, as it got very choppy because the contents gets rendered while it is animating, and it was the cause of the semi-transparent tree.

The triple click to place tables is now a thing of the past. A side effect of this is that the table will expand when you double click to place it.

Finally, the menu was pretty weird regarding right clicks, as it looked on last selected item instead of what is actually right-clicked on. That has been fixed.


kaspersorensen:
Nice to hear of all the fixes! The PR cannot be automatically merged, so can you please merge in master?

kaspersorensen:
and I would love a screenshot if possible :D

LosD:
Errrrr.... After merge, the tree doesn't auto-expand anymore. Is that on purpose?

LosD:
Anyway, it's been merged, and here's a screenshot:
image

Something also seems to have happened to the icon sizes after. They used to fit in the tree :)

LosD:
Whoops, I should probably also kill the top menu.

LosD:
Hmmm... Thinking about it, I'm still missing a lot of stuff. The tooltips should probably be there (though I don't think the DCBubbleThingy will work very well with the tree... It's too fast.), and I still don't have a solution for new extensions.

kaspersorensen:
Expand thing: Not on purpose. But I did a fix adding expandTable(...) which probably does not work now that there's a new root node.

And on that thought the... The new root looks weird and the yellow folders could use the blue folder icon :-) 

But... It Also ALSO looks very cool and functional.

Can we rename Toolbox to something else? I'm thinking Components or Library?

LosD:
Okay, will update it with the new structure.

I agree on the root node. Apparently invisible doesn't mean invisible. If it cannot be properly invisible, I think that I prefer it with a text, though I have no idea what it should be... _Root_ sounds idiotic. We should also find a proper icon.

Regarding _Toolbox_, I prefer _Library_. For the icon, we should probably come up with something that signifies what it is. It cannot be a toolbox, I guess :)

Regarding the tooltip, I think I'll be using the same tooltips as the context menu, but I'm thinking of removing the icon and title; It seems a bit silly to have them when they are right there over the tooltip... It actually makes it look less connected to what they are describing. I still can't use a simple tooltip, though, as it doesn't break long lines.

kaspersorensen:
Yeah also prefer a text if we cannot make the root invisible. But let's make absolutely sure we cannot make it invisible - it would be a whole lot prettier I think.

OK Library it is then. I think we have some icons we can use, or else we can make a new variant of the (new) blue analyzer.png, transformer.png like images maybe.

Tooltip without icon and title sounds fine in deed. I was even thinking of (a future feature) where you could right click in the tree and see an extended description. A kind of javadoc-page with everything known about the component based on all it's metadata, properties etc.

LosD:
Uhhh, that last thing would be nice!

kaspersorensen:
Also more like a library (for reading) then ;-)

LosD:
:laughing:

LosD:
Tooltips added.
image

LosD:
By the way: setRootInvisible works fine. There's just this small detail with it: It works best when you actually call it :)

There's also another matter: When there's no root, there is no collapsible icon, so you can _only_ open and close it by double clicking. The initial open will be fixed when auto-expansion is taken care of, but I'd like to find a way to make it impossible to close them, so that we don't have confused customers wondering where there ~~toolbox~~ library went

kaspersorensen:
Yeah that would be best. Small note/reminder for the tooltips: Some components have linebreaks in their descriptions. Can you verify if those are respected? If not it's no catastrophe, but would be a nice to have thing.

LosD:
That logic has been copied directly from the context menu tooltips (except the text is now updated on the text changing, as the path isn't available when createTooltip runs), so I couldn't image anything else, but will verify.

LosD:
Works! :)
image

LosD:
Hmnmm... The missing icon looks pretty silly when it is in the tree:
image

I'll stay with generic icons instead for now.

kaspersorensen:
Not sure if you're finished with all the various "todo items" yet, so do you want the review yet, or shall I wait?

kaspersorensen:
I played around with it for 2 minutes, and really like what I see. A few remarks though (I can imagine you are aware of most of this):

 * The two root-nodes (or what-looks-to-the-user-as-root-nodes) should always be expanded I think.
 * Initially the tree's panel was very very slim looking and did not give enough attention to the left part of the screen. But I think this will change if we just make the two nodes expanded.
 * The super category icon colors are too black - it's a huge contrast compared to the icons of schemas, tables etc... Let's try with a lighter color or maybe blue if that does not work.
 * We need a dedicated icon for the "library" node IMO. Maybe something like this: http://commons.wikimedia.org/wiki/File:Applications-accessories.svg (but flattened/simplified).
 * I agree on the choice of "generic" icons instead of blank icons where necesary.
 * Very minor (can be ignored unless it's a dead simple thing): The pointer when drag-dropping a component to the screen looks like the pointer of a table. In this screenshot I am dragging "Convert to number" component which has a very different icon:

image

kaspersorensen:
And a final remark (but this is not directly related to this issue - more a consequence of it) ... I think we need to reconsider some of the categories of the components. Especially the "Analyze" menu seems like it should not have any sub-categories since this is just hiding functionality that can be shown on the screen easily.

LosD:
Hmmmm... The two notes should have auto-expanded (and will never collapse) since 03fde12. They also do for me.

The supercategory icons are a huge hack. I use a filter to invert the colors of the white ones (it is actually surprisingly fast; sub-millisecond). I fully expect our resident icon-maker to make pretty ones when he returns :)

Fully agree on the library icon, and on the component categories. 

It is weird that you see the table icon. I only see the move pointer:
image

LosD:
TODOs done.

kaspersorensen:
I merged in the master branch to have the latest-and-greatest. Maybe it caused some of these issues.

Initially if I select orderdb, this is how the tree looks like (all collapsed):

image

With a CSV file it does expand. So I think the "library expansion" is only happening for single-table datastores, but it should happen always IMO.

I still think the super category colors are too dark - too much contrast with the white background, making it a bit hard to the eye.

Minor thing: The right click popup is using quite big icons. In IconUtils there is now a constant ICON_SIZE_MENU_ITEM which should probably be used here (as it is used in all popup menus):

image

kaspersorensen:
Icons updated:

image

PR #299, opened 2015-02-23
Another icon update

kaspersorensen:
Please check them out again, and if we have more eyes confirming that they are better - let's commit.

Here's a few screenshots:

image

image


ClaudiaPHI:
I think the icons are nice and much better. I will merge it.

PR #301, opened 2015-02-24
Issue #300: Adding support for Cmd + click.

### LINKED ISSUE #300: Cmd + click on Mac should result in a right click.
One of the ways to achieve a right click in Macs is clicking one-button mouse with Cmd key pressed. We should support this.

tomaszguzialek:
Fixes #300


kaspersorensen:
Very cool. I'll merge it. And also plea that we should do the same thing in the SchemaTree as it also has left/right-click handling in place. But maybe wait for/check with #298 since it is same codebase.

PR #311, opened 2015-03-05
Feature/307 bulk node removal

### LINKED ISSUE #307: Support for bulk node removal in job graph
Make it possible to select multiple nodes in the graph and do a batch/group removal of the items.

Note: Keep in mind that there are different types of "things" in the graph. Source tables and components are removed in different ways on the backing job builder.

tomaszguzialek:
Fixes #307

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphContext.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphContext.java
@@ -88,16 +88,21 @@ public class JobGraphContext {
     }
 
     public Object getSelectedVertex() {
-        final PickedState<Object> pickedVertexState = _visualizationViewer.getPickedVertexState();
-        if (pickedVertexState == null) {
-            return null;
-        }
-        final Set<Object> pickedVertices = pickedVertexState.getPicked();
+        Set<Object> pickedVertices = getSelectedVertices();
         // ensure that there is 1 and only 1 selected vertex
         if (pickedVertices == null || pickedVertices.size() != 1) {
             return null;
         }
         return pickedVertices.iterator().next();
+    }
+    
+    public Set<Object> getSelectedVertices() {
+        final PickedState<Object> pickedVertexState = _visualizationViewer.getPickedVertexState();
+        if (pickedVertexState == null) {
+            return null;
+        }
+        final Set<Object> pickedVertices = pickedVertexState.getPicked();
+        return pickedVertices;
     }
 
 }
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphContext.java
```diff
@@ -88,16 +88,21 @@ public Object getVertex(Point2D p) {
     }
 
     public Object getSelectedVertex() {
```
LosD: Is this still used anywhere after adding getSelectedVertices?

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphKeyListener.java
```diff
@@ -49,25 +50,27 @@ public JobGraphKeyListener(JobGraphContext graphContext) {
     public void keyTyped(KeyEvent e) {
         // react to DEL key strokes and delete components that are selected.
         if (e.getKeyChar() == KeyEvent.VK_DELETE) {
-            final Object vertex = _graphContext.getSelectedVertex();
-            logger.debug("Registered typed DEL. Vertex: {}", vertex);
-            if (vertex != null) {
+            final Set<Object> vertices = _graphContext.getSelectedVertices();
+            logger.debug("Registered typed DEL. Vertices: {}", vertices);
+            if ((vertices != null) && (!vertices.isEmpty())) {
```
LosD: isEmpty() seems unneeded (though harmless)

LosD:
Looks good, but see code comments.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphKeyListener.java
tomaszguzialek: I also thought so, but see message of commit no. 2:  Checking if selected vertices set is empty (not only null) :).

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphContext.java
tomaszguzialek: Does not seem so, removed it now.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphKeyListener.java
LosD: But... But... The for loop should not... What? *confused*

LosD:
I have no idea why an empty set would be a problem, but no matter what, the check is not a problem. Merging!

PR #313, opened 2015-03-05
Adds a parent classloader to the jarFileClassLoader

LosD:
Very incomplete, but at least loading gets a little further.

Please do not merge yet.


LosD:
I'm retracting this, as I do not believe it will ever properly work using the current design of CompoundClassLoaders.

PR #314, opened 2015-03-06
Issue #287: Relaxing automatic input columns selecting condition.

### LINKED ISSUE #287: Whitespace trimmer issue
We add WhiteSpaceTrimmer analyzer to canvas and then from the dialog we choose "Write to Csv" option. The selected column in whitespace trimmer is not automatically mapped in the "Write CSV" dialog.

tomaszguzialek:
Fixes #287

Not entirely sure why the condition size == 0 was there. All of the writers (Staging table, Excel, CSV...) did not pass this condition.

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/actions/DisplayOutputWritersForTransformedDataActionListener.java
+++ b/desktop/api/src/main/java/org/datacleaner/actions/DisplayOutputWritersForTransformedDataActionListener.java
@@ -58,13 +58,11 @@ public class DisplayOutputWritersForTransformedDataActionListener extends Displa
                     _transformerJobBuilder.getDescriptor());
         }
 
-        if (componentBuilder.getDescriptor().getConfiguredPropertiesForInput().size() == 1) {
-            List<InputColumn<?>> inputColumns = _transformerJobBuilder.getInputColumns();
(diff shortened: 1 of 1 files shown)


LosD:
I don't believe the condition is there to protect against zero (i is size == 1), but against more than one set of properties. I remember discussing this with @kaspersorensen (in regards of using _Link to..._ from one component to another), and this is different, but similar, so I'd wait for him.

tomaszguzialek:
Yes, I meant size == 1. 

I imagine, we set the input columns automatically if this is the only setting that the component needs from the user to run properly. However, Write to CSV has "input columns" and "sort column" which is optional. I guess, we can always set the input columns by default and if there is something more to configure - a red sign and a message in status bar is shown.

Just to be sure, let's wait for @kaspersorensen .

kaspersorensen:
My only question is: Does it work as we want it to? :) I put the size == 1 condition there to make sure it was applied in a controlled way ... But if we're OK with just delegating the "column assignment" to the underlying builder, then I guess that's also OK if we see it working like we expect it to (for now at least - or else we will get new issues, but we can tackle those as we go).

tomaszguzialek:
IMO now it works as users expect it. Even if there are more things to be configured (size > 1), we should still provide a default column selection for the user and require him/her to configure only the stuff that we cannot infer.

PR #316, opened 2015-03-09
Feature/315 better example datastore

### LINKED ISSUE #315: Better example datastore.
Issue #204 talks about introducing better job examples to DC 4.0. We should also have a better example datastore, that the jobs can run on.

tomaszguzialek:
Fixes #315.

Introduces a new example datastore ("customers.csv"). The file has 5750 records (5000 unique records + 750 exact duplicates) from 5 countries: US, UK, DE, NL and DK. Given name and last names are taken at random from a certain pool. Addresses are real (postal code, city and country are matching the street), but are not associated with a real person for privacy reasons.

Questions:
1) Are exact duplicates good enough for now?
2) Also, shall we remove the contactdata.txt example datastore as it is only a couple of records? Maybe "customers.csv" should become a newer better version of contactdata.txt file?


desktop/ui/src/main/java/org/datacleaner/user/DemoConfiguration.java
```diff
@@ -29,6 +29,7 @@
 public class DemoConfiguration {
 
     public static final String DATASTORE_FILE_CONTACTDATA = "datastores/contactdata.txt";
```
kaspersorensen: Let's remove the old "contact data" datastore now that we have a much better one!

tomaszguzialek:
Ad. 2: Just removed "contactdata.txt" from example datastrores

What about question no. 1?

kaspersorensen:
Regarding 1: No I think we should also have some fuzz... I know it's kinda lame manual effort but we need some examples of spelling mistakes, different legal forms on the company name, missing information in some records etc. Maybe take a look at the few duplicates there are in orderdb-customers... Similar things we should be able to showcase.

kaspersorensen:
If it's any comfort I think that 750 duplicates is much more than needed. A typical database does not have +10% duplicates - more like a few percent. So from that perspective I would expect maybe 50-100 duplicate groups (most with 2 but some with 3-5 records) in a 5k dataset.

tomaszguzialek:
Reduced the number of duplicates to 2% (100 records) and manually introduced anomalies. Seems fair enough for showcasing the Training Tool.

kaspersorensen:
Good work, definately a step up. Some more small requests to make it even dirtier (I think this can all be done manually):

 * Please add a primary key (you can use "Generate UUID" in DC to do it if you want). Make a few example of non-unique keys afterwards.
 * Please add 2-3 examples of duplicate groups with 3-4 records
 * Please add 2 examples of duplicate groups with 5 or more exact same records (except primary key).
 * Please add a few more strangely spelled country values. Here are some ideas:
  * Danmark
  * Dänemark
  * DK
  * GER (germany but official 3-letter ISO is DEU)
  * Ελλάδα (greece)
 * Please add a few greek names and a few chinese names to have other alphabets represented (you can probably find some on wikipedia or so)

tomaszguzialek:
Done with the improvements from the last comment.

PR #317, opened 2015-03-10
Makes DCListCellRenderer the default renderer, and tweaks colors.

### LINKED ISSUE #306: Fix coloring of combobox popup items
A regular combobox in DataCleaner (usually created with DCComboBox) does not have a nice color scheme set. There is no hover effect so it's not easy to see what item you're clicking.

There is currently a class, DCListCellRenderer, which is used in many cases. But this should really be the default behaviour and not an option. It should probably just be a tweak to the LookAndFeelManager and then a deprecation of DCListCellRenderer.

LosD:
This also moves the DCListCellRenderer to API, as it needs to be
available there.

Fixes #306

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/widgets/DCListCellRenderer.java
+++ b/desktop/api/src/main/java/org/datacleaner/widgets/DCListCellRenderer.java
@@ -41,10 +41,10 @@ public class DCListCellRenderer extends DefaultListCellRenderer {
 
 		if (isSelected) {
 			result.setForeground(WidgetUtils.BG_COLOR_BLUE_DARK);
-			result.setBackground(WidgetUtils.BG_COLOR_BRIGHTEST);
+            result.setBackground(WidgetUtils.BG_COLOR_LESS_BRIGHT);
 		} else {
-			result.setForeground(WidgetUtils.BG_COLOR_LESS_BRIGHT);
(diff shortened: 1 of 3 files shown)


desktop/api/src/main/java/org/datacleaner/widgets/DCListCellRenderer.java
```diff
@@ -41,10 +41,10 @@ public Component getListCellRendererComponent(JList<?> list, Object value, int i
 
 		if (isSelected) {
 			result.setForeground(WidgetUtils.BG_COLOR_BLUE_DARK);
-			result.setBackground(WidgetUtils.BG_COLOR_BRIGHTEST);
+            result.setBackground(WidgetUtils.BG_COLOR_LESS_BRIGHT);
 		} else {
-			result.setForeground(WidgetUtils.BG_COLOR_LESS_BRIGHT);
-			result.setBackground(WidgetUtils.BG_COLOR_DARKEST);
+			result.setForeground(WidgetUtils.BG_COLOR_DARKEST);
```
kaspersorensen: strange that there is now a mix of tabs and spaces. But kinda minor.

kaspersorensen:
LGTM. If you feel like it fix the space/tab thingy, or else just merge :)

LosD:
Spaces fixed, waiting for Travis.

PR #319, opened 2015-03-10
When processing rows multi-threaded, sometimes threads can be not utilized.

jakubneubauer:
When the main thread cannot submit task to threadpool (because of the full task queue),  it processes a row itself (CallerRunsPolicy). But if the processing takes a long time, and the other processing threads become ready for new row, they are uselessly waiting for the main thread to feed them. Fixed by "blocking" policy.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/concurrent/MultiThreadedTaskRunner.java
+++ b/engine/core/src/main/java/org/datacleaner/job/concurrent/MultiThreadedTaskRunner.java
@@ -52,20 +52,22 @@ public final class MultiThreadedTaskRunner implements TaskRunner {
     public MultiThreadedTaskRunner(int numThreads) {
         _numThreads = numThreads;
 
-        // if all threads are busy, newly submitted tasks will by run by caller
+        // if all threads are busy, newly submitted tasks will be run by caller
         final ThreadPoolExecutor.CallerRunsPolicy rejectionHandler = new ThreadPoolExecutor.CallerRunsPolicy();
 
-        // there will be a minimum task capacity of 20, and preferably
-        // numThreads * 3 (to avoid blocking buffer behaviour)
-        final int taskCapacity = Math.max(20, numThreads * 3);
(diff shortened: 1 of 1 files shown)


kaspersorensen:
I'm wondering if there are other ways we can fix this that does not involve a hack. For instance:

 * Increase the size of the queue (a lot - for instance to size 1000 or maybe even 5k or so) which would effectively alleviate the issue. In terms of memory I would still not be worried - tasks have a rather low memory footprint (typically one record's worth of data plus a bit of extra info).
 * Check if there are other rejection policies available. Maybe there is one that fits the bill already?

jakubneubauer:
Some another proposals are discussed here: http://stackoverflow.com/q/2001086/155708
If my hack seems too dirty to you, I would use the solution with Semaphore as second option, to guard the tasks insertion in the queue to not exceed the queue size.

I would prefer increasing the queue from another reason, but not as solution to this issue. When involving I/O we should count with bigger lags. Just an example calculation: imagine lag of 20 seconds, in comparison to normal duration of 1 millisecond (nothing crazy I think). Then we would need 20.000 buffer just for one worker thread to prevent its "starvation".

Anyway, the queue size increase could be useful in cases of input rows fetching lags. For example, when fetching data from DB and there is a delay, it is useful to have some data in advance. But this is another issue.

kaspersorensen:
I agree to your points. Was just hoping that there was a built-in way of doing it this way, for instance with a rejection handler. I guess that's not the case...

Did you verify that this change works in your case? And tests are passing etc...? If so I suggest you also increase the "20" constant queue size to 1000. That's a conservative increase for now because your 20 second example is really in the extreme case too I think (I know you're doing a DNS check and that will take time for sure, but it's unusual that even remote calls takes more than a few seconds).

jakubneubauer:
Yes, there is no built-in way in Java for ThreadPoolExecutor to block the task addition when the queue is full. The solution with semaphore would need to wrap tasks with another task to release semaphore, causing a little overhead. So let's keep the proposed solution.

PR #320, opened 2015-03-10
 add excel extension to filename

### LINKED ISSUE #318: Add the extension when writing an excel file.
When writing to an excel file we need to add the excel extension if the user does not do it. 
I usually forgot to write the extension because I assume the "program" will do it for me.

ClaudiaPHI:
Fixes  #318 
There was no setter so I throw an exception instead.


kaspersorensen:
OK with me.

PR #328, opened 2015-03-13
Bug/issue #1599

drexler42:
Fix in ExtensionClassLoader. This allows saved DC job runs to be opened.


kaspersorensen:
Looking at this - is there now any longer a reason that we even have ExtensionClassLoader? I mean, couldn't we just use regular URLClassLoader (preferably just by adding the parent classloader instead of the null) ?

kaspersorensen:
Looks good!

PR #329, opened 2015-03-13
Fixes #322

### LINKED ISSUE #322: ColumnMeaning detection is a bit too eager
The detection of ColumnMeaning sometimes has false positives, i.e. anything ending in ID becomes a key column, and "phone no" becomes _Street Number_

kaspersorensen:
Fixes the "too eager column meaning mapping". See #322.

### DIFF
--- a/api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
+++ b/api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
@@ -35,7 +35,7 @@ public enum ColumnMeaning implements HasName, HasAliases {
 
     OTHER("Disregard", "Other", "Nothing"),
 
-    KEY_PRIMARY("Primary Key", "Record ID", "ID", "Identifier", "key", "PKID", "Record key"),
+    KEY_PRIMARY("Primary Key", "ID", "Record ID", "Identifier", "key", "PKID", "Record key"),
 
     KEY_FOREIGN("Foreign Key", "FK"),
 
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
```diff
@@ -35,7 +35,7 @@
 
     OTHER("Disregard", "Other", "Nothing"),
 
-    KEY_PRIMARY("Primary Key", "Record ID", "ID", "Identifier", "key", "PKID", "Record key"),
+    KEY_PRIMARY("Primary Key", "ID", "Record ID", "Identifier", "key", "PKID", "Record key"),
```
LosD: Why the switch-around? Optimization? (not a problem, just curious)

kaspersorensen: Hmm coincidence I think. I first tried another approach but changed my mind. Think this is a left-over of that attempt :-)

api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
```diff
@@ -63,8 +63,8 @@
 
     COMPANY_REGISTRATION_NUMBER("Company registration number", "Reg number"),
 
-    COMPANY_NAME("Company name", "Company", "Organization", "Business", "Organisation", "Employer", "Firm",
-            "Workplace", "Works at", "Vendor", "Supplier", "Customer", "Customer name"),
+    COMPANY_NAME("Company name", "Company", "Organization", "Business", "Organisation name", "Employer", "Firm",
+            "Workplace", "Works at", "Vendor name", "Supplier name", "Customer name"),
```
LosD: Now that we don't tokenize column meanings isn't it a problem that _Supplier_ and _Vendor_ are only there in combination with name?

LosD:
Single worry, but looks good. We might want to make sure that other uses of the matcher isn't relying on the previous behavior (esp. tokenization).

api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
kaspersorensen: Yes, I'll add that.

PR #330, opened 2015-03-16
Fix pop-under issue

### LINKED ISSUE #325: Results and component dialogs may pop under instead of over main window
Sometimes, the result windows and component dialogs open below the main window. I think the frequency is maybe every 5th or 6th time on my system.

LosD:
This should fix the ResultWindows, at least simple ones. We've decided to leave component dialogs alone for now, as the pop-under is much less annoying, especially after the change that does not open new dialogs on an extra click. Also, it doesn't seem to affect all component dialogs (i.e. I couldn't trigger the String Analyzer at all), so we might want to do them on a case-by-case basis.

Fixes #325


kaspersorensen:
LGTM!

PR #331, opened 2015-03-16
Issue #204: Added job title standardization/analytics job

### LINKED ISSUE #204: Better example jobs that utilize the graph UI's power

kaspersorensen:
Fixes #204.

The new job does many things that was needed to demonstrate - first and foremost to do synonym replacement and to apply filtering.


kaspersorensen:
image

desktop/ui/src/main/resources/datacleaner-home/jobs/OrderDB Customers Duplicate detection.analysis.xml
```diff
@@ -1,64 +1,65 @@
...
+	<analysis>
+		<analyzer>
+			<descriptor ref="Duplicate Detection" />
+			<metadata-properties>
+				<property name="CoordinatesY">221</property>
+				<property name="CoordinatesX">253</property>
+			</metadata-properties>
+			<properties>
+				<property name="Model"
+					value="C:\Users\kaspers\Desktop\orderdb_dupe_model.dedupmodel.xml" />
```
tomaszguzialek: Is the hardcoded path needed here or just forgotten?

tomaszguzialek:
See the question about the hardcoded path - the rest looks OK. Cannot be merged automatically though and the build also failed partly.

PR #335, opened 2015-03-18
Filters should show the popup on mouse release, not press.

### LINKED ISSUE #333: Shift key dragging on canvas no longer works
I think this has to do with the META key support that we added for Mac...

Previously we could draw links in the graph by holding down the shift key and dragging an arrow. Now the click provokes the opening of a configuration dialog. Furthermore if a filter is clicked, the menu with it's filter outcomes is shown, which does not make sense at all (that menu should be shown by the target of the arrow, not by the component itself).

LosD:
Fixes #333

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphMouseListener.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphMouseListener.java
@@ -233,6 +233,16 @@ public class JobGraphMouseListener extends MouseAdapter implements GraphMouseLis
         logger.debug("Graph released");
 
         if (isLeftClick(me)) {
+            if(v instanceof ComponentBuilder){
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphMouseListener.java
```diff
@@ -233,6 +233,16 @@ public void graphReleased(Object v, MouseEvent me) {
         logger.debug("Graph released");
 
         if (isLeftClick(me)) {
+            if(v instanceof ComponentBuilder){
```
kaspersorensen: Missing space :-O

kaspersorensen:
Looks and feels good. Although there is a horrible, horrible missing space issue I'll go ahead and merge it ;-)

PR #337, opened 2015-03-19
Escape CoalesceUnitConverter, and make de-/serialization symmetric.

### LINKED ISSUE #336: Comma in column name causes coalesce transformation to fail
We are not escaping the comma in the column name while serializing it for the coalesce transformation. It then does not recognize it.

LosD:
Fixes #336

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/convert/SerializationStringEscaper.java
+++ b/engine/core/src/main/java/org/datacleaner/util/convert/SerializationStringEscaper.java
@@ -1,0 +1,36 @@
+package org.datacleaner.util.convert;
+
+import org.apache.commons.lang.ArrayUtils;
+
+public class SerializationStringEscaper {
+    private static final String[][] ESCAPE_MAPPING = { { "&amp;", "&" }, { "&#91;", "[" }, { "&#93;", "]" },
+            { "&#44;", "," }, { "&lt;", "<" }, { "&gt;", ">" }, { "&quot;", "\"" }, { "&copy;", "\u00a9" },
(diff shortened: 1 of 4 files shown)


engine/core/src/main/java/org/datacleaner/util/convert/SerializationStringEscaper.java
```diff
@@ -0,0 +1,55 @@
...
+public class SerializationStringEscaper {
+    private static final String[][] ESCAPE_MAPPING = { { "&amp;", "&" }, { "&#91;", "[" }, { "&#93;", "]" },
+            { "&#44;", "," }, { "&lt;", "<" }, { "&gt;", ">" }, { "&quot;", "\"" }, { "&copy;", "\u00a9" },
+            { "&reg;", "\u00ae" }, { "&euro;", "\u20a0" } };
+    private static final String[][] ESCAPE_MAPPING_REVERSED;
+    
+    // Ughhh.
+    static {
+        ESCAPE_MAPPING_REVERSED = (String[][]) ArrayUtils.clone(ESCAPE_MAPPING);
+        ArrayUtils.reverse(ESCAPE_MAPPING_REVERSED);
```
kaspersorensen: I guess if you really have a "Ughhh" feeling we COULD also simply adjust the for-loop to run through the array backwards.

kaspersorensen:
Looks good to me. I will let you decide whether we merge as-is or if you want to do something about the "Ughhh" comment ;-)

LosD:
Nah, that's even more ughhh :) Let's merge!

Issue #341, opened 2015-03-19
New transformer: Remove dictionary matches


kaspersorensen:
I am considering a transformer called "Remove dictionary matches" that would fit this scenario very well:

You have certain words (registered in a dictionary) that are unwanted for various purposes. The transformer should remove these from a string.

It could be for instance (somewhat temporal) adjectives in a job title such as "Junior", "Senior", "Principal", "Lead" and so on. For matching reasons you might very well want to remove these words from a job title because then matching is closer. Another example could be to remove legal forms from company names or whatever.

PR #342, opened 2015-03-19
Issue #341: Initial work done (and works for single word tokens).

kaspersorensen:
Still not working for multi-word tokens.

Do not merge yet, we should figure out if we want to make it work for multiple tokens (complicated and maybe less important given the described scenario in #341?)


components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -0,0 +1,94 @@
...
+    public static final String PROPERTY_DICTIONARY = "Dictionary";
+    public static final String PROPERTY_COLUMN = "Column";
+
+    @Configured(value = PROPERTY_DICTIONARY)
+    Dictionary _dictionary;
+
+    @Configured(value = PROPERTY_COLUMN)
+    InputColumn<String> _column;
+
+    private final Splitter SPLITTER = Splitter.on(' ').omitEmptyStrings();
```
LosD: Shouldn't this be ```Splitter.on(CharMatcher.WHITESPACE)```?

kaspersorensen: Good idea

PR #349, opened 2015-03-24
Feature/346 extension csv file

### LINKED ISSUE #346: add .csv extension when writing to a CSV file
Consider automatically adding the csv extension or make the user add  the extension to the file name.

ClaudiaPHI:
Fixes #346


LosD:
Are we forcing the users to write .csv by themselves? This seems completely wrong to me.

ClaudiaPHI:
I always forget to do that and i can't find the file afterwards. For me as a windows user, I have the training that I do not need to add the extension when I create a file. When I create a csv or excel file, I expect (I do not think about it) that the system will add the extension for me.

kaspersorensen:
I also feel this is kinda wrong altogether. The file property says:

```
@FileProperty(accessMode = FileAccessMode.SAVE, extension = { "csv", "tsv", "txt", "dat" })
```

This information should potentially be leveraged by the user interface to make the addition of the extension. But breaking functionality just because the user maybe insists to write it to a different file is lame. So in my opinion we need a UI improvement for the property widget of files - not a strict validation in this particular component.

ClaudiaPHI:
ok.

PR #352, opened 2015-03-24
Changed default logging directory to ${user.home}/.datacleaner/log

### LINKED ISSUE #351: Don't log to the application directory
Change our default log4j config to log to userhome/.datacleaner/log

Related to #344.

tomaszguzialek:
instead of ./log

Fixes #351 

Changed it to ${user.home}/.datacleaner/log/datacleaner.log however it would be perfect to have ${user.home}/.datacleaner/DATACLEANER.VERSION/log/datacleaner.log, but the version of DC is not available from System properties. 

JNLP already logged to user directory, but for consistency I changed it to .datacleaner subdirectory.

### DIFF
--- a/desktop/ui/src/main/resources/org/datacleaner/log4j-jnlp.xml
+++ b/desktop/ui/src/main/resources/org/datacleaner/log4j-jnlp.xml
@@ -3,7 +3,7 @@
 <log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
 
 	<appender name="fileAppender" class="org.apache.log4j.FileAppender">
-		<param name="File" value="${user.home}/datacleaner.log" />
+		<param name="File" value="${user.home}/.datacleaner/datacleaner.log" />
 		<param name="Append" value="true" />
 		<layout class="org.apache.log4j.PatternLayout">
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/resources/org/datacleaner/log4j-jnlp.xml
```diff
@@ -3,7 +3,7 @@
 <log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
 
 	<appender name="fileAppender" class="org.apache.log4j.FileAppender">
-		<param name="File" value="${user.home}/datacleaner.log" />
+		<param name="File" value="${user.home}/.datacleaner/datacleaner.log" />
```
kaspersorensen: missing /log here I think?

kaspersorensen:
Looks fine, but check the comment :-)

desktop/ui/src/main/resources/org/datacleaner/log4j-jnlp.xml
tomaszguzialek: Good catch :). Fixed.

PR #357, opened 2015-03-31
Issue #356: Lowercasing the extension before comparison with the

### LINKED ISSUE #356: Dragging CSV file onto drop zone leads to NullPointerException
When I drag any of two attached CSV files on the DropZone in DC4  pro edition, I'll get a NullPointerException.

This happens if the CSV is not yet known as a datastore in DataCleaner. Once it is added, it work OK.

The CSV files seem to be nothing special, but find them attached anyway.

tomaszguzialek:
supported ones.

Fixes #356


tomaszguzialek:
Also showing an "Illegal file type" pop-up when an unsupported file is chosen in the file dialog (it caused an NPE; dropping does not react at all for unknown files). An unsupported file is for example a JPG or no extension at all (despite being a properly formatted CSV file wrt the contents).

kaspersorensen:
Sounds good. Wording wise let's only call it 'unrecognized' file type instead of illegal... Illegal Sounds like we found it to be wrong, but could be we just don't support it.

tomaszguzialek:
I commited this change on the wrong branch. The correct PR is #359 .

PR #359, opened 2015-04-01
Bug/356 uppercase extensions

### LINKED ISSUE #356: Dragging CSV file onto drop zone leads to NullPointerException

tomaszguzialek:
Fixes #356

Also showing an "Unrecognized file type" pop-up when an unsupported file is chosen in the file dialog (it caused an NPE; dropping does not react at all for unknown files). An unsupported file is for example a JPG or no extension at all (despite being a properly formatted CSV file wrt the contents).

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/util/DatastoreCreationUtil.java
+++ b/desktop/ui/src/main/java/org/datacleaner/util/DatastoreCreationUtil.java
@@ -57,7 +57,7 @@ public class DatastoreCreationUtil {
             final String extension = FilenameUtils.getExtension(file.getName());
 
             for (FileDatastoreEnum datastoreType : EnumSet.allOf(FileDatastoreEnum.class)) {
-                if (datastoreType._extensions.contains(extension)) {
+                if (datastoreType._extensions.contains(extension.toLowerCase())) {
                     return datastoreType;
                 }
             }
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/util/DatastoreCreationUtil.java
```diff
@@ -57,7 +57,7 @@ protected static FileDatastoreEnum getDatastoreTypeFromFile(File file) {
             final String extension = FilenameUtils.getExtension(file.getName());
 
             for (FileDatastoreEnum datastoreType : EnumSet.allOf(FileDatastoreEnum.class)) {
-                if (datastoreType._extensions.contains(extension)) {
+                if (datastoreType._extensions.contains(extension.toLowerCase())) {
```
kaspersorensen: Are there any risks that "extension" variable is null? I could imagine that FilenameUtils would return null if the filename does not have any dot in it.

tomaszguzialek: FilenameUtils.getExtension() returns an empty string instead of a null.

kaspersorensen:
Ok, looks good to me then.

Issue #363, opened 2015-04-06
Refactoring of "configuration" layer


kaspersorensen:
A couple of items needs to be addressed holistically in the current "configuration" layer:

 * The central class name AnalyzerBeansConfiguration (from now on just dubbed 'the configuration class') obviously has a bad name since now AnalyzerBeans has been merged into DataCleaner. We should deprecate this interface name.
 * The configuration class should be split into two: The user's own configuration (datastores etc.) and the environment, which can be shared amongst users and changes less frequently (the task runner, descriptor provider etc.)
 * The desktop and server app have two different ways to identify the "home" of the user - DataCleanerHome (desktop) and repository (server). We should bridge this gap or else dealing with issues like #354 is bound to be too difficult.

PR #364, opened 2015-04-06
Feature/363 configuration layer

kaspersorensen:
Do not merge yet - This PR is so far only here to remind us of ongoing work in the branch and to get updates from Travis.

This is a rework of the configuration classes of DataCleaner - see #363


kaspersorensen:
Note to self: Make sure to solve the resource type handler... It should be able to load resources from the new home directory and then we can avoid a lot of hacks in monitor to make resource loading work per-tenant.

engine/core/src/main/java/org/datacleaner/configuration/DataCleanerConfigurationImpl.java
```diff
@@ -0,0 +1,143 @@
...
+    public static ReferenceDataCatalog defaultReferenceDataCatalog() {
+        return new ReferenceDataCatalogImpl();
+    }
+
+    public static DatastoreCatalog defaultDatastoreCatalog() {
+        return new DatastoreCatalogImpl();
+    }
+
+    public static RepositoryFolder defaultHomeFolder() {
+        return new FileRepository(".");
```
tomaszguzialek: So the default is still application directory? Did not decide to kill this behaviour?

kaspersorensen: It is still the default if you just instantiate a new DataCleanerConfigurationImpl. This is still fine behaviour for casual use of the class (say, in a unittest or so) but in the two "real" environments ("desktop" and "monitor") we set another home folder that fits those situations.

engine/env/cluster/src/test/java/org/datacleaner/cluster/DistributedAnalysisRunnerTest.java
```diff
@@ -118,7 +118,7 @@ public void testErrorHandlingFourSlaves() throws Exception {
 
     public void testUndistributableAnalyzer() throws Exception {
         final Datastore datastore = TestHelper.createSampleDatabaseDatastore("orderdb");
-        final AnalyzerBeansConfiguration configuration = new AnalyzerBeansConfigurationImpl()
+        final DataCleanerConfiguration configuration = new AnalyzerBeansConfigurationImpl()
```
tomaszguzialek: Why are we using the old implementation in these tests?

tomaszguzialek:
Just asked two questions (one already answered), the rest looks good... Although the diff was unhuman ;).

cat_toilet_paper

engine/env/cluster/src/test/java/org/datacleaner/cluster/DistributedAnalysisRunnerTest.java
kaspersorensen: I guess it's one of many cases where the old stuff is still in use. I didn't get around to fix ALL the deprecation warnings.

PR #366, opened 2015-04-08
Feature/354 relative file resolving

### LINKED ISSUE #354: Resolve @Configured File properties relative to job file location
Currently if a job contains a component with a Configured File property, this property value is resolved as an absolute file path or relative to the working directory of the application.

This is a bit cumbersome because it makes management of jobs a lot more difficult - typically you'd want to be able to move jobs around and if there are any accessory files with it, move them along. Because of the file paths this does not currently work. So make it so :-)

kaspersorensen:
Fixes #354

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleResourcePropertyWidget.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleResourcePropertyWidget.java
@@ -29,6 +29,7 @@ import javax.swing.filechooser.FileFilter;
 
 import org.datacleaner.api.FileProperty;
 import org.datacleaner.api.FileProperty.FileAccessMode;
+import org.datacleaner.configuration.DataCleanerConfiguration;
 import org.datacleaner.descriptors.ConfiguredPropertyDescriptor;
 import org.datacleaner.job.builder.ComponentBuilder;
 import org.datacleaner.util.StringUtils;
@@ -180,7 +181,8 @@ public final class SingleResourcePropertyWidget extends AbstractPropertyWidget<R
      * @return
      */
     protected ResourceConverter getResourceConverter() {
-        return new ResourceConverter();
+        DataCleanerConfiguration configuration = getAnalysisJobBuilder().getConfiguration();
+        return new ResourceConverter(configuration);
     }
 
     @Override
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleResourcePropertyWidget.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleResourcePropertyWidget.java
@@ -181,7 +181,7 @@ public final class SingleResourcePropertyWidget extends AbstractPropertyWidget<R
      * @return
      */
     protected ResourceConverter getResourceConverter() {
-        DataCleanerConfiguration configuration = getAnalysisJobBuilder().getConfiguration();
+        final DataCleanerConfiguration configuration = getAnalysisJobBuilder().getConfiguration();
         return new ResourceConverter(configuration);
     }
 
--- a/desktop/ui/src/test/java/org/datacleaner/test/full/scenarios/TokenizerAndValueDistributionTest.java
+++ b/desktop/ui/src/test/java/org/datacleaner/test/full/scenarios/TokenizerAndValueDistributionTest.java
@@ -33,8 +33,10 @@ import org.datacleaner.api.InputColumn;
 import org.datacleaner.beans.transform.TokenizerTransformer;
 import org.datacleaner.beans.valuedist.ValueDistributionAnalyzer;
 import org.datacleaner.beans.valuedist.ValueDistributionAnalyzerResult;
-import org.datacleaner.configuration.AnalyzerBeansConfigurationImpl;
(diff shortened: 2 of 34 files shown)


engine/core/src/main/java/org/datacleaner/util/FileResolver.java
```diff
@@ -0,0 +1,116 @@
...
+            absolute = false;
+        } else {
+            absolute = file.isAbsolute();
+        }
+
+        path = StringUtils.replaceAll(path, "\\", "/");
+        if (!absolute) {
+            // some normalization (because filenames are often used to compare
+            // datastores)
+            if (path.startsWith("/")) {
```
LosD: This probably doesn't hurt, but shouldn't the comparison be on the actual paths, instead of on raw strings?

engine/core/src/main/java/org/datacleaner/util/convert/ResourceConverter.java
```diff
@@ -94,9 +104,24 @@ public String getScheme() {
 
     /**
      * Constructs a {@link ResourceConverter} using a default set of handlers
+     * 
+     * @deprecated use another constructor instead
```
LosD: A little bit uninformative :)

engine/core/src/main/java/org/datacleaner/util/FileResolver.java
kaspersorensen: The point here is to make the strings deterministic. No matter if you refer to a file relatively or absolutely then the toPath(File) method should return the same path (relative if possible, absolute as fallback).

engine/core/src/main/java/org/datacleaner/util/convert/ResourceConverter.java
kaspersorensen: Well there are 4 other constructors. It seems needless to mention them all - the message is to pick whatever constructor is feasible, just not this one :-)

desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleResourcePropertyWidget.java
```diff
@@ -180,7 +181,8 @@ public void onItemSelected(String item) {
      * @return
      */
     protected ResourceConverter getResourceConverter() {
-        return new ResourceConverter();
+        DataCleanerConfiguration configuration = getAnalysisJobBuilder().getConfiguration();
```
ClaudiaPHI: final ?

desktop/ui/src/test/java/org/datacleaner/test/full/scenarios/TokenizerAndValueDistributionTest.java
```diff
@@ -52,9 +54,10 @@
 public class TokenizerAndValueDistributionTest extends TestCase {
 
     public void testScenario() throws Throwable {
-        TaskRunner taskRunner = new MultiThreadedTaskRunner(5);
-
-        DataCleanerConfiguration configuration = new AnalyzerBeansConfigurationImpl().replace(taskRunner);
+        final TaskRunner taskRunner = new MultiThreadedTaskRunner(5);
+        final DataCleanerEnvironment environment = new DataCleanerEnvironmentImpl().withTaskRunner(taskRunner);
+        
+        DataCleanerConfiguration configuration = new DataCleanerConfigurationImpl().withEnvironment(environment);
```
ClaudiaPHI: final ?

kaspersorensen: Not really a change from before, but OK.

engine/core/src/main/java/org/datacleaner/util/FileResolver.java
LosD: I think what I'm getting at, is that you should never care. The system should determine if it is the same file, not us, and certainly not from string comparison, if at all avoidable.

Same path (when in C:\Users\Dennis):
- C:\User\Dennis\file.txt
- C:\User\Dennis\.\file.txt
- C:\User\Dennis\test\..\file.txt
- file.txt

LosD: Yeah, there's [Files.isSameFile()] (http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#isSameFile(java.nio.file.Path,%20java.nio.file.Path))

engine/core/src/main/java/org/datacleaner/util/convert/ResourceConverter.java
LosD: I general I would expect how to get the "most equivalent" one in the simplest way, in this case ```new ResourceConverter(new DataCleanerConfigurationImpl())```, but yeah, it's should be easy to pick one :)

engine/core/src/main/java/org/datacleaner/util/FileResolver.java
LosD: Okay, seems like FilenameUtils.normalize() fixes the paths, but I'm still no fan of using Strings for... Well, anything where there's a proper way of doing it. :)

LosD:
Very minor stuff (okay, maybe not the filename stuff, but after discussing offline, it seems we have no simple way around it), but in general looks good! :)

PR #369, opened 2015-04-09
Added more connection parameters for making a better oracle connection

### LINKED ISSUE #355: The Oracle datastore does not allow to specify connections with service name or easy connect string
# Actual
The Oracle datastore type tab "Oracle connection" provides parameter fields for 'Hostname", "Port", "System ID (SID)", "Username" and "Password".

Oracle versions can also connect using the service name. However, the Oracle data store only works when you specify the SID.

Oracle also has the "easy connect string", which is also not supported.

#Expected
The Oracle datastore should allow connections by specifying the service name and also by specifying the easy connect string.

Oracle documentation on the easy connect string:
http://www.oracle.com/technetwork/database/enterprise-edition/oraclenetservices-neteasyconnect-133058.pdf

kaspersorensen:
Fixes #355 

Screenshot:

image

### DIFF
--- a/components/standardizers/src/main/java/org/datacleaner/util/NamedPattern.java
+++ b/components/standardizers/src/main/java/org/datacleaner/util/NamedPattern.java
@@ -48,9 +48,10 @@ public class NamedPattern<E extends Enum<E>> {
      */
     public static final String DEFAULT_GROUP_LITERAL = "([\\p{Lu}\\p{Ll}]+)";
 
-    private EnumMap<E, Integer> groupIndexes;
-    private Pattern pattern;
-    private Class<E> groupEnum;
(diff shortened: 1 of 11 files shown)


drexler42:
I think the user does not need to enter service name and SID. The user typically uses the Service Name or the SID. This UI does not clearly reflect that yet. The Oracle SQL Developer tool uses a radio button to let the user choose between entering the service name or the SID (see screenshot).


2015-04-09 16_40_58-new _ select database connection

kaspersorensen:
Good point and thx for the screenshot. Will see if I can improve.

kaspersorensen:
Updated: 

image

ClaudiaPHI:
Looks good to me

drexler42:
Looks good.

PR #371, opened 2015-04-09
Issue 367 - DataCleanerHomeFolder

### LINKED ISSUE #367: Make a distinct interface/class for the DC home folder
As of #363 we now have the lovely new method: DataCleanerConfiguration.getHomeFolder(). It currently returns a type RepositoryFolder.

But on second thought I think we should have a separate interface / return type for this. I would call this DataCleanerHomeFolder (to avoid name confusion with desktop-ui's DataCleanerHome class). I have two reasons for wanting it:

 * Making it possible to inject the DataCleanerHomeFolder into components etc. using just the type as the qualifier (generally much easier than having to use a qualifier).
 * The interface could contain method to get the home folder represented as various other types - RepositoryFolder, java.io.File, VFS FileObject. This would alleviates a lot of the current need for conversion between the different file subsystems that we have.

kaspersorensen:
Fixes #367


tomaszguzialek:
Looks good, merging.

PR #372, opened 2015-04-12
Issue #225: AnalyzerResults from non-analyzer (ALL) components

### LINKED ISSUE #225: Ability for non-analyzers to produce a result
It would be cool if Transformers and Filters could also produce AnalyzerResults (but OK, maybe that interface name needs to have a naming check then).

An example would be a DQ function like Name cleansing. Although it produces records, it could also produce a small report telling a summary of it's outcome and so on.

Technically I guess it should be possible just by letting every component optionally implement the HasAnalyzerResult interface.

kaspersorensen:
Fixes #225


kaspersorensen:
Now with a first real use - the TableLookupTransformer:

image

monitor/services/src/test/java/org/datacleaner/monitor/server/MonitorJobReaderTest.java
```diff
@@ -28,6 +28,10 @@
 import org.datacleaner.api.InputColumn;
 import org.datacleaner.configuration.DataCleanerConfiguration;
 import org.datacleaner.configuration.DataCleanerConfigurationImpl;
+<<<<<<< HEAD
+import org.datacleaner.configuration.DataCleanerEnvironment;
+=======
+>>>>>>> refs/heads/master
```
ClaudiaPHI: there is a conflict here .

kaspersorensen: Yup, I did see it, should be fixed now.

tomaszguzialek:
Code looks good, also checked out and verified that the TableLookupTransformer produces a nice pie chart. Just wondering: does it finally work in a distributed runner?

kaspersorensen:
I think it will not ... At least it is not an item that I specifically looked for. But I guess we could use the same approach as with analyzers - to require a custom AnalyzerResultReducer class.

tomaszguzialek:
Fair enough, merging!

Issue #343, opened 2015-03-22
Automatic component reference documentation generation


kaspersorensen:
An automatic "Component API documentation" tool for DataCleaner. Either fully integrated into the desktop app or as a tool which can generate HTML documentation (a la javadoc). Or both even.

I imagine that in the tree of the desktop app one could click e.g. the Pattern Finder and then be brought to a screen/page that contains documentation of this component. The documentation could contain many elements, all generated based on the ComponentDescriptor and related classes:

 * Name
 * Type of component (transformer, filter, analyzer)
 * Description
 * Icon
 * Execution details:
  * Is it concurrent?
  * Can it be distributed?
 * Each property can be listed, including details like:
  * Name
  * Type of property (InputColumn, String, Number, Enum etc.)
  * Description
  * Any constraints set by e.g. StringProperty, NumberProperty etc. annotations.
  * Is it mapped to any other property (via MappedProperty annotation)
 * If the component has a result, what is the result type
  * And if the result type exposes any metrics, describe each metric name and type.

ClaudiaPHI:
![image](https://cloud.githubusercontent.com/assets/6339923/7135371/61275044-e2a6-11e4-8333-6fe4cc9a8e47.png)

ClaudiaPHI:
![image](https://cloud.githubusercontent.com/assets/6339923/7271961/756ce248-e8e7-11e4-831b-878c47aa0a42.png)

PR #374, opened 2015-04-13
Feature/343 automatic reference documentation

ClaudiaPHI:
Refers to #343
I am not sure about some things so please comment.


desktop/ui/documentLoaderOutput.txt
```diff
@@ -0,0 +1,104 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
kaspersorensen: I doubt you want to have this file committed?

desktop/ui/pom.xml
```diff
@@ -257,6 +258,11 @@
 			<groupId>org.apache.httpcomponents</groupId>
 			<artifactId>httpmime</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.freemarker</groupId>
+			<artifactId>freemarker</artifactId>
+			<version>2.3.21</version>
```
kaspersorensen: Let's have some management of the version by making freemarker a managed dependency in the parent pom. Then we also should not specify the version in the monitor module(s).

desktop/ui/src/main/java/org/datacleaner/documentation/template/DocumentationLoader.java
```diff
@@ -0,0 +1,91 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import freemarker.template.Configuration;
+import freemarker.template.Template;
+import freemarker.template.TemplateException;
+
+public class DocumentationLoader {
+
+    private static final String FILENAME_TEMPLATE = "/src/main/resources/documentation_template.html";
```
kaspersorensen: This file will not be resolveable at runtime. You need to have it resolved as a resource on classpath, not a file in a directory.

desktop/ui/src/main/java/org/datacleaner/documentation/template/DocumentationLoader.java
```diff
@@ -0,0 +1,91 @@
...
+    static final String OUTPUT_FILENAME = "documentLoaderOutput.txt";
+
+    private Template _template;
+    private Map<String, Object> _data;
+    private static Logger logger = LoggerFactory.getLogger(DocumentationLoader.class);
```
kaspersorensen: 'final' on all the fields?

desktop/ui/src/main/java/org/datacleaner/documentation/template/DocumentationLoader.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    public void createDocumentation(ComponentDescriptor<?> componentdescriptor) {
+
+        try {
+            _data.put("componentname", componentdescriptor.getDisplayName());
+            _data.put("description", componentdescriptor.getDescription());
+            _data.put("icon", "loading icon"); // TODO - add an icon
+            _data.put("isDistributable", "" + componentdescriptor.isDistributable());
```
kaspersorensen: Instead of moving everything from ComponentDescriptor to a hashmap, why not just place the ComponentDescriptor into the model and then refer to it in the template markup?

kaspersorensen:
I think we need to be a bit more ambitious here :) Glad that we have something that works, but there's a lot of hardcoding here (output filenames etc. which seems like a pity. Keep the output flexible so that we can apply this class in a flexible manner.

If you look at the output it is pretty obvious that it is currently not user friendly to read.

 * The properties are layed out in a table, but in the UI you're used to presenting them in a list/vertical fashion. No reason to change that, especially considering that there's a lot more information about the properties we can add.
 * The property 'type' is a java class spec... Not really easy to understand. And if it's an array it has the cryptical [L syntax.
 * The properties are normally separated in two lists: Required and Optional. Should be preserved.
 * A less technical way to show the component would be nicer. For instance: Why prefix the description with "Description:" ... It only makes it feel like a specification document, not a humanly readable documentation.

ClaudiaPHI:
image

ClaudiaPHI:
The build passes now.

ClaudiaPHI:
What do you think on the styling ? 
image

kaspersorensen:
A few thing comes to the eye: The font is with serifs and certainly not the font we normally use in DC (Ubuntu).

The bullet list wrt. distributed seems strange. And I can think of many other aspects (and more to come in the future) that would be pluggable so I would not "ask a question" for each of them, but tag the documentation with the ones that apply. So for instance I would here have a "Distributed execution possible" tag. And to give examples of other tags: "Analyze category", "Concurrent execution", "Analyzer"/"Transformer","Filter" etc.

The properties are laid out in a table and I think that does not scale very well. Rather I would make a paragraph about every property. Each property would also have many tags that could apply. That's why I think the table does not work - the 4 columns cannot fit in all the aspects that we want to eventually include. Example of property details/tags could be "mapped to [property X]", "required". "Not negative" (number), "text/xml" (string mime type) etc... If you look at the DataCleaner-api package you can find many annotations that optionally describe this stuff.

ClaudiaPHI:
new work in progress

image

kaspersorensen:
Starting to look a lot better. My feeling is still that there are too many "static headers" like "property", "type" etc. Rather use h1, h2, h3 and css classes to signal what something is. We might add icons or stuff like that. Repetitive labels take up a lot of space and make it appear a lot more technical.

kaspersorensen:
Few remarks on the code: The filenames (output filename as well as stylesheet filename) are hardcoded and there's even a committed output file in the root of the UI module. We certainly don't want that.

ClaudiaPHI:
I think we are heading towards something very abstract without the "static" headers. It has to be understandable especially since it is not a table. A user would not want to spend to much time trying to understand the meaning of headers.

ClaudiaPHI:
Final look: 

image

kaspersorensen:
I'll pick up this tomorrow and work on the branch

desktop/ui/src/test/resources/documentCreatorBenchmarkTest.html
```diff
@@ -0,0 +1,265 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+<link rel="stylesheet" type="text/css" href=documentationReference.css />
```
kaspersorensen: This will not be resolveable except when running from the sources. Be aware that as soon as the code is running in production we will only have the packaged JAR file left and non of the resources that you put in the source tree.

desktop/ui/src/main/java/org/datacleaner/documentation/template/DocumentationCreator.java
```diff
@@ -0,0 +1,211 @@
...
+                    _data.put("properties", properties);
+                    Map<String, MappedProperty> mappedProperties = new HashMap<String, MappedProperty>();
+                    for (ConfiguredPropertyDescriptor property : properties) {
+                        final MappedProperty mappedProperty = property.getAnnotation(MappedProperty.class);
+                        if (mappedProperty != null) {
+                            mappedProperties.put(property.getName(), mappedProperty);
+                        }
+                    }
+                    _data.put("mappedproperties", mappedProperties);
+                }
```
kaspersorensen: More stuff we should have info on in the documentation:

 * The result time (if the component produces results)
 * The output column types (if the component is a transformer)
 * The filter outcomes (if the component is a filter)

kaspersorensen:
Examples:

image

image

image

kaspersorensen:
I am done for now ... The index still doesn't look pretty at all, but it's there. I meant to stop here because it gives us enough to start getting feedback (I think - what do you think?)..

You can generate documentation for your complete installation by running like this:

```
> java -cp datacleaner.jar org.datacleaner.documentation.Main
```

It will create a "documentation" folder in your datacleaner home. Run the index.html file from there...

kaspersorensen:
Note the build failed but not related to this PR - see #393 for details on the failure. Please review this code anyway.

desktop/ui/src/main/java/org/datacleaner/documentation/ComponentDocumentationBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+                /* Encode the image */
+                final String encodedImage = Base64.encodeBase64String(imageInByte);
+
+                /*
+                 * Atach the prefix that will make html <img> know how to decode
+                 * the image
+                 */
+                final String iconHtmlRepresentation = HTMLBASE64_PREFIX + encodedImage;
+
+                data.put("icon", iconHtmlRepresentation);
```
LosD: Very much nitpicking, but it would be nice to keep the icon generation separate in its own class.

LosD:
Missing importing Ubuntu font.

desktop/ui/src/main/java/org/datacleaner/documentation/IndexDocumentationBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+        writer.write("<ul>");
+        for (ComponentDescriptor<?> componentDescriptor : componentDescriptors) {
+            String filename = getFilename(componentDescriptor);
+            writer.write("<li><a href=\"");
+            writer.write(filename);
+            writer.write("\">");
+            writer.write(componentDescriptor.getDisplayName());
+            writer.write("</a></li>");
+        }
+        writer.write("</ul>");
```
LosD: Shouldn't the template engine help out here, or is it that too much?

LosD:
I don't know FreeMarker at all, but I'm surprised we have so much manual HTML in the Java code.

desktop/ui/src/main/java/org/datacleaner/documentation/IndexDocumentationBuilder.java
kaspersorensen: Yes but we don't have a template for the index at all yet. So this was just a kind of placeholder to have _something_... Eventually I think we need a much better index (if the index is the route to this documentation in general).

LosD:
Otherwise looks good! :)

kaspersorensen:
I agree there's a bit too much - mostly because of code duplication in getDescription() method ... will fix that and then merge.

LosD:
:+1:

PR #375, opened 2015-04-13
Issue #362: Added --version, -version and /version cmd line args

### LINKED ISSUE #362: Create a -version option
It would be good to add a command line option to DataCleaner that prints the version information and then exits, like the "mvn -version" command.

kaspersorensen:
Fixes #362 

Also made the new "distribution version" available in about dialog:

image


ClaudiaPHI:
It's ok with me!

PR #380, opened 2015-04-16
Properly identify columns of a table

### LINKED ISSUE #373: Adding columns to a (re)opened job are not represented in table dialog
A bit of a weird thing. Seems that when opening a job that already has some source columns defined, and more columns (from same table) is added, nothing happens.

For instance - open the job "US Customer STATE check.analysis.xml".
Then try to add the "phone" column to the existing job.
And double click the source table node in the graph.
It will still only have the previous columns defined:

image

It seems that from a input mapping perspecting it is possible to select the new phone column. But seems the table node is another Table instance of something like that. If I delete the table node then a NEW table node shows up, showing only the phone column...

kaspersorensen:
Fixes #373


tomaszguzialek:
Looks like a quick fix, so looks good :).

Issue #159, opened 2015-01-12
JobGraph: Support multiple OR'ed requirements (CompoundComponentRequirement)


kaspersorensen:
Make it possible to allow a single component to have multiple OR'ed component requirements.

This can already be done programmatically with the CompoundComponentRequirement.

 * Make sure that the graph rendering supports CompoundComponentRequirement. Each requirement should be shown as pointing to the target component.
 * Make it possible to add new OR'ed requirement via drawing a link in the graph.

tomaszguzialek:
Continuing the discussion from PR #381, how to visualize OR component requirements. I think, just adding a label "OR" or "AND" next to the component with a compound requirement is enough. Introducing a special component for flow control (something like "join" node before the actual component) would mean we should introduce a "fork" node to keep it consistent. At the moment we just have multiple edges coming out of a component, so let's just have multiple edges coming in, but with a label to indicate if it is an "OR" or "AND" condition.

kaspersorensen:
Where would you put the "OR" label? I mean, each edge is already individually labeled. Could also be that we use the icon to show a non-exclusive filter condition or whatever, I don't know.

tomaszguzialek:
I would put the label on the target component (from the rendering point of view - next to the component) when there is more than one incoming edge.

kaspersorensen:
![image](https://cloud.githubusercontent.com/assets/291450/7190706/48589350-e488-11e4-9834-b5dcd4c0d353.png)

So in this case ... same location as "String analyzer"?

tomaszguzialek:
Yes, next to the infamous "A?" that confused me before :).

kaspersorensen:
Hmmm .... But that's not the same place as the "String analyzer" label... And don't you think it is highly dependent on what angle the edges are coming in? My feeling is that one or more of the following are better approaches:

 * Use a different color for the (currently orange) icons on the edges
 * Add a new vertex representing the composite requirement (not happy about that one, but it's the most "pure" solution I can think of if we really aim for something sense-making).
 * Do nothing / keep like it is - this is not so sense-making but on the other hand it might be a matter of getting used to it.
 * Decorating the icon of the target vertex somehow but I am thinking we also do this when a component is invalidly configured and too many icon-decorations is also a bad thing.

PR #381, opened 2015-04-16
Made visualization of compound requirements work

kaspersorensen:
First part of #159 - the job graph visualization part. Still the story needs ability to make/modify compound requirements somehow to be complete.

Following is a screenshot of a compound requirement (that was made "in hand" in the .analysis.xml file):

image

The compound requirement is OR based ... is that communicated in the graph? can we improve that any way?

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/widgets/ChangeRequirementButton.java
+++ b/desktop/api/src/main/java/org/datacleaner/widgets/ChangeRequirementButton.java
@@ -31,11 +31,7 @@ import javax.swing.SwingUtilities;
 
 import org.datacleaner.job.AnyComponentRequirement;
 import org.datacleaner.job.ComponentRequirement;
-import org.datacleaner.job.FilterJob;
-import org.datacleaner.job.FilterOutcome;
 import org.datacleaner.job.builder.ComponentBuilder;
-import org.datacleaner.job.builder.FilterComponentBuilder;
-import org.datacleaner.job.builder.LazyFilterOutcome;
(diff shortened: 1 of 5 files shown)


tomaszguzialek:
A? I am confused, just by looking at the screenshot, TBH...

kaspersorensen:
Hmm yes I also think it's a bit confusing ... It makes more sense in other cases like:

```
A -> B -> C -> D -> ...
        \     /
           X
```

kaspersorensen:
How would you ideally visualize the above scenario though? Two filters on the source table and if either of them evaluates to a particular outcome, then run the component in question ....

tomaszguzialek:
I am getting confused by this "A?" - not really sure what this is supposed to mean. If there were "OR" instead of "A?" then it makes sense for me.

kaspersorensen:
Ah LOL ... that 'A' is part of the icon for 'String analyzer'. It has nothing to do with this particular issue :)

tomaszguzialek:
Oh, well... :). Anyways, we should add "OR" in the graph to notify the user that it is not "AND".

ClaudiaPHI:
I agree with Tomasz. Maybe it's better to add a visual thing on the graph.

kaspersorensen:
How would you do it? Should the composite requirement show up as a node/vertex on the graph?

kaspersorensen:
Gonna merge this since it is harmless (only an improvement over previous inability to visualize at all). Please continue the discussion about the visualization on issue #159

PR #382, opened 2015-04-17
Feature/332 library search

### LINKED ISSUE #332: Filter textbox for searching the library
Consider this way of finding what you need in a big tree of options (from eclipse):

image

We can see that it is not always trivial to find what you need in the DataCleaner "library" (component catalog) in the SchemaTree. It would be an amazing help to add a textbox like the above for searching it.

tomaszguzialek:
~~Do not merge yet. There must be some race condition. Once in 10 times maybe, when starting a new job - the left panel does not show up at all. If anyone can see where the issue is, I would appreciate the hint :).~~

Second questions: shall the search be expanded to the schema navigator as well?

Fixes #332


kaspersorensen:
I think we should not add it to the schema navigator. It would be cool but some DBs are really awfully slow in loading metadata and right now we have a lazy-loading node-expand thingie in place that would need to become eager for it to work :(

desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
```diff
@@ -269,10 +273,22 @@ private DefaultMutableTreeNode createLibrary(final DefaultMutableTreeNode librar
...
             final DefaultMutableTreeNode schemaNode = new DefaultMutableTreeNode(superCategory);
-            libraryRoot.add(schemaNode);
             final Collection<? extends ComponentDescriptor<?>> componentDescriptors = descriptorProvider
                     .getComponentDescriptorsOfSuperCategory(superCategory);
 
+            final List<ComponentDescriptor<?>> filteredComponentDescriptors = new ArrayList<>();
+
+            for (ComponentDescriptor<?> componentDescriptor : componentDescriptors) {
+                final String displayName = componentDescriptor.getDisplayName();
+                if (displayName.toLowerCase().contains(_searchTerm.toLowerCase())) {
```
kaspersorensen: Could search term ever be null? Maybe it happens for some reason and then your strange issue is that this causes a NPE? Just a thought, not sure at all.

tomaszguzialek: I doubt as it is initialized in the field itself to an empty string. A better candidate would be displayName, but I added print outs to both these variables and they are not null when the panel is not visible. Also, not a single exception is printed to the console when the panel is gone, so no trace at all...

tomaszguzialek:
@LosD was right - I checked if that race is introduced by my code and it turned out we have it on the master and also in the released version of DC. The PR is ready then.

desktop/ui/src/main/java/org/datacleaner/panels/SchemaTreePanel.java
```diff
@@ -39,73 +48,147 @@
...
+                        public void focusGained(FocusEvent e) {
+                            JTextField source = (JTextField) e.getSource();
+                            if (source.getText().equals(DEFAULT_SEARCH_FIELD_TEXT)) {
+                                source.setText("");
+                            }
+                        }
+                    });
+                    DCPanel searchPanel = new DCPanel(Color.WHITE);
+                    searchPanel.setLayout(new BorderLayout());
+                    JLabel resetSearchFieldIcon = new JLabel("X");
```
kaspersorensen: Not that it matters much, but I don't consider it to just a label - it's clearly a button IMO. So using a label is a bit strange and also means you have to use a MouseListener instead of a ActionListener (would be less cumbersome I think).

desktop/ui/src/main/java/org/datacleaner/panels/SchemaTreePanel.java
```diff
@@ -39,73 +48,147 @@
...
-					removeAll();
-					add(schemaTreeScroll, BorderLayout.CENTER);
-					updateParentPanel();
+    private static final long serialVersionUID = 1L;
+
+    private static final String DEFAULT_SEARCH_FIELD_TEXT = "Search component library...";
+
+    private final InjectorBuilder _injectorBuilder;
+    private JComponent _updatePanel;
+    private final JTextField _searchTextField;
```
kaspersorensen: One thing I notice that could go wrong here .... The searchTextField is a single instance for this panel. But the schema tree seems to be loaded for every time setDatastore(...) is called. That means the same search textbox might be used on multiple trees. And will thus be instrumented many times by multiple swing workers. That doesn't seem to work very well in this code. I think the instantiation of the textbox should follow the same lifecycle as the instantiation/injection of the schema tree.

PR #383, opened 2015-04-17
Adding listeners for analyzers of type AnalyzerResultFuture

### LINKED ISSUE #353: Do not call AnalysisListener.jobSuccess() before any AnalyzerResultFutures have finished
Right now we notify listeners of the jobSuccess() event when the last analyzer has delivered a result object. But since 4.0 we have the AnalyzerResultFuture object which indicate a "not yet finished result" object. These are listenable and this mechanism should be respected by the AnalysisListener to not prematurely indicate a fully finished job.

ClaudiaPHI:
Fixes #353 added listeners for analyzers of type AnalyzerResultFuture.

I hope this is what you meant for #353 .

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/panels/result/ResultListPanel.java
+++ b/desktop/ui/src/main/java/org/datacleaner/panels/result/ResultListPanel.java
@@ -133,6 +133,7 @@ public class ResultListPanel extends DCPanel {
                 return component;
             }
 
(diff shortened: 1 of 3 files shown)


kaspersorensen:
It's not what I meant :-) This should not be a fix in the UI. It should be fixed in engine-core somewhere. Wherever the listener is being invoked in the first place.

PR #386, opened 2015-04-21
Issue #370: Improved exception hierarchy+handling in validation cases.

### LINKED ISSUE #370: Inconsistency in @validate and automagic checks in Execute button
The execute button in DataCleaner disables when a job fails the automagic checks (eg no columns selected), but when a transformer or anylyzer does any custom checking using the @validate annotation and rejects the job, the execute button remains active. Hitting the execute button results in an exception being thrown.

kaspersorensen:
Fixes #370

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/descriptors/AbstractMethodDescriptor.java
+++ b/engine/core/src/main/java/org/datacleaner/descriptors/AbstractMethodDescriptor.java
@@ -28,8 +28,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import org.datacleaner.util.ReflectionUtils;
 import org.apache.metamodel.util.BaseObject;
+import org.datacleaner.util.ReflectionUtils;
 
 /**
  * Abstract descriptor for things that are represented by a method on a
@@ -57,6 +57,10 @@ class AbstractMethodDescriptor extends BaseObject implements Serializable {
         _name = method.getName();
         _componentDescriptor = componentDescriptor;
     }
+    
+    public ComponentDescriptor<?> getComponentDescriptor() {
+        return _componentDescriptor;
+    }
 
     public final Method getMethod() {
         if (_method == null) {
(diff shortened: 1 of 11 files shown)


engine/core/src/main/java/org/datacleaner/descriptors/AbstractMethodDescriptor.java
```diff
@@ -70,22 +74,29 @@ public final String toString() {
         return getClass().getSimpleName() + "[method=" + _name + "]";
     }
 
-    protected final void invoke(Object component) {
+    protected final void invoke(Object component) throws RuntimeException, IllegalStateException {
```
LosD: Aren't these redundant?

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -774,9 +785,11 @@ public AnalysisJobMetadata createMetadata() {
      * 
      * @return
      * @throws IllegalStateException
-     *             if the job is invalidly configured.
+     *             if the job is invalidly configured. See
+     *             {@link #isConfigured(boolean)} for detailed exception
+     *             descriptions.
      */
-    public AnalysisJob toAnalysisJob() throws IllegalStateException {
+    public AnalysisJob toAnalysisJob() throws RuntimeException {
```
LosD: Same here. The @throws is enough for documentation purposes.

LosD:
It seems like a bit more than fixing #370, as it also seems to fix the issue with transformers that produces results not allowing executing the job, but I guess they're somehow related anyway.

I'm not too fond of using the throws keyword for RuntimeExceptions, but other than that it looks good.

engine/core/src/main/java/org/datacleaner/descriptors/AbstractMethodDescriptor.java
kaspersorensen: Well yes and no :) The RuntimeException is basically thrown if anything goes wrong in the validation method itself. The IllegalStateException is explicitly mentioned also because the code here (the "wrapper" around the validation method) might also throw such an exception.

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
kaspersorensen: The doc needs an update yes. But do you think there's anything wrong with both having it in javadoc and as a method modifier? IMO that is best practice?

LosD: Hmmmm, as far as I remember best practice is more or less only to have RuntimeExceptions in ```@throws```. At least it seems to be the way their own code is used, and I think best practice would be not to having people guessing if it is a checked exception or not (of course that matters a little less now, with the IDE checking it live).

kaspersorensen: Hmm I always learned that putting the exceptions there was nice as a way of documenting it to the next guy who reads the code. Of course there are never any guarantees that an unchecked exception cannot be thrown, but if you explicitly put it on the throws-list, then you tell the consumer of the code that it is not uncommon or that there are specific scenarios where it might fail.

PR #388, opened 2015-04-22
Feature/353 future listener invocation

### LINKED ISSUE #353: Do not call AnalysisListener.jobSuccess() before any AnalyzerResultFutures have finished

kaspersorensen:
Fixes #353 

One remark/caution on this - the approach chosen (to block in CollectResultsTask) will block on of the active thread in the TaskRunner's threadpool. So if we have really a lot of AnalyzerResultFutures in the same job, this might get out of hand. But even then it should actually be OK because the futures will typically be running their work in an independent threadpool (typically MetaModel's LazyRef based).


LosD:
Looks good, merge as soon as Travis agrees

PR #389, opened 2015-04-22
Fix for reflection issue with renderer selection

kaspersorensen:
Updated ReflectionUtils.getHierarchyDistance() to support mix of class and interface hierarchy distance.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/ReflectionUtils.java
+++ b/engine/core/src/main/java/org/datacleaner/util/ReflectionUtils.java
@@ -331,54 +331,78 @@ public final class ReflectionUtils {
         throw new UnsupportedOperationException("Parameter type not supported: " + someType);
     }
 
-    public static int getHierarchyDistance(Class<?> subtype, Class<?> supertype) {
+    public static int getHierarchyDistance(Class<?> subtype, Class<?> supertype) throws IllegalArgumentException {
         assert subtype != null;
         assert supertype != null;
-
-        if (subtype == supertype) {
-            return 0;
-        }
 
         if (!ReflectionUtils.is(subtype, supertype)) {
             throw new IllegalArgumentException("Not a valid subtype of " + supertype.getName() + ": "
                     + subtype.getName());
         }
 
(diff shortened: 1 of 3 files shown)


engine/core/src/main/java/org/datacleaner/util/ReflectionUtils.java
```diff
@@ -331,54 +331,80 @@ public static boolean isWildcard(Type type) {
...
         }
     }
 
-    private static int getInterfaceHierarchyDistance(Class<?> subtype, Class<?> supertype) {
+    private static int getClassHierarchyDistance(final Class<?> subtype, final Class<?> supertype) {
         if (subtype == supertype) {
             return 0;
         }
+        if (subtype == Object.class) {
+            throw new IllegalArgumentException("Finished traversing hierarchy and found Object.class");
```
LosD: I might be wrong, but this seems like a bad use of exceptions to me; judging by the code below (wrapping them in empty catch blocks), there is nothing exceptional about this, it's just a "not found".

LosD:
Except what looks like an exception-based integer version of the [tri-state boolean](http://thedailywtf.com/articles/What_Is_Truth_0x3f_), it looks good to me :)

engine/core/src/main/java/org/datacleaner/util/ReflectionUtils.java
kaspersorensen: Yeah it is a case of "we tried to traverse through the hierarchy but didn't fint the right way this time" situation. How would you rather return? We could return -1 instead or something, but isn't that excactly the tri-state boolean? :)

LosD: Hehe, yeah maybe, but it's still better than using an unexceptional exception :)

I'd maybe consider Integer.MAX, as that is the closest thing to "infinite distance", but -1 probably also works fine (and would be 0xFFFFFFFFF, so looking at it as unsigned, it is even further away :)).

kaspersorensen:
Fixed it and merging :)

PR #391, opened 2015-04-23
Feature/225 non analyzer results

LosD:
Adds the categorization result to the country standardizer, and makes the renderer easy to extend by putting the reusable bits into a new base class.


components/standardizers/src/main/java/org/datacleaner/beans/standardize/CountryStandardizationTransformer.java
```diff
@@ -61,19 +76,49 @@ public OutputColumns getOutputColumns() {
...
+            correctedCountryName = LabelUtils.UNEXPECTED_LABEL;
         }
+        
+        // ConcurrentHashMap does not support null keys
+        synchronized (this) {
+            if (!countryCountMap.containsKey(countryName)) {
+                countryCountMap.put(correctedCountryName, _rowAnnotationFactory.createAnnotation());
+            }
+
+            _rowAnnotationFactory.annotate(inputRow, 1, countryCountMap.get(correctedCountryName));
```
kaspersorensen: The annotate method is thread-safe but it is placed in the synchronized block. A slight improvement would be to separate the resolving of the RowAnnotation (in synchronized block) from the annotate call (outside).

desktop/ui/src/main/java/org/datacleaner/widgets/result/CategorizationResultAbstractSwingRenderer.java
```diff
@@ -0,0 +1,105 @@
...
+import org.datacleaner.result.CategorizationResult;
+import org.datacleaner.result.renderer.AbstractRenderer;
+import org.datacleaner.result.renderer.RendererFactory;
+import org.datacleaner.util.IconUtils;
+import org.datacleaner.util.LabelUtils;
+import org.datacleaner.util.WidgetFactory;
+import org.datacleaner.windows.DetailsResultWindow;
+import org.jfree.data.general.DefaultPieDataset;
+
+public abstract class CategorizationResultAbstractSwingRenderer<R extends CategorizationResult> extends AbstractRenderer<R, JComponent> {
```
kaspersorensen: Naming-wise I much prefer to put Abstract as first, so it becomes AbstractCategorizationResultSwingRenderer

kaspersorensen:
Not much I could find, but I did get two good blows at your code ;-)

PR #394, opened 2015-04-23
Fixes #393 by only asserting on certain indexes in the events instead of absolute ordering.

### LINKED ISSUE #393: Sporadic unittest failure - AnalyzerResultFutureAndAnalysisListenerTest
From time to time I see travis failing with this issue:

```
  AnalyzerResultFutureAndAnalysisListenerTest.testScenario:181 expected:
<...er (2).onSuccess(), [jobSuccess, ]AnalysisResultFuture...>
 but was:
<...er (2).onSuccess(), []AnalysisResultFuture...>
```

Might be a race condition in the test or might be something related to the jobSuccess event that is bad. Take a look ...

kaspersorensen:
Fixes #393 .... The level of readability in the test is decreased I am affraid. But it will no longer fail because of race conditions among certain events in the listeners.

### DIFF
--- a/engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
+++ b/engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
@@ -210,7 +210,12 @@ public class AnalyzerResultFutureAndAnalysisListenerTest extends TestCase {
 
     private int getIndexAndVerifyExists(List<String> messagesList, String string) {
         final int index = messagesList.indexOf(string);
-        assertTrue(messagesList.toString(), index != -1);
+        if (index == -1) {
+            // use a good message because hunting down the cause of this can be
+            // pretty tough.
+            final String message = "Failed to find '" + string + "' in messages: " + messagesList.toString();
+            assertTrue(message, index != -1);
+        }
         return index;
     }
 }
--- a/engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
+++ b/engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
@@ -165,38 +165,52 @@ public class AnalyzerResultFutureAndAnalysisListenerTest extends TestCase {
         final List<String> messagesList = new ArrayList<>();
         messages.drainTo(messagesList);
 
-        final String messageFromNormalAnalyzerSuccess = "componentSuccess(normal analyzer,ListResult)";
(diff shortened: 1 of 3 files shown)


kaspersorensen:
Oh gosh ... Ran it 60 times (in a loop) locally without any issues ... back to work :)

engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
```diff
@@ -165,38 +165,57 @@ public void jobSuccess(AnalysisJob job, AnalysisJobMetrics metrics) {
...
+        assertTrue(originalMessagesString, indexAnalysisResultFutureAwaitReturned > indexListener2Success);
+
+        final int indexAnalyzerResultReady = getIndexAndVerifyExists(messagesList,
+                "AnalyzerResultFuture.isReady() (2) = true");
+
+        assertTrue(originalMessagesString, indexAnalyzerResultReady > indexAnalyzerResultNotReady);
+
+        final int indexJobSuccess = getIndexAndVerifyExists(messagesList, "jobSuccess");
+
+        assertTrue(originalMessagesString, indexJobSuccess > indexFutureAnalyzerSuccess);
```
LosD: I'm not sure I totally understand this use of the assert message field, or maybe I don't understand whats in originalMessageString?

LosD:
The assert pattern looks a bit weird, but that is probably just me.

engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
kaspersorensen: the first part is just a message that junit will output if the assertion fails. All the assertXxx methods have that as an optional first parameter. In this case it outputs all the messages that was received. That way if the test ever fails we can at least see the list of messages that we got and analyze the test issue from there.

kaspersorensen:
other than that ready to merge, or?

engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
LosD: I know the JUnit message parameter, it's just a using it for something else than an actual message like "The list should have had stuff in it" :)

LosD:
Yeah!

engine/core/src/test/java/org/datacleaner/test/full/scenarios/AnalyzerResultFutureAndAnalysisListenerTest.java
kaspersorensen: True ... But I didn't want to do all that string concatenation, haha :-D

PR #410, opened 2015-04-30
Ensures that jobgraph rendering never has redundant lines

kaspersorensen:
Before:

image

After:

image


LosD:
Not that bad, but it would be even less bad if the while loop was isolated inside removeUnnecesaryEdges.

If you don't agree, just merge.

kaspersorensen:
That makes sense. Will do.

ClaudiaPHI:
[LGTM](http://lgtm.in/i/0OXGLOITp)

[:+1:](http://lgtm.in/u/0OXGLOITp) [:-1:](http://lgtm.in/r/0OXGLOITp)

PR #414, opened 2015-05-05
GenerateIdTransformer with SEQUENCE and ROWID modes.

### LINKED ISSUE #413: Make GenerateIdTransformer use row ID.
It would be nice to have an option to generate an ID that will preserve the order of the records in a file. Adding a mode to GenerateIdTransformer that will return the row ID should do the trick.

tomaszguzialek:
Fixes #413

### DIFF
--- a/components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
+++ b/components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
@@ -38,35 +38,48 @@ import org.datacleaner.components.categories.NumbersCategory;
 @Categorized(NumbersCategory.class)
 public class GenerateIdTransformer implements Transformer {
 
-	@Configured
-	@Description("A column which represent the scope for which the ID will be generated. "
-			+ "If eg. a source column is selected, an ID will be generated for each source record. "
-			+ "If a transformed column is selected, an ID will be generated for each record generated that has this column.")
-	InputColumn<?> columnInScope;
+    public enum IdType {
+        SEQUENCE, ROWID;
+    }
 
-	@Configured
-	int offset = 0;
+    @Configured
+    @Description("A type of ID which will be generated for each record in scope. Current options: sequential numbers or row ID.")
+    IdType idType = IdType.SEQUENCE;
 
-	private final AtomicInteger _counter;
+    @Configured
+    @Description("A column which represent the scope for which the ID will be generated. "
+            + "If eg. a source column is selected, an ID will be generated for each source record. "
+            + "If a transformed column is selected, an ID will be generated for each record generated that has this column.")
+    InputColumn<?> columnInScope;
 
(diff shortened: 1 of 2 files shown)


components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
```diff
@@ -38,35 +38,48 @@
...
+
+    @Override
+    public OutputColumns getOutputColumns() {
+        return new OutputColumns(Integer.class, "Generated ID");
+    }
+
+    @Override
+    public Integer[] transform(InputRow inputRow) {
+        int id;
+        if (IdType.ROWID.equals(idType)) {
```
kaspersorensen: when doing enum comparison the double-equal operator (==) is a bit nicer to use :)

components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
```diff
@@ -38,35 +38,48 @@
 @Categorized(NumbersCategory.class)
 public class GenerateIdTransformer implements Transformer {
 
-	@Configured
-	@Description("A column which represent the scope for which the ID will be generated. "
-			+ "If eg. a source column is selected, an ID will be generated for each source record. "
-			+ "If a transformed column is selected, an ID will be generated for each record generated that has this column.")
-	InputColumn<?> columnInScope;
+    public enum IdType {
+        SEQUENCE, ROWID;
```
kaspersorensen: ROW_NUMBER would be a bit better than ROWID in my opinion

components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
```diff
@@ -38,35 +38,48 @@
...
+    }
+
+    @Override
+    public OutputColumns getOutputColumns() {
+        return new OutputColumns(Integer.class, "Generated ID");
+    }
+
+    @Override
+    public Integer[] transform(InputRow inputRow) {
+        int id;
```
kaspersorensen: final? :smiling_imp:

components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
kaspersorensen: And if you make your enum also implement HasName then they can have pretty names in the UI and documentation :-)

kaspersorensen:
Not sure if I was being too evil? :hurtrealbad:

tomaszguzialek:
I am happy to see very detailed comments, so you can keep "evil-boss-final-keyword-checkstyle" mode on :). Comments addressed, I am merging it when Travis agrees.

PR #416, opened 2015-05-05
Issue #406: Fixed (and also a few icon references)

### LINKED ISSUE #406: Styling for disabled Swing components.
I had a use case where I needed to make certain fields (text input fields) disabled for user input. The method setEnabled(false) indeed makes them disabled, but the styling of the component does not change at all. We should introduce a unified way to disable a component and communicate it clearly to the user (for example by making them gray).

kaspersorensen:
Fixes #406 
(The trick was to add a UIManager property and to make sure our colors are ColorUIResources which are apparently needed in the TextFieldUI classes).

Here's an example dialog with a disabled text field:

image


tomaszguzialek:
Looks disabled now (I mean, good :)). Merging.

PR #417, opened 2015-05-06
Feature/224 output dataset producers

### LINKED ISSUE #224: Output DataSet producers and jobs
Today one of the limitations to the compositional power of DataCleaner is that a job's data stream always ends when an analyzer is used. This is because the analyzer is performing an aggregate function on the dataset and as such it cannot produce records in the same cardinality as the input. We do have a couple of use cases where an aggregate function + further post-processing is necesary though:

 * Sorting data
 * Using a batch-based remote/web service for data transformation
  * Will typically require sending off the whole dataset and then wait for a result. Once the result comes back the stream in DC is out of scope.
 * Deduplication
  * Typically a duplicate detection component would detect duplicate records / pairs / groups (3 different output datasets).
  * A succeeding merging component would consume those records and produce merged records.

We want to make this possible so that we can model larger logical flows in DataCleaner.

From a XML spec standpoint, here's a draft example (using the deduplication scenario because that's the most complex one):

```
<analysis-job>
<source>
  <column name="person_id" id="id" />
  <column name="Full name" id="fullname" />
  <column name="Age" id="age" />
</source>
<analyzer descriptor="Duplicate detection">
  <input ref="fullname" />
  <input ref="age" />
  <output-dataset id="PAIRS" name="Pairs">
    <analysis-job>
      <source>
         <column name="RecordID1" id="id1" />
         <column name="RecordID2" id="id2" />
         <column name="Score" id="score" />
      </source>
      ...
    </analysis-job>
  </output-dataset>
  <output-dataset id="RECORDS" name="Records">
    ...
  </output-dataset>
</analyzer>
</analysis-job>
```

Things to notice:

 * The analyzer can have multiple output-datasets
 * In the output-dataset we can introduce a completely new analysis job.
 * The column definitions and references in the output-dataset's job are completely separate from the outer analysis job. In fact, they should also not have any way to access the outer data stream because the scope and granularity will not have any direct connection.

kaspersorensen:
Attempt to fix #224

__Do not merge - work in progress__

This is so far posted as a PR to follow the build in Travis and to share intermediate design ideas etc.


LosD:
Hmmmm... Another giga branch. We really need to start doing something about these, like keeping them completely unpushed until the the commits has been rewritten into logical chunks like API change, backend change, UI change and so on.

Branches like theses are bordering on unreviewable.

kaspersorensen:
Agreed. I don't want to merge it either. Just a spike.

kaspersorensen:
But good point about breaking it up later when it is ready to be reviewed!

Honestly, I only wanted the remote branch to keep the work backed up and to have a PR link to share and to see travis builds.

LosD:
The only problem is that none of us are very experienced in history rewriting, but I guess that's a chance for learning! :)

kaspersorensen:
Actually it doesn't seem that we need rewriting. What I am doing is this:

 * First of course figuring out what are logicial pieces for PRs. Right now I imagine these: 1) the API itself (done in #419) 2) The engine part which facilitates execution of these new types of jobs 3) the XML job layer and 4) the desktop UI layer
 * When I want to make a PR (now for item no. 2) then I get the diff of _this_ PR and then I apply it to a new branch.
 * I then remove the stuff from it that I don't want to include in my commit.
 * I commit and push that branch, and create a PR out of it.
 * Once the PR is (hopefully) merged, then I merge branch back into _this_ branch.

LosD:
Yeah okay, that is also an option. History rewriting is more about taking commits and reordering them, or squashing them into a single commit, or even splitting them. In your way, you don't have to deal with all that, which is certainly an advantage.

LosD:
(As someone who is fascinated but mortally afraid of history rewriting, I'd even say a HUGE advantage :))

kaspersorensen:
I'm closing this PR - superceeded by:

 * #417 (already merged)
 * #422 (please review - xsd update - pretty trivial)
 * #421 (please review - engine update - more complex)

PR #419, opened 2015-05-08
Feature/224 output data stream api

kaspersorensen:
This is the API changing part of the greater story #224. To make review easier and incremental this bit is pushed. It's actual origin is in PR #417 (which I do not want to have merged yet).

The API change is harmless to push now in my opinion, as everything here is optional and the engine is not yet aware of the new interfaces. Adding them just allows component builders to start taking advantage of the new API possibilities before an eventual DC 4.1 release.


LosD:
LGTM!

kaspersorensen:
Goodie, then we merge.

drexler42:
You just beat me to it by 41 seconds!

drexler42:
Anyways, I could not see any code change that would break existing code, so that is good.

PR #420, opened 2015-05-08
Feature/360 sort integers numerically

### LINKED ISSUE #360: Sorting in Create CSV and Create Excel works incorrectly for numeric content
In  Create CSV file and Create Excel File it is possible to select a 'Column to be sorted on'  under 'Optional properties'. 
It sorts, but not correctly for numeric input, such as ID's. Numbers are being sorted alphabetically?
ID's had been created by use of DC transformer <Generate ID>.
I even applied transformer Convert to number, but still the same incorrect ordering in output.

Ordering of output is important for tracability / readibility of results,since DC transformers randomize order due to multi-threading.

tomaszguzialek:
Fixes #360


kaspersorensen:
The code looks good, but with some code duplication in the two writers. Could we move the createComparator() method to a shared utility class for these two?

tomaszguzialek:
I introduced a SortHelper class in org.datacleaner.extension.output, so next to Create CSV and Excel analyzers - not sure if there is a better place for it.

kaspersorensen:
:+1:

PR #421, opened 2015-05-08
Issue #224: Added engine and job-builder support for OutputDataStreams

kaspersorensen:
This is the Job definition, job building and engine supporting part of the greater story #224. To make review easier and incremental this bit is pushed. It's actual origin is in PR #417 (which I do not want to have merged yet) from which also #419 was separated out as a "child PR".

This PR is still a pretty big one as it represents the main body of work needed to support output data streams. In addition there was a few refactorings made to the initialization and closing of components which was necesary in order to ensure the timely initialization and closing of the output data streams that are dependent on their publishing component's life cycle.

To give my input for the reviewer (of course he/she should make up his own mind):

 * The most important bit IMO is the new test - JobWithOutputDataStreamsTest. I see this as a kind of acceptance test of both the builder API changes and of the fact that the engine itself is working as it should wrt. OutputDataStreams. So I'm very interested in your feedback on the readability and logic of this test.
 * I've gone ahead with an approach to the engine-level work where each output data stream gets its own AnalysisJobBuilder at building-time and its own RowProcessingPublisher at execution time. With regard to the RowProcessingPublisher the main concern was that normally these are query based whereas the new use-case of the object is now push based. That's why there is a new CountDownLatch (tadaaaa) which it will wait for instead of firing a query. Not totally nice but seemed to be a pretty low-impact way of adding this feature. The latch is closed by the publishing consumer's close task which I think is a pretty well fit - because the close task is guaranteed to run always and because this means that the output data stream will stay alive as long as the publishing component is alive (also pretty logical - to me at least).
 * I'm sorry I "accidentally" also did something additional in this PR which I should have done separately. My apologies to the reviewer for these distractions. One thing took the other. So to prepare you, these things are here in addition which I consider trivial and should probably just be ignored in the review:
  * I fixed a typo in the AnalysisListener interface ("errorUknown") - this changes of course all the implementing classes too - Moved to PR #423
  * I updated a few test cases where AnalyzerBeansConfiguration was still used instead of DataCleanerConfiguration - Moved to PR #423
  * While dealing with the RowProcessingPublisher I eventually also ended up changing a bit of the logic about the Tasks that run to initialize components. There used to be some classes implementing "LifeCycleCallback" which was becoming pretty legacy to work with. So once I had to work with it I decided to refactor it too. This means that these classes have mostly been deleted and instead their logic moved to LifeCycleHelper.

Given the many aspects above, maybe it would even be good to split up this PR too (like #417).


kaspersorensen:
Updated the description of this PR with a bunch of my own notes ... Please read before reviewing :)

engine/core/src/test/java/org/datacleaner/test/full/scenarios/JobWithOutputDataStreamsTest.java
```diff
@@ -0,0 +1,155 @@
...
+            // the output stream
+            assertFalse(analyzer1.isOutputDataStreamConsumed(dataStream));
+
+            final AnalyzerComponentBuilder<MockAnalyzer> analyzer2 = outputDataStreamJobBuilder
+                    .addAnalyzer(MockAnalyzer.class);
+            analyzer2.addInputColumns(outputDataStreamColumns);
+            analyzer2.setName("analyzer2");
+            assertTrue(analyzer2.isConfigured());
+
+            // the stream is still not "consumed" yet
```
tomaszguzialek: The comments is in contradiction to the assert.

tomaszguzialek:
I was not capable of reviewing the whole stuff as it is really big and also deep down in DC internals. Following your reviewer tips, I looked into the acceptance test and it started to make sense what is going on - I like the API then.

Not able to fully review as I said, but except from the contradiction in the comment and the fact it is not mergeable anymore, looks good to me.

kaspersorensen:
Fixed the comment and merged in the new master. Let's see travis to make sure.

kaspersorensen:
I'll fix merge conflicts

kaspersorensen:
Merge conflicts resolved

LosD:
As an addition to improving comments, the OutputRowCollector JavaDoc comments no longer really match the extended usage made possible by OutputDataStreamRowCollector, it only mentions Transformers and usage in relation to them.

Other than that it seems to look good, but as @tomaszguzialek mentioned it is hard to get a feeling of how well things interacts together because of the sheer size.

LosD:
Merging and hoping for the best... Needed to get on with stuff :)

PR #422, opened 2015-05-08
Issue #224: Updated the job.xsd file

kaspersorensen:
This is the Job XML Schema part of the greater story #224. To make review easier and incremental this bit is pushed. It's actual origin is in PR #417 (which I do not want to have merged yet) from which also #419 was separated out as a "child PR".

This particular part is only about the change pertaining to the XSD (and the generated java classes from this). It is NOT about adapting the job reader+writer to actually support reading+writing output data stream jobs. That part is pending still.

### DIFF
--- a/engine/xml-config/src/main/resources/job.xsd
+++ b/engine/xml-config/src/main/resources/job.xsd
@@ -5,18 +5,25 @@
 
 	<element name="job">
 		<complexType>
-			<sequence>
-				<element name="job-metadata" type="ab:jobMetadataType"
-					minOccurs="0" maxOccurs="1" />
-				<element name="source" type="ab:sourceType" minOccurs="0"
-					maxOccurs="1" />
-				<element name="transformation" type="ab:transformationType"
-					minOccurs="0" maxOccurs="1" />
-				<element name="analysis" type="ab:analysisType" minOccurs="1"
-					maxOccurs="1" />
-			</sequence>
+			<complexContent>
+				<extension base="ab:jobType">
+				</extension>
+			</complexContent>
 		</complexType>
 	</element>
+
+	<complexType name="jobType">
+		<sequence>
+			<element name="job-metadata" type="ab:jobMetadataType"
+				minOccurs="0" maxOccurs="1" />
+			<element name="source" type="ab:sourceType" minOccurs="0"
+				maxOccurs="1" />
+			<element name="transformation" type="ab:transformationType"
+				minOccurs="0" maxOccurs="1" />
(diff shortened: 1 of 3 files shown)


engine/xml-config/src/main/resources/job.xsd
```diff
@@ -5,19 +5,26 @@
...
-					maxOccurs="1" />
-				<element name="transformation" type="ab:transformationType"
-					minOccurs="0" maxOccurs="1" />
-				<element name="analysis" type="ab:analysisType" minOccurs="1"
-					maxOccurs="1" />
-			</sequence>
+			<complexContent>
+				<extension base="ab:jobType">
+				</extension>
+			</complexContent>
```
LosD: This seems a bit overdone: If you want to make an element of jobType, ```<element name="job" type="ab:jobType"/>``` should be enough.

engine/xml-config/src/main/resources/job.xsd
```diff
@@ -156,37 +165,48 @@
...
+			<extension base="ab:componentType">
+				<sequence>
+					<element name="outcome" type="ab:outcomeType" minOccurs="1"
+						maxOccurs="unbounded" />
+				</sequence>
+			</extension>
+		</complexContent>
+	</complexType>
+
+	<complexType name="analyzerType">
```
LosD: Same here

LosD:
Without really knowing the actual needed XSD, it looks good to me except for what looks like a few misuses of extensions.

engine/xml-config/src/main/resources/job.xsd
kaspersorensen: I tried that initially actually. But then the JAXB would use as the root element "jobType" instead of a dedicated root "Job". Problem with that is then that when you try to marshall or unmarshall with JAXB it throws an exception because the JobType is not a sure root type. Bah.

engine/xml-config/src/main/resources/job.xsd
kaspersorensen: Hmm yeah but I sort of liked to have a dedicated type here in case anything particular would show up... And anyways there's a lot of parsing code that is unfortunately NOT general so it needs to know that this type is an analyzer not just any component type.

engine/xml-config/src/main/resources/job.xsd
LosD: Ugh!

engine/xml-config/src/main/resources/job.xsd
kaspersorensen: Ah, retracting my statement. Will try with type attribute.

PR #423, opened 2015-05-10
Added various improvements from #421 that are slightly unrelated.

kaspersorensen:
Made this PR with things from #421 that are not entirely related to the resolution of #224 (but quite convenient in order to fix it anyway).

### DIFF
--- a/engine/core/src/test/java/org/datacleaner/test/TestEnvironment.java
+++ b/engine/core/src/test/java/org/datacleaner/test/TestEnvironment.java
@@ -1,0 +1,41 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
+ * copy, or redistribute it subject to the terms and conditions of the GNU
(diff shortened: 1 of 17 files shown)


engine/core/src/test/java/org/datacleaner/test/TestEnvironment.java
```diff
@@ -0,0 +1,41 @@
...
+            .withTaskRunner(TASK_RUNNER);
+
+    public static MultiThreadedTaskRunner getMultiThreadedTaskRunner() {
+        return TASK_RUNNER;
+    }
+
+    public static DataCleanerEnvironment getEnvironment() {
+        return ENVIRONMENT;
+    }
+}
```
LosD: Why the static getters? Not really a problem, just curious.

LosD:
LGTM! Merging

engine/core/src/test/java/org/datacleaner/test/TestEnvironment.java
kaspersorensen: Only there to make it easy to do stuff later if we need to (as opposed to exposing the constants which would mean we couldn't intercept the call).

PR #424, opened 2015-05-10
Prepared the UI handlers to have multiple builder objects in graph

kaspersorensen:
This is part of the greater story #224. To make review easier and incremental this bit is pushed. It's actual origin is in PR #417 (which I do not want to have merged yet) from which also #419 was separated out as a "child PR".

This particular part is only about preparing the UI handlers to have multiple builder objects in graph which will be the case when #224 is done.


LosD:
LGTM! Merging

PR #426, opened 2015-05-11
Feature/425 custom header names csv excel

### LINKED ISSUE #425: CreateCSVAnalyzer and CreateExcelSpreadsheetAnalyzer should have an option to specify column headers.
Similar to "ElasticSearch indexer", there should be a possibility to specify column headers in "Create CSV Analyzer" and "Create Excel Spreadsheet Analyzer".

Screenshot from ElasticSearch indexer:
image

tomaszguzialek:
Fixes #425

### DIFF
--- a/components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
+++ b/components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
@@ -96,7 +96,7 @@ public class CreateCsvFileAnalyzer extends AbstractOutputWriterAnalyzer implemen
     @Configured(order = 6, required = false, value = PROPERTY_COLUMN_TO_BE_SORTED_ON)
     InputColumn<?> columnToBeSortedOn;
 
-    @Configured(value = PROPERTY_OVERWRITE_FILE_IF_EXISTS)
+    @Configured(value = PROPERTY_OVERWRITE_FILE_IF_EXISTS, required = false)
     boolean overwriteFileIfExists;
 
     @Configured(PROPERTY_FIELD_NAMES)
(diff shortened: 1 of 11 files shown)


components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -155,10 +161,10 @@ public OutputWriter createOutputWriter() {
 
         List<String> headers = new ArrayList<String>();
         for (int i = 0; i < columns.length; i++) {
-            String columnName = columns[i].getName();
+            String columnName = getColumnHeader(i);
             headers.add(columnName);
             if (columnToBeSortedOn != null) {
-                if (columnName.equals(columnToBeSortedOn.getName())) {
+                if (columns[i].getName().equals(columnToBeSortedOn.getName())) {
```
kaspersorensen: I would argue that we shouldn't compare names, but actual columns. I know this is not new in this PR, but I think it's bad anyway :-P

desktop/ui/src/main/java/org/datacleaner/panels/writers/CustomHeaderColumnNamesAnalyzerJobPanel.java
```diff
@@ -0,0 +1,67 @@
...
+    private final MultipleMappedStringsPropertyWidget _mappedWidget;
+    private final ConfiguredPropertyDescriptor _inputColumnsProperty;
+    private final ConfiguredPropertyDescriptor _mappedStringsProperty;
+
+    public CustomHeaderColumnNamesAnalyzerJobPanel(AnalyzerComponentBuilder<?> analyzerJobBuilder,
+            PropertyWidgetFactory propertyWidgetFactory) {
+        super(analyzerJobBuilder, propertyWidgetFactory);
+
+        _inputColumnsProperty = analyzerJobBuilder.getDescriptor().getConfiguredProperty(CreateCsvFileAnalyzer.PROPERTY_COLUMNS);
+        _mappedStringsProperty = analyzerJobBuilder.getDescriptor().getConfiguredProperty(CreateCsvFileAnalyzer.PROPERTY_FIELD_NAMES);
```
kaspersorensen: This class is used to make a panel for both CSV and Excel analyzers, but here the constants of CreateCsvFileAnalyzer is used. I would suggest you ensure that there's a common place where you can refer to them, or else the point of having them is not really met. It could maybe just be the super-class?

kaspersorensen:
Looks good to me, except for the remarks above.

tomaszguzialek:
Fixed all comments, waiting for Travis confirmation and merging.

Issue #395, opened 2015-04-26
Component documentation improvements


kaspersorensen:
Following the first stab we've made at a component documentation tool (#343), here's a bunch of improvements/applications of it that we would like to explore:

 * Right-click a component in the library tree and open up the documentation of it (either in embedded browser or just using OpenBrowserAction).
 * Add support for links to tutorials and other external documentation/sites.
 * Add support for links to other component's documentation pages.
 * Make an index page that groups the links in the same tree-structure as in DC desktop
 * Add support for adding screenshots (hosted on a public URL - we don't want screenshots in the JARs)

kaspersorensen:
As hint to how one can make a button to open up the documentation of a particular component, I did this code as a quick experiment (didn't push because I wasn't happy with icon and button placement and all that):

ComponentConfigurationDialog:
```
final JButton documentationButton = WidgetFactory.createDefaultButton(null, IconUtils.ACTION_HELP);
        documentationButton.addActionListener(new ActionListener() {
            @Override
            public void actionPerformed(ActionEvent e) {
                final DataCleanerConfiguration configuration = _componentBuilder.getAnalysisJobBuilder()
                        .getConfiguration();
                final DataCleanerHomeFolder homeFolder = configuration.getHomeFolder();

                final File homeDirectory = homeFolder.toFile();
                final File docDirectory = new File(homeDirectory, "documentation");
                if (!docDirectory.exists()) {
                    docDirectory.mkdir();
                }

                final File documentationFile = new File(docDirectory, IndexDocumentationBuilder
                        .getFilename(_componentBuilder.getDescriptor()));
                if (!documentationFile.exists()) {
                    final ComponentReferenceDocumentationBuilder builder = new ComponentReferenceDocumentationBuilder(
                            configuration.getEnvironment().getDescriptorProvider());
                    builder.writeDocumentationToDirectory(docDirectory);
                }

                new OpenBrowserAction(documentationFile.toURI()).actionPerformed(e);
            }
        });
```

PR #429, opened 2015-05-12
Issue #395: Added @ExternalDocumentation annotation 

kaspersorensen:
As part of #395 I added a new annotation for external documentation.
And I put video links to various components.

Questions for reviewer:

 * Do you like the API suggested here?
 * Any other documentation types we should support now?
 * Any special links that come to mind?


tomaszguzialek:
The API looks clear, I like it. We also have types covered as all types of blog posts, articles, forum threads and what not are covered by type TECH. I don't think we need to make it more fine-grained. Merging.

PR #431, opened 2015-05-13
Bug/360 sort integers version 2

### LINKED ISSUE #360: Sorting in Create CSV and Create Excel works incorrectly for numeric content

tomaszguzialek:
Fixes #360 

Fixes case-sensitivity while sorting lexicographically.

### DIFF
--- a/components/writers/src/main/java/org/datacleaner/extension/output/SortHelper.java
+++ b/components/writers/src/main/java/org/datacleaner/extension/output/SortHelper.java
@@ -64,10 +64,12 @@ public class SortHelper {
                     for (int i = 0; i < numberOfSelectItems; i++) {
                         final String rowValue1 = (String) row1.getValue(i);
                         final String rowValue2 = (String) row2.getValue(i);
-                        if (CompareUtils.compare(rowValue1, rowValue2) == 0) {
+                        final String rowValue1LowerCased = rowValue1.toLowerCase();
+                        final String rowValue2LowerCased = rowValue2.toLowerCase();
+                        if (CompareUtils.compare(rowValue1LowerCased, rowValue2LowerCased) == 0) {
                             continue;
                         } else {
-                            return CompareUtils.compare(rowValue1, rowValue2);
+                            return CompareUtils.compare(rowValue1LowerCased, rowValue2LowerCased);
                         }
                     }
                 }
@@ -89,6 +91,6 @@ public class SortHelper {
         if (isDate) {
             return ConvertToDateTransformer.getInternalInstance().transformValue(value);
         }
-        return value;
+        return value.toLowerCase();
     }
 }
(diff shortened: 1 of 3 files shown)


components/writers/src/main/java/org/datacleaner/extension/output/SortHelper.java
```diff
@@ -64,10 +64,12 @@ public int compare(Row row1, Row row2) {
                     for (int i = 0; i < numberOfSelectItems; i++) {
                         final String rowValue1 = (String) row1.getValue(i);
                         final String rowValue2 = (String) row2.getValue(i);
-                        if (CompareUtils.compare(rowValue1, rowValue2) == 0) {
+                        final String rowValue1LowerCased = rowValue1.toLowerCase();
+                        final String rowValue2LowerCased = rowValue2.toLowerCase();
+                        if (CompareUtils.compare(rowValue1LowerCased, rowValue2LowerCased) == 0) {
```
LosD: If you know they are that and they are Strings, you can just use String.compareToIgnoreCase() instead of lowercasing them.

components/writers/src/main/java/org/datacleaner/extension/output/SortHelper.java
```diff
@@ -64,10 +64,12 @@ public int compare(Row row1, Row row2) {
...
                             continue;
                         } else {
-                            return CompareUtils.compare(rowValue1, rowValue2);
+                            return CompareUtils.compare(rowValue1LowerCased, rowValue2LowerCased);
```
LosD: This seems a little wasteful. Why not save the value from first compare?

LosD:
Probably a little more expensive time-wise than it needs to be. I'm not sure if it will drown in the I/O, but nevertheless I think it could be better, and the code will be a little easier to read as an added bonus.

components/writers/src/test/java/org/datacleaner/extension/output/CreateCsvFileAnalyzerTest.java
```diff
@@ -204,6 +204,58 @@ public void testSortLexicographic() throws Exception {
...
+            try (DataSet dataSet = dataContext.query().from(targetFilename).selectAll().execute()) {
+                while (dataSet.next()) {
+                    Row row = dataSet.getRow();
+                    String idValue = (String) row.getValue(0);
+                    resultIds.add(idValue);
+                }
+            }
+        }
+
+        assertEquals("[Claudia, claudia, Dennis, dennis, Kasper, kasper, Tomasz, tomasz]", resultIds.toString());
```
LosD: Hehe, I like the use of us! :)

tomaszguzialek: Need to leave some trace for future generations :).

tomaszguzialek:
Adressed the remarks, will merge when Travis is happy.

LosD:
:+1:

Issue #402, opened 2015-04-28
Copy DC configuration upon software upgrade


kaspersorensen:
We have currently achieved a good level of isolation in DC's configuration by adding the version number to the home-folder path. But we also do need to consider the burden of upgrading, which is still a bit of a pain.

In the situation where we create a new home folder we should check if home folders for previous versions exist. If it does, we should take a copy of it while still overriding:

 * Example jobs
 * Example data (customers.csv)

Furthermore we should ideally assemble a new conf.xml based on the <datastore-catalog> and <reference-data-catalog> of the old conf.xml file but everything else based on the default conf.xml file.

PR #439, opened 2015-05-16
Added XML externalization for ElasticSearch, MongoDb, CouchDb and SFDC

### LINKED ISSUE #440: Use SecurityUtils by default when externalizing datastores
Currently we have a DatastoreXmlExternalizer to do the datastore-to-xml conversion. But it by default just uses plain text passwords when doing this externalization. Probably it would be a good idea to at least encode the passwords using our SecurityUtils class.

This would need a bit of explicit testing and comparison with JaxbConfigurationReader as I am not even sure that all datastore passwords are being decoded when reading the file also.

kaspersorensen:
This will ensure that in most situations we can externalize datastores of these types to the conf.xml file, instead of saving them in userpreferences.dat.

While this is actually an independent improvement, I imagine that making XML externalization better will in turn make #402 much easier to fix because we can then focus on conf.xml and not on user preferences (more volatile and difficult to migrate)

Fixes #440

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/SecurityUtils.java
+++ b/engine/core/src/main/java/org/datacleaner/util/SecurityUtils.java
@@ -36,10 +36,6 @@ import org.datacleaner.util.ws.NaiveTrustManager;
  */
 public class SecurityUtils {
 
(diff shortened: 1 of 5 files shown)


tomaszguzialek:
LGTM!

PR #441, opened 2015-05-17
Made ResultWindow size persistent and maximized by default.

### LINKED ISSUE #434: Remember the user's preferred size of the result window
Right now the result window is always coming up with the initial size. But like the main window (AnalysisJobBuilderWindowImpl) we should remember the size of the window from last time he closed such a window and then create new windows using that size. Especially also remember if it was in maximized state.

kaspersorensen:
Fixes #434


desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
```diff
@@ -543,4 +555,17 @@ public void run() {
...
+    public void windowClosed(WindowEvent e) {
+        if (this.getExtendedState() == JFrame.MAXIMIZED_BOTH) {
+            _windowSizePreference.setUserPreferredSize(null, true);
+        } else {
+            _windowSizePreference.setUserPreferredSize(getSize(), false);
+        }
+    }
+
+    @Override
+    protected boolean maximizeWindow() {
```
tomaszguzialek: This method just gives us the state of the window - is it maximized or not. The name suggests it is changing its state to maximized.

tomaszguzialek:
One comment about the naming, the rest looks solid.

desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
kaspersorensen: Agreed, but it is coming from the Swing API (overridden method) so I cannot change that.

PR #449, opened 2015-05-19
Feature/395 expose docs in ui

kaspersorensen:
Another inching towards fixing #395 :) This fixes the HiddenProperty part as well as the need to expose the docs more prominently in the desktop UI.


kaspersorensen:
And I stumbled across a bad use of the Qualifier annotation in various property metadata annotations. Removed that aswell, but it doesn't have any meaning except for symbolically.

kaspersorensen:
Error in travis seems unrelated, probably a glitch in the travis infra

desktop/ui/src/main/java/org/datacleaner/documentation/ComponentDocumentationBuilder.java
```diff
@@ -109,9 +110,13 @@ public void write(ComponentDescriptor<?> componentDescriptor, OutputStream outpu
                         configuredProperties);
                 final List<ConfiguredPropertyDocumentationWrapper> propertyList = new ArrayList<>();
                 for (ConfiguredPropertyDescriptor property : properties) {
-                    ConfiguredPropertyDocumentationWrapper wrapper = new ConfiguredPropertyDocumentationWrapper(
-                            property);
-                    propertyList.add(wrapper);
+                    final HiddenProperty hiddenProperty = property.getAnnotation(HiddenProperty.class);
+                    // we do not show hidden properties in docs
+                    if (hiddenProperty == null) {
+                        ConfiguredPropertyDocumentationWrapper wrapper = new ConfiguredPropertyDocumentationWrapper(
```
LosD: Let's add final now that we're reformatting anyway :)

LosD:
LGTM. You can add a final if you feel like it :)

PR #453, opened 2015-05-21
Issue #452: Fixed reading of analysis.xml files with chained filters

### LINKED ISSUE #452: JaxbJobReader sometimes fails to find the right component builder when setting requirements
I think we just discovered a snag in the job XML reader, probably introduced by #422:

This code was changed:
```
-            for (Object o : transformersAndFilters) {
-                if (o instanceof TransformerType) {
-                    ref = ((TransformerType) o).getRequires();
-                    if (ref != null) {
-                        final TransformerComponentBuilder<?> builder = transformerJobBuilders.get(o);
-                        final ComponentRequirement requirement = getRequirement(ref, outcomeMapping);
-                        builder.setComponentRequirement(requirement);
-                    }
-                } else if (o instanceof FilterType) {
-                    ref = ((FilterType) o).getRequires();
-                    if (ref != null) {
-                        final FilterComponentBuilder<?, ?> builder = filterJobBuilders.get(o);
-                        final ComponentRequirement requirement = getRequirement(ref, outcomeMapping);
-                        builder.setComponentRequirement(requirement);
-                    }
-                } else {
-                    throw new IllegalStateException("Unexpected transformation child element: " + o);
```

To:
```
+            for (ComponentType o : transformersAndFilters) {
+                ref = o.getRequires();
+                if (ref != null) {
+                    final TransformerComponentBuilder<?> builder = transformerJobBuilders.get(o);
+                    final ComponentRequirement requirement = getRequirement(ref, outcomeMapping);
+                    builder.setComponentRequirement(requirement);
```

And if you look very closely, then we see that it is calling ```transformerJobBuilders.get(o)``` for all components - also filters! This will resolve to null then if there is a filter with a requirement on it. And eventually a NPE will be thrown.

Therefore this must mean that if a filter depends on a filter outcome of another filter, then it breaks.

kaspersorensen:
Fixes #452

### DIFF
--- a/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
+++ b/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
@@ -61,6 +61,7 @@ import org.datacleaner.data.MutableInputColumn;
 import org.datacleaner.descriptors.AnalyzerDescriptor;
 import org.datacleaner.descriptors.ComponentDescriptor;
 import org.datacleaner.descriptors.ConfiguredPropertyDescriptor;
+import org.datacleaner.descriptors.DescriptorProvider;
 import org.datacleaner.descriptors.FilterDescriptor;
 import org.datacleaner.descriptors.TransformerDescriptor;
 import org.datacleaner.job.builder.AnalysisJobBuilder;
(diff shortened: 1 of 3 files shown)


LosD:
This is in code where I'm not entirely sure how stuff is supposed to work, and it's a bit hard to understand because of the monster method (300+ lines), but on the surface it looks good.

If you want to make it even better, I'd split it up :)

Issue #443, opened 2015-05-18
Add all tabs to result window as soon as possible


kaspersorensen:
Good suggestion by @LosD: Instead of adding tabs to the result window while the job is running, it would make the experience less "jumpy" if we already from the start produce the tabs that should be available when the job is finished.

An addition from my side: Would be neat if they are initially then disabled/grayed and then as the results become ready we will make them enabled.

This feature would also make sorting of the result tabs more consistent and deterministic - so it fixes #286

ClaudiaPHI:
:+1:

Issue #451, opened 2015-05-20
Charts on results looks stretched


kaspersorensen:
Some of the result report charts looks kinda stretched. This probably depends on the amount of categories there are in the data, but anyway we should be able to deliver a better rendering than these examples:

![image](https://cloud.githubusercontent.com/assets/291450/7727682/a6f852dc-ff0a-11e4-8b70-9899f02d2d24.png)

...

![image](https://cloud.githubusercontent.com/assets/291450/7727689/b1637864-ff0a-11e4-896a-34899c840c76.png)

PR #454, opened 2015-05-21
Issue #442: Made a new menu item/button for each component/result.

### LINKED ISSUE #442: Add a new tab to ResultWindow for each component that produces a result
Instead of grouping results per _component type_ we would prefer to just have a new tab for each component instance. This would mean a lot more tabs, but since the tabs are now shown vertically we can also better scale the UI here. It will relieve us from having to use the task panes that each result is currently placed in, and as such provide more screen real-estate.

kaspersorensen:
Fixes #442 

Would be interested in review remarks both on code but certainly also on design. Will post some screenshots in the comments here.

__do not merge in it's current state__

Also alleviated the stretching issue for pattern finder (see #451) while it is still not fixed, only updated a bit for higher resolutions.

### DIFF
--- a/components/visualization/src/main/scala/org/datacleaner/visualization/DensityAnalyzerResultSwingRenderer.scala
+++ b/components/visualization/src/main/scala/org/datacleaner/visualization/DensityAnalyzerResultSwingRenderer.scala
@@ -79,7 +79,7 @@ class DensityAnalyzerResultSwingRenderer extends Renderer[DensityAnalyzerResult,
 
     ChartUtils.applyStyles(chart);
 
-    val chartPanel = new ChartPanel(chart)
+    val chartPanel = ChartUtils.createPanel(chart, true);
 
     chartPanel.addChartMouseListener(new ChartMouseListener() {
       override def chartMouseClicked(event: ChartMouseEvent) = {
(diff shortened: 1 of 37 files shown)


kaspersorensen:
image

kaspersorensen:
image
(notice the horizontal scroll which I allowed (what do you think?) for wide result renderings).

kaspersorensen:
The pattern finder result now has a pretty big chart...
image

kaspersorensen:
Currently a problem: The dedup result display is typically too wide. This is a case of where it would be good to disable the horizontal scroll.

image

kaspersorensen:
The same applies to training mode. Here we also see that the background color has changed so the existing renderer probably needs to be adjusted:

image

kaspersorensen:
I changed the look and feel of the tabs, so that they look fine on the white background.
I also removed the horizontal scroll.

Looks good in training mode:
image

But for some reason it still looks strange with the dup. detection result. It seems that it is not fitting the result into the screen even though the horizontal scrolls are gone:

image

kaspersorensen:
There... Now that is also fixed. I am feeling done with this story, so please review.

By now I am also happy to say that it not only fixes #442 but also #443:

Fixes #443

ClaudiaPHI:
LGTM!

PR #455, opened 2015-05-22
JaxbJobReader refactor

kaspersorensen:
Refactored the JaxbJobReader to have more isolated methods and to reuse methods for component initialization

Following previous comments that this class is hard to read and has too long methods. I think this helps a lot. Also managed to reuse a lot of stuff around (generic) component initialization instead of having special treatment for filters, analyzers and transformers.


LosD:
LGTM

PR #465, opened 2015-06-01
Issue #464: Introduced more meaningful column-id generation scheme

### LINKED ISSUE #464: Try to use meaningful column IDs in XML marshalling of jobs
Right now we always generate column IDs in job XML files like col_1, col_2 etc. But it would be nice to at least attempt to make meaningful IDs. I guess we could convert the column name to lowercased alphanumerical strings. And then check that they are unique and non-blank. If so, use that name, or else use an artificial name.

This will prevent a lot of annoying diff issues when changing how jobs are built.

kaspersorensen:
Fixes #464


ClaudiaPHI:
LGTM :-)

PR #474, opened 2015-06-14
Excluded hidden columns from "select all" & drill-to-detail windows.

### LINKED ISSUE #462: Invisible columns sometimes by mistake are fed into other components
For any transformer, the user has the possibility to hide output columns by unchecking them in the transformers configuration, as per below screenshot.
2015-05-28 16_10_55-phone correction _ datacleaner.

The behavior of invisible columns is inconsistent: sometimes you don't see them, sometimes they pop up. In it's current shape, it is confusing and the user may fail to configure his job correctly.

Road show of the problem:
* Run email standardizer on an email field.
* Untick the Domain field so it is invisible
* Add String analyzer.
* In the configuration, click 'Select all'. Notice that 'Domain' is not shown in the column list
* Run the job. In the analysis result, notice that 'Domain' is present in the output.
* Go back to the config. "Domain' is still not shown in the column list, so I cannot even disable it.
* Also notice that the invisible column IS visible in the "Reorder" dialog.

kaspersorensen:
Fixes #462


LosD:
LGTM!

PR #475, opened 2015-06-14
Feature/drop table in ui

kaspersorensen:
Made it possible to right click a table and drop it from the UI.
Also fixed a bad layout issue with the 'Create staging table' property panes.


desktop/ui/src/main/java/org/datacleaner/widgets/tree/TableMouseListener.java
```diff
@@ -158,6 +162,24 @@ public void actionPerformed(ActionEvent e) {
...
+                        @Override
+                        public void actionPerformed(ActionEvent e) {
+                            final DropTableDialog dialog = new DropTableDialog(windowContext, updateableDatastore,
+                                    table, _schemaTree);
+                            dialog.open();
+                        }
+                    });
+                    popup.add(dropTableMenuItem);
+                }
+
```
LosD: If possible, it would be nice to get some of this code out of this event handler. It is already waaaaayyyy too complex.

engine/core/src/main/java/org/datacleaner/util/SchemaComparator.java
```diff
@@ -53,7 +54,7 @@ public static boolean isInformationSchema(Schema schema) {
 		if (name == null) {
 			return false;
 		}
-		return "information_schema".equals(name.toLowerCase());
+		return MetaModelHelper.isInformationSchema(schema);
```
LosD: MetaModelHelper already does the name check and lowercasing, so no need to repeat it.

LosD:
A few very minor things, but overall looks good

kaspersorensen:
Fixed the review remarks. Updated the icons too:
image

PR #476, opened 2015-06-15
Create table in ui

### LINKED ISSUE #7: Create table via DC's UI
Make it possible to create tables on (at least JDBC) datastores, since this is a common wish when wanting to insert data using the "Insert into table" writer. Currently you can only use existing tables or use the "Create staging table" component, but that is not always as convenient.

kaspersorensen:
This makes it possible to create new tables in the UI of DataCleaner. Currently by right-clicking a schema in the schema tree. I would also like to make it possible to create tables from the table-selection comboboxes in e.g. "Insert into table" component. But that could be defined as a separate story even.

Fixes #7 
Please review #475 first, since this branch builds upon that branch.

Here's a few screenshots:

image

image

(the banner title in the second screenshot says "Drop table" - that has been corrected to "Create table" of course!)


LosD:
Code looks good, but we probably need to work around the MetaModel issue where creating a new table in a CSV file will automatically and without warning nuke the old one.

kaspersorensen:
Updated. Please re-check :)

LosD:
A little hard to read because of GitHub's (Git's?) strange diff, but looks good!

PR #479, opened 2015-06-16
Feature/468 dynamic analyzer name in result screen

### LINKED ISSUE #468: Dynamic analyzer name in the analyzer result screen
As far as I am aware, the name displayed in the analyzer result screen is created from the analyzer name. For clarity, this is what I'm talking about:
analyzer result name

This fixed behaviour is problematic in analyzers for which the name depends on some settings. For example, the Identify7 analyzer has multiple modes. Even if selecting 'training' mode, the analyzer result still gets listed as 'duplicate detection'. As this is the name of another mode, the overall result is confusion.

stefanjanssen:
Made labels in the result screen use the hasLabelAdvice interface. Fixes #468 .

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/panels/result/AnalyzerResultPanel.java
+++ b/desktop/ui/src/main/java/org/datacleaner/panels/result/AnalyzerResultPanel.java
@@ -66,6 +66,7 @@ public class AnalyzerResultPanel extends DCPanel implements Scrollable {
     private final ProgressInformationPanel _progressInformationPanel;
     private final ComponentJob _componentJob;
     private final LoadingIcon _loadingIcon;
+    private final JLabel _headerLabel;
 
     public AnalyzerResultPanel(RendererFactory rendererFactory, ProgressInformationPanel progressInformationPanel,
             ComponentJob componentJob) {
@@ -81,14 +82,14 @@ public class AnalyzerResultPanel extends DCPanel implements Scrollable {
         final Icon icon = IconUtils.getDescriptorIcon(descriptor, IconUtils.ICON_SIZE_TASK_PANE);
 
         final String headerText = getHeaderText();
-        final JLabel header1 = createHeader(icon, headerText, WidgetUtils.FONT_HEADER1, WidgetUtils.BG_COLOR_DARK);
-        final JLabel header2 = createHeader(null, getSubHeaderText(componentJob, headerText), WidgetUtils.FONT_SMALL,
+        _headerLabel = createHeader(icon, headerText, WidgetUtils.FONT_HEADER1, WidgetUtils.BG_COLOR_DARK);
+        final JLabel subheader = createHeader(null, getSubHeaderText(componentJob, headerText), WidgetUtils.FONT_SMALL,
                 WidgetUtils.BG_COLOR_BLUE_MEDIUM);
 
         final DCPanel headerPanel = new DCPanel();
         headerPanel.setLayout(new VerticalLayout(4));
-        headerPanel.add(header1);
-        headerPanel.add(header2);
+        headerPanel.add(_headerLabel);
+        headerPanel.add(subheader);
         headerPanel.setBorder(new EmptyBorder(0, 0, 4, 0));
         add(headerPanel, BorderLayout.NORTH);
 
@@ -99,6 +100,7 @@ public class AnalyzerResultPanel extends DCPanel implements Scrollable {
     public void setResult(final AnalyzerResult result) {
         final String headerText = getHeaderText();
         _progressInformationPanel.addUserLog("Rendering result for " + headerText);
+        _headerLabel.setText(LabelUtils.getLabel(_componentJob, result, false, false, false));
 
         // use a swing worker to run the rendering in the background
         new SwingWorker<JComponent, Void>() {
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/tabs/VerticalTabbedPane.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/tabs/VerticalTabbedPane.java
@@ -56,8 +56,7 @@ public class VerticalTabbedPane extends DCPanel {
 
     private static final Color COLOR_SELECTED_FOREGROUND = WidgetUtils.BG_COLOR_BLUE_DARK;
     private static final Color COLOR_SELECTED_BACKGROUND = WidgetUtils.BG_COLOR_LESS_BRIGHT;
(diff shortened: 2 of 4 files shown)


desktop/ui/src/main/java/org/datacleaner/panels/result/AnalyzerResultPanel.java
```diff
@@ -99,6 +100,7 @@ public AnalyzerResultPanel(RendererFactory rendererFactory, ProgressInformationP
     public void setResult(final AnalyzerResult result) {
         final String headerText = getHeaderText();
         _progressInformationPanel.addUserLog("Rendering result for " + headerText);
+        _headerLabel.setText(LabelUtils.getLabel(_componentJob, result, false, false, false));
```
kaspersorensen: I guess you don't need to move this logic down here - the ComponentJob is also available in the constructor, so why not just do it immediately and not change the label later?

engine/core/src/main/java/org/datacleaner/util/LabelUtils.java
```diff
@@ -96,13 +97,26 @@ public static String getLabel(ComponentJob job) {
...
+    public static String getLabel(ComponentJob job, AnalyzerResult result, boolean includeDescriptorName, boolean includeInputColumnNames,
             boolean includeRequirements) {
         final String jobName = job.getName();
         final StringBuilder label = new StringBuilder();
         if (Strings.isNullOrEmpty(jobName)) {
             ComponentDescriptor<?> descriptor = job.getDescriptor();
-            label.append(descriptor.getDisplayName());
+            String displayName = descriptor.getDisplayName();
+            if (result != null) {
+                if (result instanceof HasLabelAdvice) {
```
kaspersorensen: Normally HasLabelAdvice is applied to the component, not the result. Do we need to change that? I worry if it will create confusion about which label will be used - the component's label or the result's label.

If we decide to also support it on the result class, then we should at least update the HasLabelAdvice javadoc accordingly.

desktop/ui/src/main/java/org/datacleaner/widgets/tabs/VerticalTabbedPane.java
```diff
@@ -56,8 +56,7 @@
 
     private static final Color COLOR_SELECTED_FOREGROUND = WidgetUtils.BG_COLOR_BLUE_DARK;
     private static final Color COLOR_SELECTED_BACKGROUND = WidgetUtils.BG_COLOR_LESS_BRIGHT;
-    private static final Border BORDER_TABS = new CompoundBorder(WidgetUtils.BORDER_LIST_ITEM_SUBTLE, new EmptyBorder(
-            10, 4, 10, 4));
+    private static final Border BORDER_TABS = new CompoundBorder(WidgetUtils.BORDER_LIST_ITEM_SUBTLE, new EmptyBorder(10, 4, 10, 4));
```
kaspersorensen: I see quite a lot of these diffs caused by different formatter settings. I think we normally use a line width of 140 chars - how wide are your settings?

desktop/ui/src/main/java/org/datacleaner/panels/result/AnalyzerResultPanel.java
stefanjanssen: Yes ComponentJob is, but AnalyzerResult is not. The HasLabelAdvice interface is implemented by the AnalyzerResult, not the ComponentJob. So first I set the label without the result using the logic that was already there, and later I update it with the analyzerresult (if the result has the interface, otherwise it is just the same).

engine/core/src/main/java/org/datacleaner/util/LabelUtils.java
stefanjanssen: The related code has nothing related to the component, so I don't think that will result in any confusion. Or atleast, I couldn't find a way to go from the ComponentJob to the Component without breaking many layers of abstraction. Should probably update the javadoc though.

desktop/ui/src/main/java/org/datacleaner/widgets/tabs/VerticalTabbedPane.java
stefanjanssen: Not 140  :stuck_out_tongue_winking_eye:

engine/core/src/main/java/org/datacleaner/util/LabelUtils.java
kaspersorensen: I can see your point that it's difficult to invoke any method on the component object from where you are in the code here. But ideally speaking the DC framework should take the burden away from the extension developer and I guess also it would be preferable if the label/name of a component is represented in the same way on the desktop UI graph as well as in the result screen. Currently the HasLabelAdvice does influence the label/name on the graph - just not the result screen. So having a (potentially different) label on the result screen could end up feeling a bit chaotic/inconsistent.

The ComponentJob is what DC uses (mostly internally although it has a public interface) to represent the component and it's configuration without actually instantiating it. At the same time the component itself and HasLabelAdvice is implemented by the extension developer, which probably doesn't care about the ComponentJob representation.

One aspect of ComponentJob is that it has a getName() method. We could consider if we wanted to apply a name to the ComponentJob at the time of building it (at that time there _is_ a component instance available). That way the LabelUtils class can stay as it was because it is already checking if there is a name. If we can go down this route to solving the issue, then you should be looking into ComponentBuilder and it's implementing classes.

desktop/ui/src/main/java/org/datacleaner/widgets/tabs/VerticalTabbedPane.java
kaspersorensen: Hehe could we change that? I don't mind increasing the line width actually, but probably shouldn't be done on a single file+contributor at a time.

engine/core/src/main/java/org/datacleaner/util/LabelUtils.java
stefanjanssen: The solution with getName sounds nice to me if there are not too many implementing classes. Will look into that. It would also avoid the sudden updating of the tab name that currently happens. Also, for the record, I blame all misdirection on the hints provided by someone in #468 .

kaspersorensen: I'm probably guilty as charged. That's what you get for trying to explain something, I better shut up next time ;-)

I think there are only 3 implementing classes where the name stuff is used/produced (one for each of Transformer, Filter and Analyzer).

stefanjanssen:
Closing as I implemented a nicer solution in #484 .

PR #481, opened 2015-06-16
Made it possible to create tables from the "table name" property widget

### LINKED ISSUE #480: Offer a "create table" option when selecting table in "Insert into table" component dialog

kaspersorensen:
Fixes #480 

image


LosD:
In this step we already have the probably column names and types the user wants, so it should be possible to pre-fill the dialog, which should create a nice user experience

kaspersorensen:
Done :-) Please sign it off if you agree to the solution :)

LosD:
LGTM!

PR #486, opened 2015-06-18
Made labels in the result screen use the hasLabelAdvice interface. 

### LINKED ISSUE #468: Dynamic analyzer name in the analyzer result screen

stefanjanssen:
Fixes #468 . Instead of the previous PR. Three times is a charm or something.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/LabelUtils.java
+++ b/engine/core/src/main/java/org/datacleaner/util/LabelUtils.java
@@ -26,12 +26,15 @@ import java.util.Date;
 
 import org.datacleaner.api.HasLabelAdvice;
 import org.datacleaner.api.InputColumn;
+import org.datacleaner.configuration.DataCleanerConfiguration;
 import org.datacleaner.descriptors.ComponentDescriptor;
+import org.datacleaner.job.AnalysisJob;
 import org.datacleaner.job.AnalyzerJob;
 import org.datacleaner.job.AnyComponentRequirement;
 import org.datacleaner.job.ComponentJob;
 import org.datacleaner.job.ComponentRequirement;
 import org.datacleaner.job.builder.ComponentBuilder;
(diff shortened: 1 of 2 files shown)


engine/core/src/main/java/org/datacleaner/util/LabelUtils.java
```diff
@@ -102,7 +105,19 @@ public static String getLabel(ComponentJob job, boolean includeDescriptorName, b
...
             ComponentDescriptor<?> descriptor = job.getDescriptor();
-            label.append(descriptor.getDisplayName());
+            String baseName = descriptor.getDisplayName();
+            if (ReflectionUtils.is(descriptor.getComponentClass(), HasLabelAdvice.class)) {
+                try {
+                    HasLabelAdvice c = (HasLabelAdvice) descriptor.newInstance();
+                    LifeCycleHelper lch = new LifeCycleHelper((DataCleanerConfiguration) null, (AnalysisJob) null,
+                            false);
+                    lch.assignConfiguredProperties(descriptor, c, job.getConfiguration());
+                    baseName = c.getSuggestedLabel();
```
kaspersorensen: this could return null, so I guess a small addition would be to check that it is not (empty string or) null so that we don't get null-labels

kaspersorensen:
Looks good to me except for the little gotcha wrt. null.

stefanjanssen:
I added a null/empty check.

kaspersorensen:
LGTM!

Issue #430, opened 2015-05-12
Possibility to surpress the "Save Result" buttons / Minimal analyzer result screen


stefanjanssen:
Currently DataCleaner always shows several button above analyzer results. While these make sense for most analyzers, for some it does not. Examples are ID7 training tool, but also write to file like analyzers. If there would be a possibility to hide these buttons, it would result in a cleaner, less confusing UI for users.

Ideally, the buttons would not only be hidden, but the datacleaner banner in general would be minimized, to allow maximum screen space for the analyzer result itself.

stefanjanssen:
So I made a quick paint edit of how this more minimal screen could look:

![modified_result_screen](https://cloud.githubusercontent.com/assets/6106276/7682213/c54f97c6-fd76-11e4-90f2-06a585326e09.png)

kaspersorensen:
Oh I think this actually is fixed, it just never got closed!

PR #487, opened 2015-06-25
Feature/430 minimal result screen

stefanjanssen:
Some work towards #430 . Not entirely where I would like it to be, but it is a first start.

2015-06-25 11_06_48-customer profiling analysis xml _ customers _ analysis results _ datacleaner

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
@@ -40,6 +40,7 @@ import javax.swing.Icon;
 import javax.swing.JButton;
 import javax.swing.JComponent;
 import javax.swing.JFrame;
+import javax.swing.SwingConstants;
 import javax.swing.SwingUtilities;
 import javax.swing.border.Border;
 import javax.swing.border.CompoundBorder;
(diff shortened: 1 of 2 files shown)


drexler42:
Its a start, indeed. If I had my way, we would hide these buttons even further. Like having just a single button in the left hand side: "DataCleaner Monitor", and hide the three buttons behind that.

kaspersorensen:
Or how about just a single button with label "Save result" and when clicked a popup menu would show with these options:

 * Save as result file
 * Export as HTML report
 * Publish to server

Question on scalability of the design: How does it behave if I have many results in the job (or just a pretty small window). I would prefer it if a scrollbar shows next to the result elements but that this scrollbar ends just above the "Save result" button so that it is fixed in the bottom-left corner.

stefanjanssen:
Alright, I changed the three buttons to a single popupbutton and improved the scroll thing so the button is always fixed.

2015-07-06 14_43_28-customer profiling analysis xml _ customers _ analysis results _ datacleaner

And a pretty small window:
2015-07-06 14_45_09-customer profiling analysis xml _ customers _ analysis results _ datacleaner

kaspersorensen:
Unable to review the code right now (stepping on to a plane) but the screenshots look good and in line with what I was thinking.

I recall there's also a static list of 'pluggable' buttons. Is that also covered?

stefanjanssen:
I didn't touch those: they should still appear on their old location.

desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
```diff
@@ -359,15 +398,18 @@ protected JComponent getWindowContent() {
...
+        _tabbedPane.addListener(new VerticalTabbedPane.Listener() {
+            @Override
+            public void stateChanged(int newIndex, Tab<?> newTab) {
+                banner.setTitle2(newTab.getTitle());
+                banner.updateUI();
+            }
+        });
 
         for (JComponent pluggableButton : _pluggableButtons) {
             banner.add(pluggableButton);
```
kaspersorensen: Here the pluggable buttons are added to the banner. You didn't touch those, but maybe you should? :)
At least the only pluggable button we right now (in commercial editions) have is the "export metrics to csv" button (I think it's called that - or something very similar).... I think that fits in best with the other buttons that you moved.

kaspersorensen: Or alternatively: Make the pluggable buttons deprecated and expose a similar API for adding pluggable menu items.

stefanjanssen: I did see those, but I didn't see anything where those were used. Where exactly is it used in commercial editions? In terms of API to be changed, strictly speaking it wouldn't be pluggable jbuttons anymore, but pluggable jmenuitems. Also, would it make sense to redo the 'publish to server' and maybe 'export to html' buttons as a pluggable button aswell?

kaspersorensen: Yes those are also good improvements, but I am only trying to make it simple for you to get this PR merged :-)

Currently the only pluggable button is this one:

image

kaspersorensen: I would suggest the following:

 * Add a new field for pluggable menu items
 * Deprecate the old field for pluggable buttons
 * Convert the buttons into menu items by taking the button's text, icon and action listener.

stefanjanssen:
Moved the 'cancel job' and pluggable buttons. Notice that the interface for the pluggable buttons is  quite awkward, as it is a static list that  contains JComponents. Originally it added all components as is, but I can't do that anymore, so anything that is not a JMenuItem or Jbutton gets ignored.

kaspersorensen:
Good work, thanks. I'll merge.

PR #493, opened 2015-07-08
Put latch countdowns in finally block.

### LINKED ISSUE #492: JobCompletionTaskListener should protect latch countdowns
In JobCompletionTaskListener, there is a countdownlatch used for halting an awaiting running jobs' completion. It is important that all uses of the latch are highly protected, so that errors does not affect the countdown of the latch.

Unfortunately, the current implementation is a bit naïve, which makes it pretty sensitive to errors in the in the analysis listeners.

LosD:
This avoids error in external code from affecting the countdown latches.

Fixes #492


kaspersorensen:
That's a good fix!

PR #495, opened 2015-07-08
HDFS resource support

kaspersorensen:
This PR proposes a way to add HdfsResource support and in general to improve the desktop UI with a better resource selector widget, and to use it for configuring CSV file source in the CsvDatastoreDialog.


LosD:
Ouch, big! The last two commits seems like they belong in their own PRs?

kaspersorensen:
Hmm yeah I guess you could argue that there is a separate "engine" and "desktop" story here. Will try and split off the engine part only in a separate PR.

kaspersorensen:
Please see #502 first then. If that is merged we'll have to then merge master back into this branch I assume.

tomaszguzialek:
When switching to URL in the resource type, there is no field to type a URL...

image

kaspersorensen:
Updated it quite a lot now ... Could you try again?

desktop/ui/src/main/java/org/datacleaner/windows/CsvDatastoreDialog.java
```diff
@@ -217,79 +217,94 @@ protected String getBannerTitle() {
...
-                _escapeCharField.setSelectedItem(ESCAPE_BACKSLASH);
+                showPreview = true;
+                validateAndUpdate();
             }
-        }
-
-        showPreview = true;
-        validateAndUpdate();
+        }.execute();
+        ;
```
ClaudiaPHI: slet, slet, slet :-D

kaspersorensen:
I'm assuming you didn't find anything else in your review @ClaudiaPHI? Gonna merge it then.

PR #499, opened 2015-07-10
Issue #396: Added country output format names

### LINKED ISSUE #396: Improve naming of "ISO2" and "ISO3" in Country standardizer
The country standardizer has output formats as "ISO2" and "ISO3" which actually mean "2/3-letter ISO code". Or rather alpha-2 and alpha-3 of ISO 3166-1 *.

We should let the enum implement HasName so that we can present a humanly readable name that makes more sense in the UI. My suggested names would be:

 * 2-letter ISO country code
 * 3-letter ISO country code
 * Country name

* see http://en.wikipedia.org/wiki/ISO_3166-1_alpha-2 and http://en.wikipedia.org/wiki/ISO_3166-1_alpha-3

kaspersorensen:
Fixes #396

image


ClaudiaPHI:
LGTM!

PR #501, opened 2015-07-10
Issue #500: Added custom endpoint url option to SFDC datastore

### LINKED ISSUE #500: Add endpoint URL option to Salesforce datastore
At the engine level we can decide a SFDC endpoint URL (for instance to use their sandbox environment instead of production). That would be very useful and should be configurable along with the existing username, password and security token.

kaspersorensen:
Fixes #500


LosD:
The reformatting of the XSD seems rather broken. At least the documentation parts are worse than before.

LosD:
Besides that LGTM

PR #502, opened 2015-07-10
HdfsResource in engine of DC

kaspersorensen:
This is a part of #495 - it will add HDFS support to CSV datastore and other when configured via conf.xml or programmatically.


LosD:
LGTM!

Issue #506, opened 2015-07-16
Pattern Finder memory footprint too big?


kaspersorensen:
Have a few reports of the pattern finder easily taking up a lot of memory. Resulting in OutOfMemoryError with either "heap space" message or "GC overhead limit exceeded".

Example stack:

```
java.lang.OutOfMemoryError: GC overhead limit exceeded
	at java.util.Arrays.copyOfRange(Unknown Source)
	at java.lang.String.<init>(Unknown Source)
	at java.lang.StringBuilder.toString(Unknown Source)
	at org.datacleaner.beans.stringpattern.SimpleToken.getString(SimpleToken.java:45)
	at org.datacleaner.beans.stringpattern.DefaultTokenizer.preliminaryTokenize(DefaultTokenizer.java:180)
	at org.datacleaner.beans.stringpattern.DefaultTokenizer.tokenizeInternal(DefaultTokenizer.java:107)
	at org.datacleaner.beans.stringpattern.DefaultTokenizer.tokenize(DefaultTokenizer.java:100)
	at org.datacleaner.beans.stringpattern.PatternFinder.run(PatternFinder.java:74)
	at org.datacleaner.beans.stringpattern.PatternFinderAnalyzer.run(PatternFinderAnalyzer.java:201)
	at org.datacleaner.beans.stringpattern.PatternFinderAnalyzer.run(PatternFinderAnalyzer.java:196)
```

I recall recently optimizing the concurrency of Pattern Finder by using ConcurrentHashMap. Maybe this added also additional memory usage because of too wasteful use of e.g. "putIfAbsent" instead of "contains check + put" (which instantiates new patterns later).

PR #513, opened 2015-07-17
Issue #506: Implemented a lower-footprint row annotation factory

kaspersorensen:
Attempt to "fix"/improve #513

So far I did not find any obvious greedy "mistakes" or something like that in the Pattern Finder. But I did find that it depends a lot on the RowAnnotationFactory. And here we might improve a bit on the memory footprint.

Note: I am still testing this solution, do not merge yet. Will post test results on this PR's comments.


kaspersorensen:
__results of using this PR__

 * Scenario 1:
  * 1.8 M records with 35 fields
  * 4 pattern finders used on the first 4 fields
  * Execution time: 1,172 min
  * Allocated memory after run and after GC: 1,467,904 gb

kaspersorensen:
__results from latest DC release (4.0.8)__

 * Scenario 1:
  * (same input file, job etc.)
  * Execution time: 1,065 min
  * Allocated memory after run and after GC: 1,559,552 kb

kaspersorensen:
New test run, 1,5 M record file with 5 fields (the old file had a bunch of inconsistencies which made it a strage test file)

After the new PR:
image

kaspersorensen:
With the previous implementation we're above the 750 MB tick:

image

kaspersorensen:
More importantly, the new implementation let's you tweak it. In conf.xml you can now play around with these values to increase/decrease sampling in memory:

```
			<row-annotation-storage>
				<in-memory max-rows-threshold="500" max-sets-threshold="500"/>
			</row-annotation-storage>
```

kaspersorensen:
Case in point, here's the memory profile when running with just 10 sets and a threshold of 10 records:

image

stefanjanssen:
About the last two evaluations: If I understand the threshold thing right, it seems as if the rows are in memory even when there is no job running?

engine/core/src/main/java/org/datacleaner/storage/AbstractRowAnnotationFactory2.java
```diff
@@ -0,0 +1,73 @@
...
+package org.datacleaner.storage;
+
+import java.util.List;
+
+import org.datacleaner.api.InputRow;
+
+/**
+ * Abstract implementation of {@link RowAnnotationSampleContainer}
+ */
+public abstract class AbstractRowAnnotationFactory2 implements RowAnnotationFactory, RowAnnotationSampleContainer,
```
stefanjanssen: Maybe a more descriptive name than Factory2 would be nice? The javadoc doesn't give any insight in why this is number two either.

engine/core/src/main/java/org/datacleaner/storage/InMemoryRowAnnotationFactory2.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    private Collection<InputRow> getInputRowCollection(int defaultSize, RowAnnotation annotation) {
+        List<InputRow> rowCollection = _storage.get(annotation);
+        if (rowCollection == null) {
+            if (_storage.size() >= _maxSampleSets) {
+                return null;
+            }
+            rowCollection = new ArrayList<InputRow>(defaultSize);
```
stefanjanssen: Strictly speaking this is not properly concurrent: two threads can modify the same ArrayList at the same time. ArrayList is not synchronized.

stefanjanssen:
Other than the remarks above, the changes look good to me. If you aim at reducing the memory footprint even further I would probably take a look at the InputRow class, as this class is (by nature) referencing a lot of those.

engine/core/src/main/java/org/datacleaner/storage/InMemoryRowAnnotationFactory2.java
stefanjanssen: Also, ConcurrentHashMap has a setting for the maximum number of concurrent modifications, which you can tweak (or just put to one) to optimize memory usage as it is doing a concurrency/memory trade off.

kaspersorensen:
Good remarks. Thanks.

PR #523, opened 2015-07-23
Feature/507 hive support

### LINKED ISSUE #507: Add datastore icon and dialog for Apache Hive connection
Hive is using the Jdbc module of MetaModel, so reuse the JDBC datastore dialog, but add a separate button on the datastore mgmt. screen and in the "database" dropdown of the source screen.

Also add a DatabaseConnectionPresenter for Hive.

Tip: If you look into the current master branch of MetaModel you can the integration-test properties file which has sample connection URLs, driver class name etc.

kaspersorensen:
Fixes #507 

As can be seen in the diff, I took also the liberty to update a few existing JDBC driver names, download links (updated version) etc. but that part should be trivial.


ClaudiaPHI:
LGTM!

PR #524, opened 2015-07-23
Feature/491 http components upgrade

### LINKED ISSUE #491: Roll HttpComponent version forward
The httpcomponents.version number appears to collide with the correct version of apache http core when using apache http fluent to make rest calls.  It is an old enough version to collide with newer packages relying on apache http components.  Currently it's 4.2.5 many of the apache http packages are up to 4.4+ now.

kaspersorensen:
Fixes #491

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/extensionswap/ExtensionSwapClient.java
+++ b/desktop/ui/src/main/java/org/datacleaner/extensionswap/ExtensionSwapClient.java
@@ -23,7 +23,8 @@ import java.io.File;
 import java.util.List;
 
 import org.apache.commons.vfs2.FileObject;
-import org.apache.http.client.HttpClient;
+import org.apache.http.impl.client.CloseableHttpClient;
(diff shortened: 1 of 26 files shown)


desktop/ui/src/main/java/org/datacleaner/extensionswap/ExtensionSwapClient.java
```diff
@@ -23,7 +23,8 @@
 import java.util.List;
 
 import org.apache.commons.vfs2.FileObject;
-import org.apache.http.client.HttpClient;
+import org.apache.http.impl.client.CloseableHttpClient;
```
LosD: Using an impl package from outside the project usually worries me, most projects has internal classes in there. Are you sure this is being used right?

Of course, it might just be the Http project that doesn't follow that :)

LosD:
Wow, quite big changes for a minor version change. Looked further into the .impl. package, and it seems like it is just how they expect it to be used.

LGTM!

PR #526, opened 2015-07-23
Feature/324 execute button options

### LINKED ISSUE #324: Alternative execution button options
We currently just have a single execution button, but sometimes we get requests like:

 * Could we run the job just on a sample of the data (kinda like a quick preview)
 * How to trigger the job to run clustered
 * How to trigger the job to run remotely
 * How to run the job with a custom amount of threads (for benchmarking etc).

We'd still want the simple execute button as default, but offer more options in a drop-down button a la this mash-up:

image

kaspersorensen:
Fixes #324 

Note: This wasn't quite as trivial as I thought, had to redo a bunch of the Dependency Injection stuff with Guice. Good thing is that I feel the DCModule is a better object to pass around than Injector and InjectorBuilder, and that's now the case.


desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -819,6 +833,75 @@ public void actionPerformed(ActionEvent e) {
...
+    }
+
+    private ActionListener executePreview() {
+        return new ActionListener() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                final String maxRowsString = JOptionPane.showInputDialog("How many records do you want to process?",
+                        "100");
+                final Number maxRows = ConvertToNumberTransformer.transformValue(maxRowsString);
+                if (maxRows == null) {
```
LosD: What happens if I put garbage in there? Or i.e. -1?

LosD:
I'm a little worried about the dialog for numbers, otherwise, LGTM

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
kaspersorensen: Right ... Garbage would become null, but -1 would in deed become -1, so we should check for that.
Feel free to take over the branch since I'm going out for a few weeks ;-)

LosD: No working in the vacation? You lazy bastard :)

I'll just dump a ´´´|| maxRows < 1```in there (I guess 0 rows doesn't really make sense either).

LosD:
Issue with zero or negative number fixed. Merging.

LosD:
*facepalm* I didn't push first! :(

LosD:
That'll have to be on master.

PR #529, opened 2015-07-24
Progress bar fix

### LINKED ISSUE #514: Sometimes progress bar shows "Integer.MAX_VALUE" as expected records
This happens inconsistently:

image

I think this is somewhere that our listener might get multiple events and then a race condition can occur. Or maybe it is in the progressbar itself because I recall it is also doing some magic with Math.max or something like that.

kaspersorensen:
Fixes #514


kaspersorensen:
Anyone wanna review, maybe @LosD? :-)

LosD:
Errr... Yeah. I'm not entirely sure why that has gone under the radar.

Anyways, LGTM!

PR #530, opened 2015-07-24
Protect delegates of CompositeAnalysisListener from each others exceptions

### LINKED ISSUE #504: CompositeAnalysisListener: Errors in "Delegate" listeners will block subsequent listeners from being invoked
We currently don't catch any exceptions from delegate listeners of CompositeAnalysisListener. This means that if there are more than one delegates and any of them breaks, the remaining ones will never get notified.

LosD:
Wraps every call to a delegate in a try...catch block to assure that following listeners will also be called.

Fixes #504

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /.classpath
 /target
 /.DS_Store
+.idea
+*.iml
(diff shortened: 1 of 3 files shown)


kaspersorensen:
LGTM, except maybe we don't want to format the license header.

PR #532, opened 2015-07-24
Only do a partial validation when updating datastore name.

### LINKED ISSUE #518: "Datastore name" field in CSV datastore dialog (and possibly others) cause refresh of preview data
In the CsvDatastoreDialog when I create a new datastore and change the name of the datastore, then everything is refreshing. This is especially a problem in the CSV case because there is a preview table below which is re-fetched every time.

The datastore name is just a personalized identifier. It does not affect the loading of the datastore in any way, so this specific field should not influence the preview table at all.

LosD:
Instead of a doing full validation on datastore name change, only validate the name itself and enable the save button.

Also added a bit of code cleanup

Fixes #518

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/AbstractDatastoreDialog.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/AbstractDatastoreDialog.java
@@ -120,16 +120,20 @@ public abstract class AbstractDatastoreDialog<D extends Datastore> extends Abstr
         _datastoreNameTextField.getDocument().addDocumentListener(new DCDocumentListener() {
             @Override
             protected void onChange(DocumentEvent event) {
-                validateAndUpdate();
+                validateAndUpdateInternal();
             }
         });
     }
-    
+
     protected void validateAndUpdate() {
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/java/org/datacleaner/windows/AbstractDatastoreDialog.java
```diff
@@ -120,16 +120,20 @@ public void actionPerformed(ActionEvent e) {
...
             }
         });
     }
-    
+
     protected void validateAndUpdate() {
+        validateAndUpdateInternal();
+    }
+
+    private void validateAndUpdateInternal() {
```
kaspersorensen: Does this change anything? Wouldn't subclasses just override validateForm()? Or maybe that should only be really for the form inputs individually. Maybe we should document what subclasses should/could override to make it clearer.

LosD: I don't think so... Validation is just validation, it's the "AndUpdate" part of validateAndUpdate that is the problem :)

PR #533, opened 2015-07-28
Cache and store fo Componets

KarelHouzvicka:


jakubneubauer:
Karel's unsuccessfull attempt to do a pull request :-)

PR #539, opened 2015-08-05
Fix for bug#538

ankit2711:
Refactored the CaptureChangedRecordsFilter to name the state field as XXX.GreatestLastModifiedValue instead of XXX.GreatestLastModifiedTimestamp. The filter also works with ID fields which contain numbers.


kaspersorensen:
LGTM

PR #545, opened 2015-08-07
Feature/543 refactor registering new datastores

### LINKED ISSUE #543: Refactor registering new datastores.
After introducing #508, the code for registering new datastore types is duplicated in 3 places: DatastoreManagementPanel, SingleDatastorePropertyWidget and the place under the file dropzone has its own code. It includes defining what datastores are available in DC, so when we would like to add support for a new DB, all these places needs to be reviewed.

We should have a single class exposing descriptors of the datastore types available in current DC instance (the instances can vary wrt the DB drivers installed). The descriptor should expose information such as datastore name, datastore class, datastore dialog class, icon, maybe some groups the datastore belong to (e.g. file-based, NoSQL) for easier presentation in the UI.

tomaszguzialek:
Fixes #543


desktop/ui/src/main/java/org/datacleaner/panels/AddDatastorePanel.java
```diff
@@ -240,17 +212,17 @@ public void windowClosed(WindowEvent e) {
...
-                    }
-                });
+                // dialog.addWindowListener(new WindowAdapter() {
+                // @Override
+                // public void windowClosed(WindowEvent e) {
+                // if (dialog.getSavedDatastore() != null) {
+                // _datastoreSelectedListener.datastoreSelected(dialog.getSavedDatastore());
+                // }
+                // }
+                // });
```
tomaszguzialek: Oh, I forgot this. Will double check that it is not necessary anymore and delete.

kaspersorensen:
Could we make this in a bit more unintuitive way for the core/engine? I wasn't expecting to influence the DatastoreCatalog class for instance. And it slightly skews the responsibility of that class into a new direction where it is not a user-specific catalog but a metadata store about any datastore. For now I would just put it in a desktop-only class similar to DatabaseDriverCatalog.

tomaszguzialek:
During testing, I realized some dialogs got broken during refactoring. Do not merge yet, need to fix it.

tomaszguzialek:
Fixed the issues, PR ready to be reviewed again.

api/src/main/java/org/datacleaner/connection/DatastoreDescriptor.java
```diff
@@ -0,0 +1,40 @@
...
+    
+    public String getName();
+    
+    public String getDescription();
+    
+    public Class<? extends Datastore> getDatastoreClass();
+    
+    public boolean isUpdatable();
+    
+    public List<String> getTags();
```
kaspersorensen: From what I can see this is used to depict just if the datastore is put under the "Cloud service" or "Database" button in the dropzone screen. But a concept like tags is very generic and hard to interpret from reading this interface. And since it's entirely a UI representation thing, maybe having this categorization just in the UI would be better.

desktop/ui/src/main/java/org/datacleaner/connection/DatastoreDescriptorDesktopBindings.java
```diff
@@ -0,0 +1,186 @@
...
+import org.datacleaner.windows.HBaseDatastoreDialog;
+import org.datacleaner.windows.JdbcDatastoreDialog;
+import org.datacleaner.windows.JsonDatastoreDialog;
+import org.datacleaner.windows.MongoDbDatastoreDialog;
+import org.datacleaner.windows.SalesforceDatastoreDialog;
+import org.datacleaner.windows.SasDatastoreDialog;
+import org.datacleaner.windows.SugarCrmDatastoreDialog;
+import org.datacleaner.windows.XmlDatastoreDialog;
+
+public class DatastoreDescriptorDesktopBindings {
```
kaspersorensen: I like this class :-) And I think the DatastoreDescriptor should be closer tied to it - it's a (desktop) UI thing only right now, so I would argue that they should both be in the desktop module. Furthermore I would find it pretty natural that this class had a couple of methods to help layout like "getDatabaseDatastoreDescriptors", "getCloudServiceDatastoreDescriptors" and "getFileDatastoreDescriptor". That would eliminate the need for the tags system.

desktop/ui/src/main/java/org/datacleaner/connection/DatastoreDescriptorDesktopBindings.java
```diff
@@ -0,0 +1,186 @@
...
+    
+    private static final DatastoreDescriptor COMPOSITE_DATASTORE_DESCRIPTOR = new DatastoreDescriptorImpl("Composite datastore",
+            "Create a composite datastore", CompositeDatastore.class);
+
+    private static Map<DatastoreDescriptor, String> _iconPaths = new HashMap<DatastoreDescriptor, String>();
+
+    private static Map<DatastoreDescriptor, Class<? extends AbstractDatastoreDialog<? extends Datastore>>> _dialogClasses = new HashMap<>();
+
+    static {
+        _iconPaths.put(CSV_DATASTORE_DESCRIPTOR, IconUtils.CSV_IMAGEPATH);
```
kaspersorensen: If you make DatastoreDescriptor just a desktop-ui-thing, then you can add the icon paths directly on the DatastoreDescriptor class.

desktop/ui/src/main/java/org/datacleaner/connection/DatastoreDescriptorDesktopBindings.java
```diff
@@ -0,0 +1,186 @@
...
+        _iconPaths.put(HIVE_DATASTORE_DESCRIPTOR, "images/datastore-types/databases/hive.png");
+        _iconPaths.put(MYSQL_DATASTORE_DESCRIPTOR, "images/datastore-types/databases/mysql.png");
+        _iconPaths.put(POSTGRESQL_DATASTORE_DESCRIPTOR, "images/datastore-types/databases/postgresql.png");
+        _iconPaths.put(ORACLE_DATASTORE_DESCRIPTOR, "images/datastore-types/databases/oracle.png");
+        _iconPaths.put(SQLSERVER_DATASTORE_DESCRIPTOR, "images/datastore-types/databases/microsoft.png");
+        _iconPaths.put(COMPOSITE_DATASTORE_DESCRIPTOR, IconUtils.COMPOSITE_IMAGEPATH);
+    }
+
+    static {
+        _dialogClasses.put(CSV_DATASTORE_DESCRIPTOR, CsvDatastoreDialog.class);
```
kaspersorensen: ... and then you can also add the dialog class direcly on the datastore descriptor instead of having this additional map.

engine/core/src/main/java/org/datacleaner/database/DatabaseDriverCatalog.java
```diff
@@ -263,4 +286,164 @@ public boolean isInstalled(String databaseName) {
         }
         return getState(databaseDriver) == DatabaseDriverState.INSTALLED_WORKING;
     }
+    
+    /**
+     * Returns the descriptors of datastore types available in DataCleaner.
+     */
+    public List<DatastoreDescriptor> getAvailableDatastoreDescriptors() {
```
kaspersorensen: The changes/additions here in this class seems redundant with the complete DatastoreDescriptorDesktopBindings class?

tomaszguzialek:
Refactored again. Datastores are available as static fields in DatastoreDescriptors class. This class also has non-static methods determining file-based and database-based datastores that are available in current DC instance (drivers are present).

kaspersorensen:
LGTM. A lot of lines of code. Not sure I can see through all the details of it, but the design is a lot better now IMO.

kaspersorensen:
One sanity test we should probably do:

Start up this new version alongside a previous version. Make sure that on all the screens the same datastore options pop up.

PR #548, opened 2015-08-09
Feature/537 docs improvements

### LINKED ISSUE #537: Documentation: add <multiple-connections>
In the jdbc datastore examples in the doc section about conf.xml we don't have any examples that explicitly define <multiple-connections>. 

Just had to explain that and found it missing :-)

kaspersorensen:
Fixes #537 and adds a few other minor documentation improvements as well.

Regarding the datastores (many red and green lines in the diff viewer) - no content has changed, except the addition of the <multiple-connections> tags. The diffs are there because of an indentation change (moved JDBC datastore, Excel datastore etc. to a subsection of "Datastores").

### DIFF
--- a/documentation/src/docbkx/chapter41-configuration-file.xml
+++ b/documentation/src/docbkx/chapter41-configuration-file.xml
@@ -372,9 +372,62 @@
 
 	<section id="configuration_file_reference_data">
 		<title>Reference data</title>
-		<para>Reference data are defined in the configuration file in the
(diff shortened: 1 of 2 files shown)


kaspersorensen:
Hmm the travis failure seems unrelated to this change ... but pretty weird that a unittest would fail suddenly. Must be an hour-vs-day difference or so (build was close to midnight here).

PR #550, opened 2015-08-10
Use OutputDataStream table for OutputRowDataStreamRowCollector.

### LINKED ISSUE #549: OutputRowCollector based on original column
The OutputRowCollector is based on the column headers in the original file, instead of the ones that the component actually want to output. This only becomes a problem if the column count differs between them, as DefaultRow only validates that. It has fallen through the cracks as both tests and currently implemented (proprietary) components has been using the same column count as the input data.

Instead, base the OutputRowCollector on the Table in the OutputStream.

LosD:
Fixes #549


LosD:
Since this is an issue with a bit of assumption on the idea behind the Query object, and @kaspersorensen was the original author, I'm assigning Kasper as the reviewer.

kaspersorensen:
It's OK with me. It's a bit distant from memory but it seems fine and anyways in the first pass we just want something that works for basic scenarios.

PR #553, opened 2015-08-11
Publish character set distribution result as output datastream

### LINKED ISSUE #549: OutputRowCollector based on original column

LosD:
Fixes #549


components/i18n/src/main/java/org/datacleaner/beans/CharacterSetDistributionAnalyzer.java
```diff
@@ -162,6 +171,40 @@ public CharacterSetDistributionResult getResult() {
...
+            }
+        }
+
         return new CharacterSetDistributionResult(_columns, unicodeSetNames, crosstab);
     }
+
+
+    @Override
+    public OutputDataStream[] getOutputDataStreams() {
+        final OutputDataStreamBuilder streamBuilder = OutputDataStreams.pushDataStream("character sets");
```
tomaszguzialek: Let's use a constant.

components/i18n/src/test/java/org/datacleaner/beans/CharacterSetDistributionAnalyzerTest.java
```diff
@@ -130,4 +150,76 @@ public void testSimpleScenario() throws Exception {
         assertEquals("Thaana                0      0 ", resultLines[23]);
         assertEquals("Thai                  0      0 ", resultLines[24]);
     }
+
+    public void testOutputDataStream() throws Throwable {
+        final MultiThreadedTaskRunner taskRunner = new MultiThreadedTaskRunner(16);
+        final DataCleanerEnvironment environment = new DataCleanerEnvironmentImpl()
+                .withTaskRunner(taskRunner);
+        final FileResource file = new FileResource("src/test/resources/strings.txt");
```
tomaszguzialek: I would prefer to have a more informative name for the file. Maybe testOutputDataStream.txt?

tomaszguzialek:
LGTM.

kaspersorensen:
Ehm but isn't the feature here quite strange? I understand we needed an example for testing but this isn't something I would thing anyone wants to have, so it just adds complexity and slightly nonsensical feeling to the user.

LosD:
I somehow agree, although it is possible you'd want to analyze the number distribution of... Errrr, your character sets. 

Patten analyzer and value distribution would probably make more sense, but in the current design, those are not possible because of the multiple analyzers for multiple columns, as the output streams does not get initialized in that case, and we're not entirely sure if that is by design. 

Anyway, we needed _something_ and this one was easy, and is easy to rip out again when something better comes along. And it already served its purpose by uncovering a bug :)

kaspersorensen:
Glad about the bug of course :-) 

The problem is that the moment someone starts using it then we need to have continued support for it and so on. Just removing it would render jobs that use it invalid. I would propose instead two things:

An output stream om Completeness Analyzer called "Incomplete records". And maybe another called "Complete records". At least I have had a few users wish for that, so it's something better worth doing and supporting.

Create another mock implementation in testware and include testware as an extension if you need to test.

LosD:
Perfect idea! I'll get right on that

PR #554, opened 2015-08-11
Feature/516 outputdatastreams xml support

### LINKED ISSUE #516: OutputDataStreams: Add Job XML support
Related to new data streams in jobs - #224

This part of the bigger story is about making sure we can save and load analysis XML job files which contain melon jobs.

Technical note to developers: Both these should to be properly test-driven endevours, the story is very well fit for it.

 - [ ] Add functionality to JaxbJobWriter
 - [ ] Add functionality to JaxbJobReader

### LINKED ISSUE #551: Add OutputDataStream suppport to analyzer
Currently, there is no analyzer actually publishing OutputDataStream(s), so doing real-world tests of #516 and (especially) #519 is a bit problematic.

Identify a good candidate and implement HasOutDatastream on it.

LosD:
Implements reading and writing of OutDataStreams from/to analysis XML jobs.

A small change needed to be made to the job XSD, as the JobType inside OutputDataStreams has no data context. We tried using restrictions and extensions on the original job type, but restrictions can only be done if the removed element is optional. We might want to rethink how to best do this.

Fixes #516 
Fixes #551


kaspersorensen:
As mentioned in #553 I think it's unfortunate to add the output stream to Charset Distribution. That shouldn't influence this PR of course, but unfortunately this PR does it's testing using that analyzer. So maybe change the test of this PR so that it does not require the unfortunate change to Charset Distribution.

engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
```diff
@@ -446,29 +448,66 @@ private AnalysisJobBuilder create(JobType job, SourceColumnMapping sourceColumnM
...
+        wireOutputDataStreams(componentBuilders);
+    }
+
+    private void wireOutputDataStreams(Map<ComponentType, ComponentBuilder> componentBuilders) {
+        for (Map.Entry<ComponentType, ComponentBuilder> entry : componentBuilders.entrySet()) {
+            ComponentType componentType = entry.getKey();
+            ComponentBuilder componentBuilder = entry.getValue();
+            for (OutputDataStreamType outputDataStreamType : componentType.getOutputDataStream()) {
+                String name = outputDataStreamType.getName();
+                final OutputDataStream outputDataStream = componentBuilder.getOutputDataStream(name);
```
kaspersorensen: Kinda mixed use of final and not final in this method.

kaspersorensen:
Other than that the PR looks pretty good to me :)

tomaszguzialek:
Switched to Completness Analyzer, looking good, feeling good, merging.

PR #556, opened 2015-08-12
Avoid creating a derby.log file

### LINKED ISSUE #528: Get rid of derby.log file from DC application directory
Since we've introduced the Apache Hive support I am now noticing an unfortunate side-effect: Every time I run DataCleaner I see that a file called "derby.log" is located in my application directory. The file is empty, yet is an annoyance and I am not sure if it could even be a problem on installations where the user does not have write-access to the application directory.

I did a bit of research and it seems that Apache Hive uses Derby for the so-called "metastore", a metadata database. We should see what we can do to avoid this log file in DC just because of that.

LosD:
This forces Derby to use a dummy OutputStream instead of dumping logs to a file.

Fixes #528


kaspersorensen:
Very good solution IMO :-)

PR #557, opened 2015-08-12
Replaced proprietary "javax.annotation" package from geronimo

kaspersorensen:
Replaced proprietary "javax.annotation" package from geronimo with official one.

This doesn't matter much, but it's better to use the standard artifact for the javax stuff.


LosD:
LGTM. Let's merge if Travis is happy.

Issue #519, opened 2015-07-21
OutputDataStreams UI graph support


kaspersorensen:
Related to output data streams - #224

This story is about adding desktop UI support for such jobs. Things to cover (break down if necesary):

 - [x] Drawing of the OutputDataStreamJobs on the canvas (assuming a job is opened which already has one or more "melons").
 - [x] Right clicking on a component that has output data streams should yield a new option: "Link [stream name] to ..."
 - [x] Linking a new component to a OutputDataStreamJob. This involves "moving" it from the parent job to a child job builder.

kaspersorensen:
Trying to flesh out a bit about how this could/should work:

When a user right-clicks a component which has any OutputDataStreams, the context menu shown should feature a "Link to" menu item for each output data stream available. The label of them should be "Link (name of stream) to ..."

When then dropping such a link onto a component, that component should be moved from it's current AnalysisJobBuilder to the AnalysisJobBuilder that is associated with the stream.

PR #558, opened 2015-08-13
Feature/519 render melon jobs

LosD:
This implements the actual rendering of the melon jobs.

Fixes the rendering part of #519 (Do not close until other parts are complete).

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphNodeBuilder.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphNodeBuilder.java
@@ -68,18 +68,17 @@ class JobGraphNodeBuilder {
     public DirectedGraph<Object, JobGraphLink> buildGraph() {
 
         final DirectedGraph<Object, JobGraphLink> graph = new DirectedSparseGraph<Object, JobGraphLink>();
+        final List<Table> sourceTables = _analysisJobBuilder.getSourceTables();
 
-        buildGraphInternal(graph, _analysisJobBuilder);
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphNodeBuilder.java
```diff
@@ -296,6 +306,19 @@ private void addNodes(DirectedGraph<Object, JobGraphLink> graph, SourceColumnFin
                     }
                 }
             }
+
+            if (item instanceof ComponentBuilder) {
+                ComponentBuilder componentBuilder = (ComponentBuilder) item;
+
+                for(OutputDataStream outputDataStream : componentBuilder.getOutputDataStreams()){
+                    if(componentBuilder.isOutputDataStreamConsumed(outputDataStream)) {
```
kaspersorensen: Minor: Space missing after "if" :)

kaspersorensen:
Looks good to me. Can you supply also a screenshot? The way I read it, it sounds like each stream also gets a table node, is that correct? I had imagined it without a table node, but with just an edge from the producing component to the first component(s) of the "substream job". But it's one of those things that maybe it makes just as much sense when you see it in a different way.

LosD:
It's with table nodes. When there's multiple streams it pretty hard to understand what is happening otherwise.

Unfortunately, multiple streams also made it obvious that we have some kind of engine problem. I'm not sure what's going on, but something explodes completely, and some analyzers gets run without being initialized, leading to NPEs. I'm not entirely sure why that doesn't happen with the MockAnalyzer. We'll probably have to do some pairing on that Monday. I spent almost all Friday trying to figure out what in the world is going on.

PR #562, opened 2015-08-18
Implements linking of outputdatastreams in the ui

### LINKED ISSUE #519: OutputDataStreams UI graph support

LosD:
This makes it possible to link a component that has an outputdatastream to other components. This is in no way a complete implementation of ways to link (i.e. it is not possible to link a component using the column list).

Fixes #519

### DIFF
--- a/api/src/main/java/org/datacleaner/job/output/OutputDataStreamBuilderImpl.java
+++ b/api/src/main/java/org/datacleaner/job/output/OutputDataStreamBuilderImpl.java
@@ -22,7 +22,9 @@ package org.datacleaner.job.output;
 import org.apache.metamodel.schema.Column;
 import org.apache.metamodel.schema.ColumnType;
 import org.apache.metamodel.schema.MutableColumn;
+import org.apache.metamodel.schema.MutableSchema;
 import org.apache.metamodel.schema.MutableTable;
+import org.apache.metamodel.schema.Schema;
 import org.apache.metamodel.schema.Table;
 import org.datacleaner.api.OutputDataStream;
 
@@ -33,7 +35,10 @@ final class OutputDataStreamBuilderImpl implements OutputDataStreamBuilder {
 
     public OutputDataStreamBuilderImpl(String name) {
         _name = name;
-        _table = new MutableTable(name);
+        MutableSchema schema = new MutableSchema();
+        schema.setName(null);
+        _table = new MutableTable(name, schema);
+        schema.addTable(_table);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLinkPainter.java
```diff
@@ -201,7 +229,18 @@ private boolean createLink(final Object fromVertex, final Object toVertex, final
...
+
                 try {
+                    if (sourceAnalysisJobBuilder != componentBuilder.getAnalysisJobBuilder()){
+                        if(componentBuilder.getInput().length > 0) {
+                            logger.debug("createLink(...) returning false - cannot change analysisJobBuilder for a component with configured inputColumns");
+                            return false;
+                        }
+                        componentBuilder.getAnalysisJobBuilder().removeComponent(componentBuilder);
+                        sourceAnalysisJobBuilder.addComponent(componentBuilder);
+                        componentBuilder.setAnalysisJobBuilder(sourceAnalysisJobBuilder);
```
kaspersorensen: Maybe avoid this "remove + add + set" sequence have a single method (in either AnalysisJobBuilder or ComponentBuilder) that "moves" the component to a different scope?

kaspersorensen:
Mostly looks good to me, so I am ok with merging. My only hesitation is that this is the first PR that really "exposes" the output data streams to the end users and since other features are still dangling it also exposes bugs that was previosly harmless.

Maybe we should at least first create 4.0.x branch from the master _before_ merging this. That way we can still maintain 4.0.x releases while building version 4.1 with output data streams.

LosD:
Is there any reason to do that before problems arrive? The only change is that we'll be porting all non-melon changes back.

We can always make the branch later from the last non-melon commit, and then cherry pick the changes we want.

LosD:
Fixed comment. Good catch, no need to make the JobGraphLinkPainter have even more non-link-painting to do :)

api/src/main/java/org/datacleaner/job/output/OutputDataStreamBuilderImpl.java
```diff
@@ -33,7 +35,10 @@
 
     public OutputDataStreamBuilderImpl(String name) {
         _name = name;
-        _table = new MutableTable(name);
+        MutableSchema schema = new MutableSchema();
+        schema.setName(null);
+        _table = new MutableTable(name, schema);
+        schema.addTable(_table);
```
LosD: Damn, this actually belonged in the now-merged fix for #563

LosD:
Merging after oral agreement.

PR #564, opened 2015-08-18
Initialize consumers in output datastream as part of parent consumer

### LINKED ISSUE #563: OutputDataStreams does not always get initialized before running
OutputDataStreams will not always be initialized by the RowProcessingPublisher. This can happen because there is a race condition between IntilazationTask and RunRowProcessingPublisherTask; Sometimes all initialization tasks will finish before the row processing starts, and sometimes they will not.

LosD:
This should avoid races between initializers and row processing.

This also extends testing of the reader to ensure that a built job is also runnable. This tests more than the reader, so it might not be the optimal place, but I'm not really sure where it would be, as it is more of an "internal" integration test, testing that the parts assembling and running a job work properly together.

I'll also be lazy and put a small fix of CompletenessAnalyzer on here: There was absolutely no need to create a dataset header, and it actually introduced a bug.

Fixes #563


kaspersorensen:
LGTM!

PR #572, opened 2015-08-20
Use "Register datastore" phrase

### LINKED ISSUE #568: Use phrase "Register" datastore instead of "Save" datastore in datastore dialogs
Self-explanatory I think :-) Check all the datastore dialog buttons and use the word "Register" instead of "Save".

kaspersorensen:
Oh man, the complexity of this patch ;-)

Fixes #568


tomaszguzialek:
LGTM :).

PR #573, opened 2015-08-21
Output data streams as edges

### LINKED ISSUE #567: Represent output data streams on graph as edges (with labels) instead of vertices
With our current support for output data streams, the stream is represented using a table node/vertex. But I think it would be rather better to have them represented just as an edge with a special label and icon - similar to how the filter conditions are represented.

Edge label should be the name of the stream.
Edge icon should (for now) just be the table icon. Or come up with a better proposal if possible ;-)

kaspersorensen:
Fixes #567

Screenshot:

image


desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLink.java
```diff
@@ -63,45 +73,29 @@ public String toString() {
...
-        result = prime * result + ((_requirement == null) ? 0 : _requirement.hashCode());
-        result = prime * result + ((_to == null) ? 0 : _to.hashCode());
-        return result;
+        final HashCodeBuilder hashCodeBuilder = new HashCodeBuilder();
+        hashCodeBuilder.append(_filterOutcome);
+        hashCodeBuilder.append(_from);
+        hashCodeBuilder.append(_requirement);
+        hashCodeBuilder.append(_to);
+        hashCodeBuilder.append(_outputDataStream);
+        return hashCodeBuilder.toHashCode();
```
LosD: Why not the stock Objects.hash(...)?

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphNodeBuilder.java
```diff
@@ -302,7 +286,7 @@ private void addNodes(DirectedGraph<Object, JobGraphLink> graph, SourceColumnFin
                     final Table table = inputColumn.getPhysicalColumn().getTable();
                     if (table != null) {
                         addNodes(graph, scf, table, recurseCount);
-                        addEdge(graph, table, item, null, null);
+                        addEdge(graph, table, item, null, null, null);
```
LosD: All these calls adding a null are a little nasty. Wouldn't an overloaded addEdge() be better?

LosD:
A few comments, but looks good

PR #577, opened 2015-08-21
Feature/components rest api

jakubneubauer:
this PR contains working components API and also integrated REST documentation framework - Swagger, WADL, Miredot.


kaspersorensen:
Minor remark: We tend to prefer not to have the ```@author``` tag in javadoc since it can give a feeling of a class "belonging" to a specific person. It is still here and there in the codebase but I always remove it when I can.

monitor/ui/src/main/webapp/doc/css/screen.css
```diff
@@ -0,0 +1,1279 @@
+/* Original style from softwaremaniacs.org (c) Ivan Sagalaev <Maniac@SoftwareManiacs.Org> */
```
LosD: This copyright header worries me. What is the license for this?

monitor/api/src/main/java/org/datacleaner/monitor/configuration/ComponentsStore.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.datacleaner.monitor.configuration;
+
+/**
+ * Class ComponentsStore
+ *
+ * @author k.houzvicka
+ * @since 11.8.15
+ */
+public interface ComponentsStore {
```
kaspersorensen: Naming-wise I would normally not use the 's' in Components. ComponentStore has same meaning and I believe it's more standard naming.

monitor/api/src/main/java/org/datacleaner/monitor/configuration/TenantContext.java
```diff
@@ -155,4 +155,12 @@
...
      */
     public void onConfigurationChanged();
+
+
+    /**
+     * Return the storage for components.
+     *
+     * @return
+     */
+    public ComponentsStore getComponentsStore();
```
kaspersorensen: Same naming remark here of course.

LosD:
All the CSSs and HTML looks a bit like something that should be autogenerated. Does it really need to be stored in the codebase?

monitor/services/pom.xml
```diff
@@ -200,5 +200,25 @@
 			<version>2.2.4</version>
 			<scope>provided</scope>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.4</version>
```
kaspersorensen: Let's have this in the parent POM under <dependencyManagement> ... I think we would like to move to commons-lang3 in general so having it there makes it easier to manage and avoid version conflicts.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsCache.java
```diff
@@ -0,0 +1,109 @@
...
+ * Class ComponentsCache
+ * For caching and storing Components and configuration
+ *
+ * @author k.houzvicka
+ * @since 24.7.15
+ */
+public class ComponentsCache {
+    private static final Logger logger = LoggerFactory.getLogger(ComponentsCache.class);
+
+    ConcurrentHashMap<String, ComponentsCacheConfigWrapper> data = new ConcurrentHashMap<>();
```
kaspersorensen: Maybe (can be future improvement) we should rather use a Guava cache here with a timeout. I wouldn't want it to be a short timeout or anything like that, but maybe after a day or two we should clean up just to prevent a long-running server instance from growing out of memory on ancient components.

Alternatively it could be a configuration item when creating the component. I guess it deserves it's own issue.

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
```diff
@@ -0,0 +1,251 @@
...
+                break;
+            }
+            Column col = table.getColumn(i);
+            values.add(convertTableValue(col.getType().getJavaEquivalentClass(), value));
+        }
+        return values.toArray(new Object[values.size()]);
+    }
+
+    private Object convertPropertyValue(ConfiguredPropertyDescriptor propDesc, JsonNode userPropValue) {
+        // TODO: other properties type. Find some generic way to convert JsonNode to the specific property type
```
kaspersorensen: Generic way would be to use StringConverter from DC-engine-core. This is the same class that is used when reading analysis.xml files and I think using the same syntax is much preferable .... Even also for InputColumns (which don't get into this method currently).

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
```diff
@@ -0,0 +1,251 @@
...
+                if(propDesc.isInputColumn()) {
+                    List<String> colNames = convertToStringArray(userPropValue);
+                    List<InputColumn> inputCols = new ArrayList<>();
+                    for(String columnName: colNames) {
+                        inputCols.add(getOrCreateInputColumn(columnName));
+                    }
+                    propDesc.setValue(component, inputCols.toArray(new InputColumn[inputCols.size()]));
+                } else {
+                    Object value = convertPropertyValue(propDesc, userPropValue);
+                    propDesc.setValue(component, value);
```
kaspersorensen: Here I would prefer to create a ComponentConfiguration object and set the values on that. Then below you would call LifeCycleHandler.assignConfiguredProperties(...)

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentsController.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Controller for DataCleaner components (transformers and analyzers). It enables to use a particular component
+ * and provide the input data separately without any need of the whole job or datastore dcConfiguration.
+ * @author j.horcicka (GMC)
+ * @since 24. 07. 2015
+ */
+@RequestMapping("/{tenant}/components")
+public interface ComponentsController {
+    public static final String MIME_TYPE_JSON = "application/json";
```
kaspersorensen: Minor: Spring has a MediaType or MediaTypes class that can be used for this

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentsControllerV1.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Controller for DataCleaner components (transformers and analyzers). It enables to use a particular component
+ * and provide the input data separately without any need of the whole job or datastore configuration.
+ * @author k.houzvicka, j.horcicka
+ * @since 8. 7. 2015
+ */
+@Controller
+public class ComponentsControllerV1 implements ComponentsController {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ComponentsControllerV1.class);
+    private static final String TENANT = "tenant";
```
kaspersorensen: This confused me, thinking it was a hardcoded tenant-name. But I see it's the parameter name (so no problem as such). But I would rename it to PARAMETER_NAME_TENANT then.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/RepositoryZipController.java
```diff
@@ -177,12 +181,22 @@ public void run(InputStream fileInput) throws Exception {
                 }
             });
             zipOutput.closeEntry();
+            itemsCount++;
         }
 
         final List<RepositoryFolder> folders = folder.getFolders();
         for (RepositoryFolder subFolder : folders) {
             String name = subFolder.getName();
+            System.out.println("Directory: " + path + name + "/");
```
kaspersorensen: Please use logger instead of System.out
(and please do a search in the complete PR for this)

kaspersorensen:
The /doc location for swagger (I think?) is maybe a bit too generic since it is "only" REST API documentation. Could we put it under /swagger instead?

kaspersorensen:
And if possible, could we have the complete swagger addition in a separate PR, because it's a really confusing thing to combine your code with someone else's code.

monitor/ui/src/main/webapp/doc/css/screen.css
jhorcicka: When you get complete "swagger-ui-master" there is a file LICENSE with the link to: http://www.apache.org/licenses/LICENSE-2.0

components/basic-transformers/src/main/java/org/datacleaner/beans/codec/XmlEncoderTransformer.java
```diff
@@ -19,24 +19,18 @@
...
-import org.datacleaner.api.Transformer;
+import com.google.common.escape.Escaper;
+import com.google.common.xml.XmlEscapers;
+import org.datacleaner.api.*;
 import org.datacleaner.components.categories.StringManipulationCategory;
 import org.datacleaner.data.MockInputColumn;
 
-import com.google.common.escape.Escaper;
-import com.google.common.xml.XmlEscapers;
+import javax.inject.Named;
```
LosD: There seem to be some code style differences here. I guess it is the difference between Eclipse and IntelliJ: Eclipse puts _all_ top-level packages in its own group, IntelliJ only puts a few (java and javax comes to mind).

I had to add top-level packages found ( .com, .org, etc) to my IntelliJ code style configuration, which is a bit of a pain, especially since it is an ongoing process.

image

components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateTimestampTransformer.java
```diff
@@ -19,21 +19,16 @@
...
-
 import org.apache.metamodel.util.HasName;
-import org.datacleaner.api.Categorized;
-import org.datacleaner.api.Configured;
-import org.datacleaner.api.Description;
-import org.datacleaner.api.InputColumn;
-import org.datacleaner.api.InputRow;
-import org.datacleaner.api.OutputColumns;
-import org.datacleaner.api.Transformer;
+import org.datacleaner.api.*;
```
LosD: Same with the *-ing of a package with many imports. I change my IntelliJ's limit to 30 to not clash with the rest of our team's codestyle.

api/src/main/java/org/datacleaner/api/WSStatelessComponent.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Components with the {@link org.datacleaner.api.WSStatelessComponent} annotation are provided
+ * in a web service for components.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+public @interface WSStatelessComponent {
```
kaspersorensen: Shouldn't this just be the default always? Otherwise we are effectively asking everybody to add this annotation, instead that we just make it always the case.

api/src/main/java/org/datacleaner/api/WSPrivateProperty.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Fields with the {@link WSPrivateProperty} annotation are not provided in a web service for components.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD })
+public @interface WSPrivateProperty {
```
kaspersorensen: I would prefer to (re)use the existing ```@HiddenProperty``` instead of introducing a new annotation for this.

monitor/api/src/main/java/org/datacleaner/monitor/configuration/ComponentsStoreHolder.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Class ComponentsStoreHolder
+ * Object for storing to component cache.
+ * 
+ * @since 11.8.15
+ */
+public class ComponentsStoreHolder implements Serializable {
```
kaspersorensen: Naming-wise having the additional "s" that was removed from ComponentStore.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsCache.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Class ComponentsCache
+ * For caching and storing Components and configuration
+ *
+ * @since 24.7.15
+ */
+public class ComponentsCache {
```
kaspersorensen: Also this one has then the additional "s" which we removed in ComponentStore.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsCacheConfigWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.datacleaner.monitor.server.components.ComponentHandler;
+
+/**
+ * Class ComponentsCacheConfigWrapper Simple wrapper for store to cache with expiration time.
+ * 
+ * @author k.houzvicka
+ * @since 28.7.15
+ */
+public class ComponentsCacheConfigWrapper {
```
kaspersorensen: And this one :)

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.datacleaner.monitor.server.components.ComponentHandler;
+
+/**
+ * Class ComponentsFactory
+ *
+ * @author k.houzvicka
+ * @since 18.8.15
+ */
+public class ComponentsFactory {
```
kaspersorensen: And this one :)
Maybe it should actually be renamed to ComponentHandlerFactory, as this is what it produces.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/TenantContextFactory.java
```diff
@@ -29,4 +31,6 @@
     public TenantContext getContext(TenantIdentifier tenant);
     
     public TenantContext getContext(String tenantId);
+
+    public Set<String> getAllTenantsName();
```
kaspersorensen: Naming-wise: Should be getAllTenantNames() to make it grammatically correct.

kaspersorensen: On the content side we should make this an Iterator<String> because we can potentially be getting a VERY long list of tenants when we deploy this on datacleaner.org (15-20,000 registered users)

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsCache.java
```diff
@@ -0,0 +1,233 @@
...
+
+        @Override
+        public void run() {
+            while (running) {
+                if (firstRun) {
+                    firstRun = false;
+                } else {
+                    Set<String> tenants = _tenantContextFactory.getAllTenantsName();
+                    Set<String> allIdInCache = new HashSet<>(data.keySet());
+                    for (String tenant : tenants) {
```
kaspersorensen: Ouch, this is not a good approach... We should cache this in the TenantContexts that are active only. Here we are iterating and calling getContext(tenant) on _all_ the tenants. That's potentially a very big iteration and calling getContext(tenant) is very expensive to do in cases where there is no need for the context.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsCache.java
```diff
@@ -0,0 +1,233 @@
...
+ * @since 24.7.15
+ */
+public class ComponentsCache {
+    private static final Logger logger = LoggerFactory.getLogger(ComponentsCache.class);
+    private static final long CHECK_INTERVAL = 5 * 60 * 1000;
+    private static final long CLOSE_TIMEOUT = 60 * 1000;
+
+    private TenantContextFactory _tenantContextFactory;
+
+    private ConcurrentHashMap<String, ComponentsCacheConfigWrapper> data = new ConcurrentHashMap<>();
```
kaspersorensen: A waaaay easier and more fail-safe suggestion for caching: Use a Google Guava cache :-)

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsCache.java
jhorcicka: After getting familiar with Guava Cache options and after a team discussion we decided not to use Guava Cache and keep the original implementation of the cache system.

1) Guava Cache does not provide a way how to set various timeout values for various elements in the store. AFAIK there is only one global setting for the whole cache object.
2) Our cache handles two sets of data. The first one is the configuration set in the memory which could have been covered by Guava Cache. The second is the repository set which is stored on the disk. We need the possibility to handle both sets independently. For example remove only one of them for the specified key (instanceId).

api/src/main/java/org/datacleaner/api/WSStatelessComponent.java
jhorcicka: Well, we think that supposing all components are stateless and provided in REST API implicitly is a potential problem. Firstly, some transformers have their state (for example ID Generator). Secondly, not all components are required to be available in the REST API. So if someone marks the component stateless by the annotation it is his responsibility that it is not stateless and accessible. ("Do what is allowed" politics instead of "do not do what is not allowed". )

kaspersorensen: I still feel this is not correct though. I can agree to the fact that some transformers should maybe not be exposed this way. But that's a more general characteristic of some (quite few) transformers and such characteristic should be explicitly annotated, not assumed. In other words: I think the default here is wrong. By default we should expose all components, but we could add an annotation for excluding particular components (separate story, please don't add any more to this PR since it's too big - and I have quite some comments on such an annotation also, since we should use the same statefulness flag in other areas than WS specific).

components/basic-transformers/src/main/java/org/datacleaner/beans/codec/XmlEncoderTransformer.java
kaspersorensen: I second this remark. For me it's not an absolute blocker, but it is quite unfortunate that there are so many files in the diff where the only change is a reorganization of imports.

engine/core/src/test/java/org/datacleaner/util/ReflectionUtilsTest.java
```diff
@@ -154,6 +155,20 @@ public void testGetFields() throws Exception {
         assertEquals("a", fields[1].getName());
     }
 
+    private Field[] getNonSyntheticFields(Class clazz) {
```
kaspersorensen: This change is now "fixing" something in the test. But it seems to me that maybe it is ReflectionUtils itself that needs this fix?

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentCacheMapImpl.java
```diff
@@ -0,0 +1,241 @@
...
+        @Override
+        public void run() {
+            while (running) {
+                if (firstRun) {
+                    firstRun = false;
+                } else {
+                    Iterator<RepositoryFolder> repositoryFolderIterator = _tenantContextFactory.getRepositoryFolderIterator();
+                    Set<String> allIdInCache = new HashSet<>(data.keySet());
+
+                    while (repositoryFolderIterator.hasNext()) {
```
kaspersorensen: This is (still) going to iterate through every possible tenant there is and initialize it using the TenantContextFactory.getContext(name) method. Why is this necesary? Can't we just keep track of which tenants are currently having any components active? That stuff is going to be in the ConcurrentHashMap I think?

components/basic-transformers/src/main/java/org/datacleaner/beans/codec/XmlEncoderTransformer.java
jhorcicka: OK. I am working on it.

jhorcicka: OK. It should be better now.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentCacheMapImpl.java
jhorcicka: The reason for cycling through all the tenants is a potential problem in load balancing. When nodeA goes down after some time of work it has some repositories stored in the shared file system but the content of the cache (in memory) is gone. NodeB then has to take care of the "lost" repositories. 

If this is a really big problem we could think of another solution: let's not use tenants' directories but a special new shared directory for storing component configurations. They have an unique ID. I am unaware of the complexity of this but I can certainly try it. 

Or do you have a better idea?

engine/core/src/test/java/org/datacleaner/util/ReflectionUtilsTest.java
jhorcicka: I moved the method getNonSyntheticFields into ReflectionUtils but I think it is needed only in tests. The reason for this is the new maven profile for code-coverage. The JaCoCo library enhances classes with an additional (synthetic) field.

monitor/api/src/main/java/org/datacleaner/monitor/configuration/ComponentsStoreHolder.java
jhorcicka: I renamed all our Components* classes to Component*.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentsCache.java
jhorcicka: Thanks for a tip. I will try to rewrite this with Guava.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ComponentCacheMapImpl.java
kaspersorensen: At least I am wondering: If we added a method to TenantContextFactory to get just the tenants that was already loaded... Would that be sufficient then?

I mean - if there is a component cached/stored in some other tenant, is there any reason to clean it up yet? If the tenant is not loaded anyway then the component is not taking up any resources on the server (except for a tiny bit of storage on disk, but that's dismissable I think).

jhorcicka: OK. We can try this approach. At least as a first variant to finish the pull request. Later we can decide to find a better solution. (The only problem is that small storage waste. )

monitor/services/src/main/java/org/datacleaner/monitor/configuration/TenantContextFactory.java
```diff
@@ -29,4 +32,8 @@
     public TenantContext getContext(TenantIdentifier tenant);
     
     public TenantContext getContext(String tenantId);
+
+    public Iterator<RepositoryFolder> getRepositoryFolderIterator();
```
kaspersorensen: This method seems to be only used in unittests which tells me we probably should add it to any interface. And adding it kinda establishes something as a rule which is currently only an implementation detail, namely that each tenant is identified by a repository folder.

Maybe you can just add it to the implementation class as a method, but don't expose it in any interface.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/TenantContextFactory.java
```diff
@@ -29,4 +32,8 @@
...
+
+    public Iterator<TenantContext> getActiveTenantContextIterator();
```
kaspersorensen: Could we instead return Iterable here, and not add the return-type class to the method name, so it will just be "getActiveTenantContexts()" ?

monitor/services/src/main/java/org/datacleaner/monitor/configuration/TenantContextFactory.java
jhorcicka: I removed it completely. It was redundant. Thanks for the notice.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/TenantContextFactory.java
jhorcicka: Of course.

kaspersorensen:
OK I think it's finally time to merge this PR :-) Thanks for the patience. When Travis is green then let's do it.

kaspersorensen:
Merged, hurray! :-)

Issue #571, opened 2015-08-20
Naive implementation of DC on Spark


tomaszguzialek:
Issue #534 made a first step towards executing DC jobs on Spark. Now, when we have the input file and the DC job available for Spark, let's try to execute it.

PR #586, opened 2015-08-24
Feature/571 naive dc spark job

tomaszguzialek:
Ref #571 

The specified job is executed up to the transformation part (transformers and filters). Analyzers are not included in this PR.

It is based on the SNAPSHOT version of MetaModel as METAMODEL-179 is affecting reading of jobs from HDFS.


spark/src/main/java/org/datacleaner/spark/CsvParser.java
```diff
@@ -0,0 +1,58 @@
...
+import org.datacleaner.data.MockInputRow;
+
+import au.com.bytecode.opencsv.CSVParser;
+
+public class CsvParser implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    public static InputRow prepareInputRow(Collection<InputColumn<?>> jobColumns, CsvConfiguration csvConfiguration, String csvLine) throws IOException {
+        CSVParser openCsvParser = new CSVParser(csvConfiguration.getSeparatorChar(), csvConfiguration.getQuoteChar());
```
kaspersorensen: here we should also add escape char

spark/src/main/java/org/datacleaner/spark/CsvInputRowMapper.java
```diff
@@ -0,0 +1,61 @@
...
+            throw new IllegalStateException("Datastore referred by the job (" + datastoreName
+                    + ") has not been found in the specified DataCleanerConfiguration");
+        }
+
+        DataContext dataContext = datastore.openConnection().getDataContext();
+        if (dataContext instanceof CsvDataContext) {
+            CsvDataContext csvDataContext = (CsvDataContext) dataContext;
+            _csvConfiguration = csvDataContext.getConfiguration();
+        } else {
+            throw new IllegalArgumentException("Only CSV datastores are supported");
```
kaspersorensen: I guess we should have also a few more checks in place because we actually also only support CSV datastores when:

 * multilineValues = false
 * Encoding is UTF-8 (at least I think Spark requires this when calling .textFile(..)?)

kaspersorensen: and I am not sure what we do wrt. headers, but I think we also require:

 * columnNameLineNumber = 1

spark/src/main/java/org/datacleaner/spark/SparkJobLauncher.java
```diff
@@ -60,18 +66,40 @@ public void launchJob(AnalysisJob analysisJob) {
...
-                    long lineCount = instructionsRDD.count();
-        
-                    logger.info("Line count: " + lineCount);
+                    Accumulator<Integer> dataCleanerConfigurationReadsCounter = sc.accumulator(0,
+                            "DataCleanerConfigurationConfiguration reads counter");
+                    Accumulator<Integer> analysisJobReadsCounter = sc.accumulator(0, "AnalysisJob reads counter");
+
+                    SparkDataCleanerContext sparkDataCleanerContext = new SparkDataCleanerContext(_dataCleanerConfigurationPath, analysisJobXmlPath);
+                    sparkDataCleanerContext.addAccumulator(dataCleanerConfigurationReadsCounter);
+                    sparkDataCleanerContext.addAccumulator(analysisJobReadsCounter);
```
kaspersorensen: Would it not be nicer to just add SparkContext to SparkDataCleanerContext? Then you don't need to create the accumulators up front here, but just create them when/if requested. And then you can also support more than those two specific accumulators.

spark/src/main/java/org/datacleaner/spark/SparkJobLauncher.java
```diff
@@ -60,18 +66,40 @@ public void launchJob(AnalysisJob analysisJob) {
...
+                    
+                    JavaRDD<String> inputRDD = sc.textFile(datastoreFilePath);
+                    JavaRDD<InputRow> inputRowRDD = inputRDD.map(new CsvInputRowMapper(sparkDataCleanerContext));
+                    JavaRDD<InputRow> transformationRDD = inputRowRDD.flatMap(new TransformationAction(sparkDataCleanerContext));
+                    List<InputRow> transformedRows = transformationRDD.collect();
```
kaspersorensen: I guess we should delete this and every remaining line?

spark/src/main/java/org/datacleaner/spark/TransformationAction.java
```diff
@@ -0,0 +1,47 @@
...
+    public TransformationAction(final SparkDataCleanerContext sparkDataCleanerContext) {
+        super(sparkDataCleanerContext);
+    }
+
+    @Override
+    public List<InputRow> call(InputRow inputRow) throws Exception {
+        ConsumeRowHandler.Configuration configuration = new ConsumeRowHandler.Configuration();
+        configuration.includeAnalyzers = false;
+        ConsumeRowHandler consumeRowHandler = new ConsumeRowHandler(getAnalysisJob(), getDataCleanerConfiguration(),
+                configuration);
```
kaspersorensen: Creating the handler is sorta heavy weight and each as instantiation will create new analyzers etc. This should be done only once per partition, so I think we should first call RDD.glom() and then build this handler for each action iterator in the glom's stream.

ClaudiaPHI:
LGTM!

ClaudiaPHI:
@tomaszguzialek  How is your test ?

tomaszguzialek:
Executing jobs from command line does not work anymore, need to use ExampleLaunch class. IMO that's not a blocker for this PR. Let's merge it.

PR #587, opened 2015-08-24
Using mapPartitions() instead of flatMap() to create ConsumeRowHandler

tomaszguzialek:
once per worker and operate on the iterator of rows instead of a single
row at a time.

@kaspersorensen RDD.mapPartitions() is essentially the same as RDD.glom().flatMap(), but it is a preferrable way in some cases.


kaspersorensen:
This looks good (and I will merge) but I still think we should then no longer collect InputRows but go directly to map to AnalyzerResults. That way we don't have to build more lists of InputRow in memory.

PR #588, opened 2015-08-24
Write CSV resource

### LINKED ISSUE #494: "Create CSV file" to use a Resource instead of a java.util.File
Instead of restricting ourselves to writing CSV _files_ we should rather change the "Create CSV file" analyzer to use a org.apache.metamodel.util.Resource instead of a java.util.File. That way we will enable also writing CSV content in other file systems than just local disk.

kaspersorensen:
This fixes #494.


tomaszguzialek:
LGTM.

PR #589, opened 2015-08-24
Various refactorings to DC spark module

kaspersorensen:
Renamed a few classes.
Made launcher (now runner) implement AnalysisRunner.
Used pair-based mapping and reduction to keep component state between
stream steps.
Consolidated file-loading to a single file: SparkJobContext.
Moved main method to separate class to allow composition of job runs.


tomaszguzialek:
LGTM, I will merge it (into our "main" working branch).

PR #590, opened 2015-08-25
Possibility to get the top-level builder and all available builders

### LINKED ISSUE #570: Ability to set the scope (AnalysisJob) of a component in its dialog
If a job has output data streams then it's currently not possible to change the scope of a component except for using the "Link to" option.

In the component configuration dialog we wish to have a way to set the scope. When a scope changes the available input columns will thus change and the user should see this instantly.

LosD:
This adds methods to get the top-level AnalysisJobBuilder as well as all AnalysisJobBuilders.

To make that possible each AnalysisJobBuilder now know its parent, creating a chain of AnalysisJobBuilders.

This is a prerequisite for #570

Change: No longer getting _all_, only getting descendants

Edit: 
Accidentally added the rest of the #570 fix, so if merged, fixes #570.

Please note that there is still a few limitations: Only the default scope will instantly reflect the changes, other scopes will update upon closing the dialog. The scope label does not show which component it comes from, making i.e. several CompletenessAnalyzers will end up with several "Complete rows" scopes and several "Incomplete rows" scopes.

Edit: This should be mergeable, though we have some serious space and overlapping issues now:
image

(Just realized this may be confusing: "Completeness analyzer 1" and "Completeness analyzer 2" are the actual names of the AnalyzerBuilder instances, the menu is not counting them)

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
+++ b/engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
@@ -1220,7 +1220,7 @@ public final class AnalysisJobBuilder implements Closeable {
         return result;
     }
 
-    public AnalysisJobBuilder getTopLevelJobBuilder(){
+    public AnalysisJobBuilder getRootJobBuilder(){
         AnalysisJobBuilder builder = this;
         AnalysisJobBuilder tempBuilder = builder._parentBuilder;
         while(tempBuilder != null){
@@ -1229,27 +1229,5 @@ public final class AnalysisJobBuilder implements Closeable {
         }
 
         return builder;
-    }
-
-    private List<AnalysisJobBuilder> getChildren() {
-        List<AnalysisJobBuilder> childJobBuilders = new ArrayList<>();
-
-        for (ComponentBuilder componentBuilder : getComponentBuilders()) {
-            for (OutputDataStream outputDataStream : componentBuilder.getOutputDataStreams()) {
-                childJobBuilders.add(componentBuilder.getOutputDataStreamJobBuilder(outputDataStream));
-            }
-        }
-
-        return childJobBuilders;
-    }
-
-    public List<AnalysisJobBuilder> getDescendants() {
-        List<AnalysisJobBuilder> descendants = new ArrayList<>();
-        for(AnalysisJobBuilder child : getChildren()){
-            descendants.add(child);
-            descendants.addAll(child.getDescendants());
-        }
-
-        return descendants;
     }
 }
--- a/engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
+++ b/engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
@@ -39,6 +39,7 @@ import org.datacleaner.api.AnalyzerResult;
 import org.datacleaner.api.Filter;
 import org.datacleaner.api.HasAnalyzerResult;
 import org.datacleaner.api.InputColumn;
+import org.datacleaner.api.OutputDataStream;
 import org.datacleaner.api.Transformer;
 import org.datacleaner.api.Validate;
 import org.datacleaner.configuration.DataCleanerConfiguration;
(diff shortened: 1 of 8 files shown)


engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -1000,7 +1014,7 @@ public void setDefaultRequirement(final ComponentRequirement defaultRequirement)
     /**
      * Gets a default requirement, which will be applied to all newly added row
      * processing components.
-     * 
+     *
```
kaspersorensen: In the diff these single-whitespace changes look a bit weird. I can live with it, but if it's a simple thing to fix it would be nice.

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -1205,4 +1219,37 @@ public Boolean eval(InputColumn<?> inputColumn) {
 
         return result;
     }
+
+    public AnalysisJobBuilder getTopLevelJobBuilder(){
```
kaspersorensen: In terms of naming I kinda like the term "ultimate parent" rather than "top level". Maybe just me, but a suggestion at least.

kaspersorensen: And I would almost also expect a normal "get parent" now that we are adding these.

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -1205,4 +1219,37 @@ public Boolean eval(InputColumn<?> inputColumn) {
...
+        for (ComponentBuilder componentBuilder : getComponentBuilders()) {
+            for (OutputDataStream outputDataStream : componentBuilder.getOutputDataStreams()) {
+                childJobBuilders.add(componentBuilder.getOutputDataStreamJobBuilder(outputDataStream));
+            }
+        }
+
+        return childJobBuilders;
+    }
+
+    public List<AnalysisJobBuilder> getDescendants() {
```
kaspersorensen: Maybe we should add "JobBuilder" to the method names here and above. So it becomes getChildrenJobBuilders and getDescendantJobBuilders

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
LosD: They look perfectly reasonable to me, why would there be a space after the *? Do you add spaces to lines between two method scopes?

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
LosD: No fan of "ultimate", if not top level, then maybe root?

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
LosD: I actually loved finding a name with no annoying repeat of the class name... But will add if it causes confusion.

LosD:
Hmmm... Maybe we should rethink this a bit; We now have the possibility to get all the AnalysisJobBuilders inside a job... But they are hard to use for anything, as it is not possible to find out what ComponentBuilder an AnalysisJobBuilder belongs to, without manually traversing all ComponentBuilders and their streams, and checking if it is the right one, essentially re-doing the same work as the AnalysisJobBuilder already did.

I'm thinking of either only keeping the top level/root part here, and let the user do the traversing, maybe with a few helper methods in the AnalysisJobBuilder.

LosD:
Oh... I accidentally pushed the other part of the feature to this branch. Well, that was stupid.

@kaspersorensen do you prefer to keep it as one big PR, or that I split them?

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
kaspersorensen: It's just my IDE that does it I guess :)

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
kaspersorensen: Ok, Root is better then ;)

I don't know, but adding one without the other seems strange :)

kaspersorensen:
I am fine with one big PR

ClaudiaPHI:
LGTM! I will test it.

LosD:
If you have any ideas for the overlapping label and button, it would be grrrrrrreat :)

PR #591, opened 2015-08-25
Feature/571 naive dc spark job analyzers

tomaszguzialek:


spark/src/main/java/org/datacleaner/spark/AnalyzerResultReduceFunction.java
```diff
@@ -0,0 +1,102 @@
...
+        return reducedTuple;
+    }
+
+    private AnalyzerResultReducer<AnalyzerResult> initializeReducer(
+            final Class<? extends AnalyzerResultReducer<?>> resultReducerClass) {
+
+        final DataCleanerConfiguration configuration = _sparkJobContext.getConfiguration();
+        final InjectionManager injectionManager = configuration.getEnvironment().getInjectionManagerFactory()
+                .getInjectionManager(configuration, _sparkJobContext.getAnalysisJob());
+        final LifeCycleHelper lifeCycleHelper = new LifeCycleHelper(injectionManager, true);
```
kaspersorensen: the true should be false - please check the name of the parameter :D

spark/src/main/java/org/datacleaner/spark/AnalyzerResultReduceFunction.java
```diff
@@ -0,0 +1,102 @@
...
+        ResultDescriptor rd = getResultDescriptor(componentJob, analyzerResult1);
+        final Class<? extends AnalyzerResultReducer<?>> resultReducerClass = rd.getResultReducerClass();
+
+        if (resultReducerClass == null) {
+            throw new IllegalStateException("The analyzer (" + analyzerResult1 + ") is not distributable!");
+        }
+
+        AnalyzerResultReducer<AnalyzerResult> reducer = initializeReducer(resultReducerClass);
+        
+        AnalyzerResult reducedAnalyzerResult = reducer.reduce(Arrays.asList(analyzerResult1, analyzerResult2));
```
kaspersorensen: Would be slightly better if we could get a bigger list of results here so that we didn't have to reduce many times. Or at least do bigger chunks than 1+1. Maybe there's a kinda combiner for reducing?

tomaszguzialek: I used reduceByKey() transformation which is an equivalent of M/R reducer with a combiner - it will reduce first all the records on the same node and send only the reduced value. If groupByKey() followed by (flat)map is used then the combiner is not used and the records are shuffled between nodes instead of just sending a reduced value.

spark/src/main/java/org/datacleaner/spark/AnalyzerResultReduceFunction.java
tomaszguzialek: That was a copy and paste, will fix it. However, if there are non-distributable analyzers in the job, it will throw an exception before it even gets to this line.

tomaszguzialek:
Two same analyzers are treated as one and records are shuffled incorrectly. Need to fix it before merging.

tomaszguzialek:
Fixed incorrect shuffling, IMHO mergeable.

spark/src/main/java/org/datacleaner/spark/AnalyzerResultReduceFunction.java
kaspersorensen: OK let's see. My only concern is with overhead of the AnalyzerResultReducer.

kaspersorensen:
LGTM

PR #597, opened 2015-08-27
Long(er) chains of output data streams

### LINKED ISSUE #583: Jobs with multiple OutputDataStream chained together
Right now we support a single OutputDataStream, but when adding more somehow we run into situations where AnalysisListener.componentBegin is called thousands of times and typically also the components are wrongly configured. Somewhere there must be a recursion gone wrong or something :-)

Can be reproduced simply by adding 3 Completeness analyzers to a job. Chain them all together by using e.g. the "Incomplete rows" stream. It works with 2 Completeness analyzers (1 output data stream) but breaks at 3+.

kaspersorensen:
Fixes #583 by introducing an identifyable "RowProcessingStream" to hold Table and AnalysisJob

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
+++ b/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
@@ -175,7 +175,7 @@ public class ConsumeRowHandler {
         final RowProcessingPublisher publisher;
         if (rowConsumeConfiguration.table != null) {
             @SuppressWarnings("deprecation")
-            RowProcessingPublisher tablePublisher = rowProcessingPublishers
+            final RowProcessingPublisher tablePublisher = rowProcessingPublishers
                     .getRowProcessingPublisher(rowConsumeConfiguration.table);
             if (tablePublisher == null) {
                 throw new IllegalArgumentException("Job does not consume records from table: "
--- a/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
+++ b/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
@@ -174,11 +174,14 @@ public class ConsumeRowHandler {
 
         final RowProcessingPublisher publisher;
         if (rowConsumeConfiguration.table != null) {
-            publisher = rowProcessingPublishers.getRowProcessingPublisher(rowConsumeConfiguration.table);
-            if (publisher == null) {
+            @SuppressWarnings("deprecation")
+            RowProcessingPublisher tablePublisher = rowProcessingPublishers
+                    .getRowProcessingPublisher(rowConsumeConfiguration.table);
+            if (tablePublisher == null) {
                 throw new IllegalArgumentException("Job does not consume records from table: "
                         + rowConsumeConfiguration.table);
             }
+            publisher = tablePublisher;
         } else {
             final Collection<RowProcessingPublisher> publisherCollection = rowProcessingPublishers
                     .getRowProcessingPublishers();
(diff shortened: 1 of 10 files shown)


engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
```diff
@@ -174,11 +174,14 @@ public ConsumeRowResult consumeRow(final InputRow row) {
 
         final RowProcessingPublisher publisher;
         if (rowConsumeConfiguration.table != null) {
-            publisher = rowProcessingPublishers.getRowProcessingPublisher(rowConsumeConfiguration.table);
-            if (publisher == null) {
+            @SuppressWarnings("deprecation")
+            RowProcessingPublisher tablePublisher = rowProcessingPublishers
```
LosD: final missing. But why introduce the temp variable at all? Original seemed clearer.

engine/core/src/main/java/org/datacleaner/job/runner/RowProcessingPublishers.java
```diff
@@ -134,13 +141,30 @@ private void registerOutputDataStream(final RowProcessingPublisher parentPublish
         return physicalColumnsArray;
     }
 
+    @Deprecated
```
LosD: JavaDoc with a deprecation reason here would be nice

engine/core/src/main/java/org/datacleaner/job/runner/RowProcessingStream.java
```diff
@@ -0,0 +1,97 @@
...
+        return _table;
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((_analysisJob == null) ? 0 : _analysisJob.hashCode());
+        result = prime * result + ((_table == null) ? 0 : _table.hashCode());
+        return result;
```
LosD: Objects.hash(...)?

engine/core/src/main/java/org/datacleaner/job/runner/RowProcessingStream.java
```diff
@@ -0,0 +1,97 @@
...
+                return false;
+        } else if (!_analysisJob.equals(other._analysisJob))
+            return false;
+        if (_table == null) {
+            if (other._table != null)
+                return false;
+        } else if (!_table.equals(other._table))
+            return false;
+        return true;
+    }
```
LosD: Something like:
```
if (this == object) {
    return true;
} else if (obj == null || !(obj instanceof RowProcessingStream)) {
    return false;
} else {
    return Objects.equals(_analysisJob, other._analysisJob && Objects.equals(_table, other._table)
}
```
Seems quite a bit more readable to me.

engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
```diff
@@ -88,7 +93,10 @@ public void onComplete(Task task) {
 
     @Override
     public void onError(Task task, Throwable throwable) {
-        _success.set(false);
+        boolean alreadyRegisteredError = _errorsReported.getAndSet(true);
+        if (!alreadyRegisteredError) {
```
LosD: Why introduce a variable here?

LosD:
Looks good, besides minor stuff.

engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
kaspersorensen: Right. The temp variable is needed for the ```@SuppressWarnings``` annotation :S

engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
kaspersorensen: Mostly because I find the meaning of the return of .getAndSet a little bit hard to infer. So naming it makes it clearer to read the if sentence: "if not already registered error".

PR #600, opened 2015-08-27
Skip header lines of CSV files in Spark module

### LINKED ISSUE #598: Spark job execution reports one more record than desktop execution
When the same DC job is run in desktop app and in the Spark cluster, the number of records reported by this job is off by one.

Most probably, Spark loads the whole file into RDD (including the header line) while desktop runner filters the header line out. Spark runner needs to take into account header line number from CsvConfiguration in input datastore and filter out this line from the RDD.

kaspersorensen:
Fixes #598

### DIFF
--- a/spark/src/main/java/org/datacleaner/spark/SkipHeaderLineFunction.java
+++ b/spark/src/main/java/org/datacleaner/spark/SkipHeaderLineFunction.java
@@ -1,0 +1,40 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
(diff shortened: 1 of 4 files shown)


tomaszguzialek:
Code-wise it looks good, but I did not see the difference in number of records between this branch and the master. Could not verify now with desktop because of #602

kaspersorensen:
The unittest shows it (row count = 7)

LosD:
Travis is angry, though it seems unrelated

tomaszguzialek:
True, the unit proves it. I must have missed sth important in my manual testing (maybe some component filtered the header row already). I cannot reproduce this discrepancy now, I will merge it.

PR #601, opened 2015-08-27
Simplify Salesforce configuration

### LINKED ISSUE #517: A better widget for entering Salesforce.com endpoint URL
Today we have just a plain text field for entering SFDC's endpoint URL:
image

While in 99% cases the URL would either be blank (default) or the test/sandbox URL: https://test.salesforce.com/services/Soap/u/28.0

So let's instead make an (editable) combobox with those two choices and then also the option to enter a custom URL.

LosD:
This makes the Salesforce dialog a little simpler, with no possibility to enter your own endpoint, only a combobox with default or test endpoints. In respecting users with special needs, any externally entered URL will be added to the list.

Fixes #517 

With "normal" endpoints:
image

With externally configured endpoint
image


kaspersorensen:
Looking at their website they don't use a capital "F" in "Sales__f__orce" so I think we should correct that in the labels :-)

LosD:
Whoops, well found. Fix committed.

LosD:
There, now the PR title is also correct :)

kaspersorensen:
Looking good.

PR #608, opened 2015-08-28
Hiding preview button for child-job transformers

### LINKED ISSUE #605: Disable transformer "preview" button for transformers in non-main-scope
With output data streams we cannot easily provide a "preview" functions for transformers. So if we show a transformer dialog then we should disable/hide the preview button if the transformer does not pertain to the main/root AnalysisJobBuilder.

kaspersorensen:
Fixes #605


tomaszguzialek:
LGTM.

PR #609, opened 2015-08-28
Elasticsearch shield

tomaszguzialek:


desktop/api/src/main/java/org/datacleaner/util/WidgetUtils.java
```diff
@@ -655,4 +655,21 @@ public static void setDefaultButtonStyle(AbstractButton b) {
             b.setUI(DefaultButtonUI.get());
         }
     }
-}
\ No newline at end of file
+    
+    public static void disableComponent(JComponent component) {
```
kaspersorensen: This seems very unrelated? And not really the way I would prefer disabling to work because it is ultimately a ComponentUI responsibility. Background colors etc. to use for disabled/enabled will vary a lot depending on where the component is.

desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
```diff
@@ -47,25 +58,129 @@
...
         _hostnameTextField = WidgetFactory.createTextField();
         _portTextField = WidgetFactory.createTextField();
         _portTextField.setDocument(new NumberDocument(false));
-        _clusterNameTextField = WidgetFactory.createTextField();
-        _indexNameTextField = WidgetFactory.createTextField();
+        _usernameTextField = WidgetFactory.createTextField();
+        _passwordField = WidgetFactory.createPasswordField();
+        _keystorePathField = new FilenameTextField(userPreferences.getOpenDatastoreDirectory(), true);
+        // FIXME: Hack-ish way to make it fit...
+        final double columns = WidgetFactory.TEXT_FIELD_COLUMNS * 0.6;
```
kaspersorensen: Not sure this is needed ... Many places we combine normal text fields and filename text fields. The filename text field obviously has the additional "Browse" button but it usually looks OK as long as the text field part is aligned with the other boxes.

desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
```diff
@@ -47,25 +58,129 @@
...
+                            _portTextField.setText("9300");
+                            _sslCheckBox.setSelected(DEFAULT_SSL);
+                        }
+                    }
+                }
+
+            }
+        });
+
+        _sslCheckBox = new JCheckBox("SSL", DEFAULT_SSL);
```
kaspersorensen: I would phrase it "Enable SSL?"

desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
```diff
@@ -96,20 +211,53 @@ protected void onChange(DocumentEvent event) {
                 validateAndUpdate();
             }
         });
+        _usernameTextField.getDocument().addDocumentListener(new DCDocumentListener() {
+            @Override
+            protected void onChange(DocumentEvent event) {
+                validateAndUpdate();
+            }
+        });
+        _passwordField.getDocument().addDocumentListener(new DCDocumentListener() {
```
kaspersorensen: Looks like the same DCDocumentListener is created 3 times in this block of code. Would be easier to reuse

engine/core/pom.xml
```diff
@@ -7,6 +8,19 @@
 	</parent>
 	<artifactId>DataCleaner-engine-core</artifactId>
 
+	<repositories>
+		<repository>
+			<id>elasticsearch-releases</id>
+			<url>http://maven.elasticsearch.org/releases</url>
```
kaspersorensen: We're not allowed to add extra repositories to our POM (will mean our deployment to central repo goes haywire). Is this needed?

kaspersorensen: If it's needed we may need to make this a DC commercial edition feature only

engine/xml-config/src/main/resources/configuration.xsd
```diff
@@ -516,6 +516,11 @@
 							</sequence>
 						</complexType>
 					</element>
+					<element name="username" minOccurs="0" maxOccurs="1" type="string" />
+					<element name="password" minOccurs="0" maxOccurs="1" type="string" />
+					<element name="ssl" minOccurs="0" maxOccurs="1" type="boolean" />
+					<element name="keystorePath" minOccurs="0" maxOccurs="1" type="string" />
```
kaspersorensen: Our convention for naming XML nodes is to use dash instead of camelCase. So this one should be "keystore-path" and below "keystore-password".

engine/xml-config/src/main/java/org/datacleaner/configuration/DatastoreXmlExternalizer.java
```diff
@@ -334,6 +334,11 @@ public Element toElement(ElasticSearchDatastore datastore) {
         appendElement(ds, "port", datastore.getPort());
         appendElement(ds, "cluster-name", datastore.getClusterName());
         appendElement(ds, "index-name", datastore.getIndexName());
+        appendElement(ds, "username", datastore.getUsername());
+        appendElement(ds, "password", encodePassword(datastore.getPassword()));
+        appendElement(ds, "ssl", datastore.getSsl());
+        appendElement(ds, "keystorePath", datastore.getKeystorePath());
```
kaspersorensen: Once .xsd is updated, this and the below string should also be updated of course.

desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
tomaszguzialek: I hoped I can just add this component to the grid as it will adjust itself, but it did not work well (at least on my machine). The button went down to the next line and was covered by other components. Originally, this component was used in Excel datastore dialog and it has more space there...

desktop/api/src/main/java/org/datacleaner/util/WidgetUtils.java
tomaszguzialek: We discussed it back in the days and decided that we need to introduce some logic to make the component look disabled - setEnabled(false) did not make a difference visually. I chose "graying out" the component as I had some problem to add/remove it based on the checkbox selection. It is related to the PR, but ofc we can discuss where such code should land. Is your suggestion to override setEnabled methods on the components themselves?

engine/core/pom.xml
tomaszguzialek: It is required, Elasticsearch Shield is not in the central repo.

desktop/api/src/main/java/org/datacleaner/util/WidgetUtils.java
kaspersorensen: No, normally you have a UI class which acts similar to a CSS document/stylesheet. We have for instance DCComboBoxUI, PrimaryButtonUI, DarkButtonUI etc.

This is necesary because not every button or every text field should have the same style, so you don't want it to be applied to _every_ such button or text field.

desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
kaspersorensen: OK. Strange. This is not a big concern, but maybe something to spend an extra 10 minutes on checking (if it works for Excel it ought to be the same here - why isn't it?)

tomaszguzialek:
I removed the Shield dependency (and the Elasticsearch repo). The dependency will be shipped with community edition of DataCleaner (I did a test build and it works). The core build throws NoNodeAvailableException: None of the configured nodes are available: [] when trying to connect to a shielded instance of ES and it does not interfere with unshielded instances.

tomaszguzialek:
Refactored disabling the components.

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
```diff
@@ -757,6 +758,11 @@ private Datastore createDatastore(String name, CassandraDatastoreType datastoreT
     private Datastore createDatastore(String name, ElasticSearchDatastoreType datastoreType) {
         final String clusterName = getStringVariable("clusterName", datastoreType.getClusterName());
         final String hostname = getStringVariable("hostname", datastoreType.getHostname());
+        final String username = getStringVariable("username", datastoreType.getUsername());
+        final String password = getPasswordVariable("password", datastoreType.getPassword());
+        final boolean ssl = getBooleanVariable("ssl", datastoreType.isSsl(), false);
+        final String keystorePath = getStringVariable("keystore-path", datastoreType.getKeystorePath());
+        final String keystorePassword = getPasswordVariable("keystore-password", datastoreType.getKeystorePassword());
```
kaspersorensen: Eh maybe I said it wrong, but actually _here_ we would use camelCase :-P Reason is that this is referring to system-properties which has a different convention.

tomaszguzialek:
es-shield

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
tomaszguzialek: Fixed :).

PR #617, opened 2015-09-02
Moving spark module to engine/env/spark

### LINKED ISSUE #547: Move spark module to engine/env
The spark module represents IMO an environment where the DC engine could run. So would it not make best sense to move the module to ```/engine/env/spark``` instead of just ```/spark``` in the module/directory structure?

tomaszguzialek:
Fixes #547


LosD:
LGTM, merge when Travis is happy.

PR #618, opened 2015-09-02
Feature/reference data improvements

### LINKED ISSUE #182: Revise and optimize the reference data implementations
Current implementations of Dictionary and SynonymCatalog are quite obscure. There are several unstandard ways of archieving some caching features etc. which I think is both risky and bulky... So let's re-implement these quite simple components using a Guava cache and maybe something like a bloom filter or so. And let's make these features pluggable so that the implementation differences become more composeable and clear.

kaspersorensen:
Fixes #182 

Most important in this change is an interface change. Instead of doing the matching/lookups directly on the ReferenceData interfaces, I've introduced "connections", like we have it with Datastores. That allows the ReferenceData objects to only contain the "configuration" information about the reference data and to fetch the actual data when a connection is opened.

So for instance - instead of:
```
Dictionary d = ...
boolean b1 = d.containsValue("foo");
boolean b2 = d.containsValue("bar");
```

Now you would then do

```
Dictionary d = ...
DictionaryConnection dc = d.openConnection(...)
boolean b1 = dc.containsValue("foo");
boolean b2 = dc.containsValue("bar");
dc.close();
```

The difference here is that the explicit scoping of the session/connection allows for much easier handling of state in the implementations, and to consider at "connection creation time" whether something is to be loaded into memory or not etc. - and to garbage collect it when the connection is closed.


components/basic-filters/src/main/java/org/datacleaner/beans/filter/StringPatternFilter.java
```diff
@@ -37,46 +40,67 @@
...
+        this.column = column;
+        this.stringPatterns = stringPatterns;
+        this.matchCriteria = matchCriteria;
+        this.configuration = configuration;
+    }
+
+    public StringPatternFilter() {
+    }
+
+    public void init() {
```
LosD: Missing annotation.

components/basic-filters/src/main/java/org/datacleaner/beans/filter/StringPatternFilter.java
```diff
@@ -37,46 +40,67 @@
...
+        stringPatternConnections = new StringPatternConnection[stringPatterns.length];
+        for (int i = 0; i < stringPatterns.length; i++) {
+            stringPatternConnections[i] = stringPatterns[i].openConnection(configuration);
+        }
+    }
+
+    public void close() {
```
LosD: Missing annotation.

components/basic-filters/src/test/java/org/datacleaner/beans/filter/StringPatternMatchFilterTest.java
```diff
@@ -28,37 +30,45 @@
...
+    private final DataCleanerConfiguration configuration = new DataCleanerConfigurationImpl();
 
-		assertEquals(ValidationCategory.VALID, filter.categorize(new MockInputRow().put(column, "kasper@eobjects.org")));
-		assertEquals(ValidationCategory.INVALID, filter.categorize(new MockInputRow().put(column, "kasper@")));
+    public void testFilterSinglePattern() throws Exception {
+        StringPattern stringPattern = new RegexStringPattern("very simple email pattern", ".+@.+", true);
+        MockInputColumn<String> column = new MockInputColumn<String>("my col", String.class);
+        StringPatternFilter filter = new StringPatternFilter(column, new StringPattern[] { stringPattern },
+                MatchFilterCriteria.ANY, configuration);
+        filter.init();
```
LosD: And here the reason the missing annotations wasn't spotted.

Would it be viable to make a test that runs it through the engine to find these kinds of bugs?

desktop/ui/src/main/java/org/datacleaner/windows/SimpleDictionaryDialog.java
```diff
@@ -65,9 +67,14 @@ protected SimpleDictionaryDialog(@Nullable SimpleDictionary dictionary, MutableR
...
 
+        _caseSensitiveCheckBox = new DCCheckBox<>("Case-sensitive?", false);
+        _caseSensitiveCheckBox.setForeground(WidgetUtils.BG_COLOR_BRIGHTEST);
+        _caseSensitiveCheckBox.setOpaque(false);
+        _caseSensitiveCheckBox.setToolTipText("Only match on dictionary terms when text-case is the same.");
+
         if (dictionary != null) {
             _nameTextField.setText(dictionary.getName());
-            Collection<String> values = dictionary.getValues().getValues();
+            Collection<String> values = dictionary.getValueSet();
```
LosD: final

desktop/ui/src/main/java/org/datacleaner/windows/SimpleDictionaryDialog.java
```diff
@@ -104,25 +112,30 @@ protected JComponent getDialogContent() {
...
+                final String values = _valuesTextArea.getText();
                 if (StringUtils.isNullOrEmpty(values)) {
                     JOptionPane.showMessageDialog(SimpleDictionaryDialog.this, "Please fill out the values");
                     return;
                 }
 
-                SimpleDictionary dict = new SimpleDictionary(name, values.split("\n"));
+                final boolean caseSensitive = _caseSensitiveCheckBox.isSelected();
+
+                SimpleDictionary dict = new SimpleDictionary(name, caseSensitive, values.split("\n"));
```
LosD: final

components/basic-filters/src/test/java/org/datacleaner/beans/filter/StringPatternMatchFilterTest.java
kaspersorensen: I think we should maybe consider having an integration test module where we have a bunch of jobs that enables us to test many different components, and their combinations.

On short term I will do a little trick here to use LifeCycleHelper to initialize the component, but it's not very reader-friendly from a unittesting perspective.

engine/core/src/main/java/org/datacleaner/reference/AbstractReferenceData.java
```diff
@@ -21,61 +21,71 @@
...
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (!(obj instanceof AbstractReferenceData)) {
+            return false;
+        }
+        final AbstractReferenceData other = (AbstractReferenceData) obj;
+        return Objects.equals(getClass(), other.getClass()) && Objects.equals(getName(), other.getName());
```
LosD: You _could_ replace the instanceof check with ```getClass() == other.getClass()``` and skip it here. There's no real benefit with using Objects.equals with getClass(), as it is never null (though you may prefer it for consistency)

engine/core/src/main/java/org/datacleaner/reference/RegexStringPattern.java
```diff
@@ -37,73 +38,78 @@
...
+        return new StringPatternConnection() {
+
+            private final Pattern _pattern = Pattern.compile(_expression);
+
+            @Override
+            public boolean matches(String string) {
+                if (string == null) {
+                    return false;
+                }
+                Matcher matcher = _pattern.matcher(string);
```
LosD: final

engine/core/src/main/java/org/datacleaner/reference/SimpleDictionary.java
```diff
@@ -21,51 +21,147 @@
...
+                } else {
+                    valueSet.add(value.toString().toLowerCase());
+                }
+            }
+        }
+        return valueSet;
+    }
+
+    private void readObject(ObjectInputStream stream) throws IOException, ClassNotFoundException {
+        Adaptor adaptor = new Adaptor() {
```
LosD: final (and the whole variable seems pretty useless, but we probably wont agree on that :))

engine/core/src/main/java/org/datacleaner/reference/TextFileSynonymCatalog.java
```diff
@@ -71,19 +70,61 @@ public TextFileSynonymCatalog(String name, String filename, boolean caseSensitiv
...
+                        if (values.length > 1) {
+                            for (int i = 1; i < values.length; i++) {
+                                synonyms.put(values[i], values[0]);
+                            }
+                        }
+                    }
+                } catch (IOException e) {
+                    throw new IllegalStateException(e);
+                } finally {
+                    FileHelper.safeClose(reader);
```
LosD: Maybe consider try-with-resources

components/pattern-finder/src/main/java/org/datacleaner/reference/SimpleStringPattern.java
```diff
@@ -43,77 +45,95 @@
...
+        _expression = expression;
+        _configuration = configuration;
+    }
+
+    private void readObject(ObjectInputStream stream) throws IOException, ClassNotFoundException {
+        ReadObjectBuilder.create(this, SimpleStringPattern.class).readObject(stream);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
LosD: Not implementing hashCode() when implementing equals() is breaking the contract of hashCode(); If two objects are equal, they must also return the same hashcode.

desktop/ui/src/main/java/org/datacleaner/regexswap/RegexSwapStringPattern.java
```diff
@@ -19,57 +19,53 @@
...
+    public String toString() {
+        return "RegexSwapStringPattern[regex=" + _regex + "]";
+    }
 
-	@Override
-	public String toString() {
-		return "RegexSwapStringPattern[regex=" + _regex + "]";
-	}
+    @Override
+    public boolean equals(Object obj) {
```
LosD: Also hashCode() contract breaking here.

components/reference-data/src/main/java/org/datacleaner/beans/filter/DictionaryFilter.java
```diff
@@ -47,20 +52,39 @@
...
+    @Provided
+    DataCleanerConfiguration configuration;
+
+    private DictionaryConnection dictionaryConnection;
+
     public DictionaryFilter() {
     }
 
-    public DictionaryFilter(InputColumn<String> column, Dictionary dictionary) {
+    public DictionaryFilter(InputColumn<String> column, Dictionary dictionary, DataCleanerConfiguration configuration) {
```
LosD: Any reason to have this? It's only used by it's test, which is in the same package, so the properties should be available through the default scope anyway.

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -53,15 +58,21 @@
...
+    
+    private DictionaryConnection dictionaryConnection;
+
     private final Splitter SPLITTER = Splitter.on(CharMatcher.WHITESPACE).omitEmptyStrings();
 
     public RemoveDictionaryMatchesTransformer() {
     }
 
-    public RemoveDictionaryMatchesTransformer(InputColumn<String> column, Dictionary dictionary) {
+    public RemoveDictionaryMatchesTransformer(InputColumn<String> column, Dictionary dictionary, DataCleanerConfiguration configuration) {
```
LosD: Same with the semi-unneeded constructor here.

components/reference-data/src/main/java/org/datacleaner/beans/transform/StringPatternMatcherTransformer.java
```diff
@@ -22,84 +22,113 @@
...
+
+    @Configured
+    MatchOutputType _outputType = MatchOutputType.TRUE_FALSE;
+
+    @Provided
+    DataCleanerConfiguration _configuration;
+
+    private StringPatternConnection[] stringPatternConnections;
+
+    public StringPatternMatcherTransformer(InputColumn<?> column, StringPattern[] stringPatterns, DataCleanerConfiguration configuration) {
```
LosD: Same with the semi-unneeded constructor here.

components/reference-data/src/main/java/org/datacleaner/beans/transform/SynonymLookupTransformer.java
```diff
@@ -67,15 +72,21 @@
...
+    DataCleanerConfiguration configuration;
+
+    private SynonymCatalogConnection synonymCatalogConnection;
+
     public SynonymLookupTransformer() {
     }
 
     public SynonymLookupTransformer(InputColumn<String> column, SynonymCatalog synonymCatalog,
-            boolean retainOriginalValue) {
+            boolean retainOriginalValue, DataCleanerConfiguration configuration) {
```
LosD: Same with the semi-unneeded constructor here.

components/reference-data/src/main/java/org/datacleaner/beans/transform/DictionaryMatcherTransformer.java
```diff
@@ -45,14 +50,20 @@
...
+	@Provided
+	DataCleanerConfiguration _configuration;
+	
+	private DictionaryConnection[] dictionaryConnections;
 
 	public DictionaryMatcherTransformer() {
 	}
 
-	public DictionaryMatcherTransformer(InputColumn<?> column, Dictionary[] dictionaries) {
+	public DictionaryMatcherTransformer(InputColumn<?> column, Dictionary[] dictionaries, DataCleanerConfiguration configuration) {
```
LosD: Same with the semi-unneeded constructor here.

LosD:
That was a big one. Few comments, but nothing real problematic.

components/pattern-finder/src/main/java/org/datacleaner/reference/SimpleStringPattern.java
kaspersorensen: it was calculated, since two dictionaries with the same name are usually equal - the user is required to give them a unique name :-)

components/reference-data/src/main/java/org/datacleaner/beans/filter/DictionaryFilter.java
kaspersorensen: That whole constructor is only used in test. It's just to make it easy to identify if a unittest has satisfied all the properties or not.

components/pattern-finder/src/main/java/org/datacleaner/reference/SimpleStringPattern.java
LosD: Makes sense

PR #623, opened 2015-09-04
OutputDataStreamBuilder convenience method addition

kaspersorensen:
Added a few convenience methods to easily add columns to an output data stream that are similar to other columns.

### DIFF
--- a/api/src/main/java/org/datacleaner/job/output/OutputDataStreamBuilder.java
+++ b/api/src/main/java/org/datacleaner/job/output/OutputDataStreamBuilder.java
@@ -19,8 +19,10 @@
  */
 package org.datacleaner.job.output;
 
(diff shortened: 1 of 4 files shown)


tomaszguzialek:
LGTM.

LosD:
LGTMToo

kaspersorensen:
LGTM2 ?

tomaszguzialek:
The failure of the "push" build seems unralated to this PR, but we the log suggest we have some race condition we should have a look someday:
```
Failed tests: 

  UsageAwareDatastoreConnectionTest.testCloseNoRaceConditions:152 null
```

kaspersorensen:
Hmm looking at that test it seems that it is missing a Thread.join() call ... I'll check it out a bit ...

PR #624, opened 2015-09-04
Feature/619 spark melon jobs

### LINKED ISSUE #619: Support for OutputDataStreams in Spark

tomaszguzialek:
Fixes #619

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
+++ b/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
@@ -183,9 +183,7 @@ public class ConsumeRowHandler {
             }
             publisher = tablePublisher;
         } else {
-            Collection<RowProcessingPublisher> publishers;
-
-            publishers = rowProcessingPublishers.getRowProcessingPublishers();
+            Collection<RowProcessingPublisher> publishers = rowProcessingPublishers.getRowProcessingPublishers();
             publisher = publishers.iterator().next();
             for (RowProcessingPublisher aPublisher : publishers) {
                 if (aPublisher != publisher) {
--- a/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
+++ b/engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
@@ -172,7 +172,7 @@ public class ConsumeRowHandler {
         final RowProcessingPublishers rowProcessingPublishers = new RowProcessingPublishers(analysisJob,
                 analysisListener, taskRunner, lifeCycleHelper, sourceColumnFinder);
 
-        final RowProcessingPublisher publisher;
+        final Collection<RowProcessingPublisher> publishers;
         if (rowConsumeConfiguration.table != null) {
             @SuppressWarnings("deprecation")
             final RowProcessingPublisher tablePublisher = rowProcessingPublishers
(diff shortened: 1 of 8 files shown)


engine/core/src/main/java/org/datacleaner/job/runner/ConsumeRowHandler.java
```diff
@@ -183,13 +183,18 @@ public ConsumeRowResult consumeRow(final InputRow row) {
...
             publisher = tablePublisher;
         } else {
-            final Collection<RowProcessingPublisher> publisherCollection = rowProcessingPublishers
-                    .getRowProcessingPublishers();
-            if (publisherCollection.size() > 1) {
-                throw new IllegalArgumentException(
-                        "Job consumes multiple tables, but ConsumeRowHandler can only handle a single table's components. Please specify a Table constructor argument.");
+            Collection<RowProcessingPublisher> publishers;
+
+            publishers = rowProcessingPublishers.getRowProcessingPublishers();
```
kaspersorensen: Quite minor, but these three lines could be one I think.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkAnalysisRunner.java
```diff
@@ -91,7 +92,8 @@ public AnalysisResultFuture run(AnalysisJob job) {
         JavaPairRDD<String, AnalyzerResult> finalAnalyzerResultsRDD = namedAnalyzerResultsRDD
                 .mapValues(new ExtractAnalyzerResultFunction());
         
-        finalAnalyzerResultsRDD.persist(StorageLevel.MEMORY_AND_DISK());
+        // TODO: Not needed
+        finalAnalyzerResultsRDD = finalAnalyzerResultsRDD.persist(StorageLevel.MEMORY_AND_DISK());
```
kaspersorensen: I suggest to remove. It's just noisy to have stuff we don't need.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkAnalysisRunner.java
```diff
@@ -115,9 +117,12 @@ public AnalysisResultFuture run(AnalysisJob job) {
     }
 
     private boolean isDistributable(AnalysisJob job) {
-        for (AnalyzerJob analyzerJob : job.getAnalyzerJobs()) {
-            if (!analyzerJob.getDescriptor().isDistributable()) {
-                return false;
+        for (ComponentJob componentJob : _sparkJobContext.getComponentList()) {
+            if (componentJob instanceof AnalyzerJob) {
```
kaspersorensen: Actually _all_ components have a distributable flag, so you shouldn't filter on analyzers here.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -122,16 +126,73 @@ public String getAnalysisJobPath() {
...
     
     public ComponentJob getComponentByKey(String key) {
-        final List<AnalyzerJob> analyzerJobs = getAnalysisJob().getAnalyzerJobs();
-        for (AnalyzerJob analyzerJob : analyzerJobs) {
-            if (key.equals(LabelUtils.getLabel(analyzerJob))) {
-                return analyzerJob;
+        List<ComponentJob> componentJobList = getComponentList();
+        for (int i = 0; i < componentJobList.size(); i++) {
+            final ComponentJob componentJob = componentJobList.get(i);
+            if (key.equals(getComponentKey(componentJob))) {
```
kaspersorensen: It seems that getComponentKey(ComponentJob) is also calling getComponentList() which is not a very quick operation to do I can imagine (quite some graph traversal in it). Maybe we should provide a overridden method where you can also pass in the component list so that this can be done just once.

kaspersorensen: Scratch that ... I just noticed the caching of the list! :)

kaspersorensen:
Looks good to me! Let's have Travis' confirmation :)

tomaszguzialek:
Test failure unrelated:
```
Failed tests: 
  SchedulingServiceImplIntegrationTest.testActiveQuartzTriggersInScenario:160 Unexpected files in target/example_repo/tenant1/results: [target/example_repo/tenant1/results/random_number_generation-1441360801949.analysis.execution.log.xml] expected:<0> but was:<1>
```

PR #629, opened 2015-09-07
Add ordered by clause to paging query if not present yet

HettyK:


drexler42:
This fixes an issue in DataHub connector. Results were inconsistent because queries with LIMIT / OFFSET did not have an order by clause. The change adds an ORDER BY on the PK (field "id" in the MDM case) when there is none in the query itself.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDataSet.java
```diff
@@ -158,6 +160,10 @@ private String createEncodedUri(DataHubConnection connection, Table table) {
     }
 
     private String getQueryString(Query query, Table table) {
+        final OrderByClause orderByClause = query.getOrderByClause();
+        if(orderByClause.isEmpty()) {
```
kaspersorensen: beware that not all queries _can_ have an order by clause. For instance a "SELECT COUNT(*) from table" cannot have an appended order by

HettyK: We are adding the ordered by now only when the select item is not a function (but a plain select of columns).

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDataSet.java
```diff
@@ -158,10 +161,36 @@ private String createEncodedUri(DataHubConnection connection, Table table) {
     }
 
     private String getQueryString(Query query, Table table) {
+        final SelectClause selectClause = query.getSelectClause();
+        if (isPlainSelectOfColumns(selectClause)) {
```
kaspersorensen: Other queries that wouldn't work would be queries with GROUP BY.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDataSet.java
```diff
@@ -158,10 +161,36 @@ private String createEncodedUri(DataHubConnection connection, Table table) {
...
+     * Checks if the select clause is a plain select of columns and not a function e.g. SELECT COUNT (*)
+     * @param selectClause
+     * @return true 
+     */
+    private boolean isPlainSelectOfColumns(SelectClause selectClause) {
+        final boolean hasMultipleSelectItems = selectClause.getItems().size() != 1;
+        if(hasMultipleSelectItems) {
+            return true;
+        }
+        final boolean itemIsNotAFunction = selectClause.getItem(0).getFunction() == null;
```
kaspersorensen: Would be better to iterate through all items and if any of them has a function return false, else return true.

HettyK:
Ok, it takes some commits but now we are checking that none of the select items is a function AND there is no group by present. Only then we are adding an ordered by clause.

kaspersorensen:
LGTM

PR #630, opened 2015-09-07
Propagate builder listeners

### LINKED ISSUE #580: When removing a OutputDataStream-consuming component, graph doesn't update
Whenever a component that consumes an OutputDataStream is removed, the graph should update and remove it from the view. Unfortunately, that doesn't happen.

This is probably due to listeners not being transferred to the new AnalysisJob.

### LINKED ISSUE #594: (Some) listeners should be added to OutputStream AnalysisJobBuilders
We have some issues regarding updating the graph on changes in output streams, as the listeners from the root analysis job builder does not propagate to the output stream job builders. We do not want all listeners to just be copied, but somehow that the listeners are given a choice, maybe using some kind of "outputstream added" event that they could react to.

LosD:
This creates a new listener for AnalysisJobBuilder changes, which is used for propagating listeners to newly consumed AnalysisJobBuilders.

Fixes #594 
Fixes #580


kaspersorensen:
Quick glance over the code looks good!

ClaudiaPHI:
LGTM!

PR #631, opened 2015-09-07
Feature/512 write job variables

### LINKED ISSUE #512: DataCleaner removes the variables from the source definition in jobs
we have a job with a variable definition in the source section. This variable is used to determine the location of the output CSV file at runtime.

```
[... snip ...]
       </columns>
        <variables>
           <variable id="output_file_name" value="undefined"/>
        </variables>       
</source>
[... snip ...]
       <analyzer requires="outcome_0" name="ALL_OUT.csv">
 [... snip ...]
           <properties>
                <property name="File" ref="output_file_name"/>
```

When we open the file in DataCleaner 4.0.8 and then save it again (unmodified):
1. The variable definition dissappears completely
2. The reference to the variable is replaced by `value="some/path/in/.datacleaner"`

Desired functionality:
- Keep the variable definition and reference.

tomaszguzialek:
Fixes #512

### DIFF
--- a/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
+++ b/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
@@ -106,7 +106,7 @@ import com.google.common.collect.Lists;
 
 public class JaxbJobReader implements JobReader<InputStream> {
 
-    public static final String DATACLEANER_JAXB_VARIABLE_PREFIX = "datacleaner.jaxb.variable";
+    public static final String DATACLEANER_JAXB_VARIABLE_PREFIX = "datacleaner.jaxb.variable.";
 
     private static final Logger logger = LoggerFactory.getLogger(JaxbJobReader.class);
 
@@ -943,7 +943,7 @@ public class JaxbJobReader implements JobReader<InputStream> {
                         throw new ComponentConfigurationException("No such variable: " + variableRef);
                     }
                     
-                    builder.getMetadataProperties().put(DATACLEANER_JAXB_VARIABLE_PREFIX + "." + configuredProperty.getName(), variableRef);
+                    builder.getMetadataProperties().put(DATACLEANER_JAXB_VARIABLE_PREFIX + configuredProperty.getName(), variableRef);
                 }
 
                 final Class<? extends Converter<?>> customConverter = configuredProperty.getCustomConverter();
--- a/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
+++ b/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
@@ -943,7 +943,7 @@ public class JaxbJobReader implements JobReader<InputStream> {
                         throw new ComponentConfigurationException("No such variable: " + variableRef);
                     }
                     
-                    builder.getMetadataProperties().put(DATACLEANER_JAXB_VARIABLE_PREFIX + "." + descriptor.getDisplayName() + "." + configuredProperty.getName(), variableRef);
+                    builder.getMetadataProperties().put(DATACLEANER_JAXB_VARIABLE_PREFIX + "." + configuredProperty.getName(), variableRef);
                 }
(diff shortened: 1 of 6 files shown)


engine/xml-config/src/test/java/org/datacleaner/job/JaxbJobWriterTest.java
```diff
@@ -365,6 +368,19 @@ public void testReadAndWriteOutputDataStreamsJob() throws Exception {
...
+    public void testWriteVariable() throws Exception {
+        final DescriptorProvider descriptorProvider = new ClasspathScanDescriptorProvider().scanPackage(
+                "org.datacleaner", true);
+        CsvDatastore datastore = new CsvDatastore("date-datastore", "src/test/resources/example-dates.csv");
+        DataCleanerConfiguration configuration = new DataCleanerConfigurationImpl().withDatastores(datastore)
+                .withEnvironment(new DataCleanerEnvironmentImpl().withDescriptorProvider(descriptorProvider));
+        JaxbJobReader reader = new JaxbJobReader(configuration);
+        File file = new File("src/test/resources/example-job-variables.xml");
+        AnalysisJobBuilder ajb = reader.create(file);
+
```
kaspersorensen: I'm missing here an assertion that shows that the reading of the variable has gone well.

kaspersorensen: And maybe that there is the temporary metadata about it.

engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
```diff
@@ -105,6 +106,8 @@
 
 public class JaxbJobReader implements JobReader<InputStream> {
 
+    public static final String DATACLEANER_JAXB_VARIABLE_PREFIX = "datacleaner.jaxb.variable";
```
kaspersorensen: It seems that everywhere that this constant is referred to, it is also concatenated with a dot. So maybe a slight improvement would be to simply add the dot here and avoid all the extra concatenations :-)

engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
```diff
@@ -443,6 +446,12 @@ private AnalysisJobBuilder create(JobType job, SourceColumnMapping sourceColumnM
                         convertToDate(metadata.getUpdatedDate()), datastore.getName(), getSourceColumnPaths(job),
                         getSourceColumnTypes(job), variables, getMetadataProperties(metadata));
                 analysisJobBuilder.setAnalysisJobMetadata(immutableAnalysisJobMetadata);
+            } else {
+                if (variables.size() > 0) {
+                    final MutableAnalysisJobMetadata mutableAnalysisJobMetadata = new MutableAnalysisJobMetadata();
```
kaspersorensen: Here the metadata is being overwritten but I think actually rather it should only set/add variables to any existing metadata. Or else if there was already some metadata (about other things than variables) it will get wiped by this action.

tomaszguzialek: It is in the else block and the if statement checks for metadata being null.

engine/xml-config/src/test/java/org/datacleaner/job/JaxbJobWriterTest.java
tomaszguzialek: There was already a test in the JaxbJobReaderTest.testReadVariable. This one compares the output of the write with the benchmark and it needs to have the variable written back.

engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
tomaszguzialek: It felt ugly with a dot at the end :-P. But OK, I will inline the dot... :).

kaspersorensen:
LGTM

PR #638, opened 2015-09-07
Feature/635 vd grey color support

### LINKED ISSUE #635: Value distribution chart should have color mapping for value "GREY"
Observation at workshop today: We have nice color-mapping for colors/values like GREEN, RED, ORANGE. But not for GREY which is also a common traffic light color. We should add it.

kaspersorensen:
Fixes #635


tomaszguzialek:
LGTM.

PR #639, opened 2015-09-07
Value distribution drill button showing when there's no sample set

### LINKED ISSUE #633: Green "drill to details" arrow button is visible in Value Distribution result even though sample set is missing
Observation at workshop today: Running a Value Distribution on a table with many distinct values (more than 500 at least) caused sample sets to not be available. That's working OK as designed, but the problem was that the green arrow button was still shown even for sets that was not yielding any drill results.

kaspersorensen:
Fixes #633

### DIFF
--- a/desktop/ui/src/test/java/org/datacleaner/widgets/result/ValueDistributionResultSwingRendererGroupDelegateTest.java
+++ b/desktop/ui/src/test/java/org/datacleaner/widgets/result/ValueDistributionResultSwingRendererGroupDelegateTest.java
@@ -28,100 +28,97 @@ import org.datacleaner.result.SingleValueFrequency;
 import org.datacleaner.result.ValueCountListImpl;
 import org.datacleaner.storage.RowAnnotationImpl;
 
-public class ValueDistributionResultSwingRendererGroupDelegateTest extends
-		TestCase {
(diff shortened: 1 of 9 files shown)


desktop/ui/src/test/java/org/datacleaner/widgets/result/ValueDistributionResultSwingRendererGroupDelegateTest.java
```diff
@@ -28,100 +28,97 @@
 import org.datacleaner.result.ValueCountListImpl;
 import org.datacleaner.storage.RowAnnotationImpl;
 
-public class ValueDistributionResultSwingRendererGroupDelegateTest extends
```
tomaszguzialek: I think there are no changes in the test, depite the diff. Are we covered properly by test?

tomaszguzialek:
LGTM, but just making sure we are covered by tests.

desktop/ui/src/test/java/org/datacleaner/widgets/result/ValueDistributionResultSwingRendererGroupDelegateTest.java
kaspersorensen: Right. Only the new constructors have changed and then indentation.

PR #644, opened 2015-09-08
Column IDs unique between scopes

### LINKED ISSUE #642: Column ID clashes when saving OutputDataStream jobs to XML
I am unable to reopen a job where there is both a (source table) column with id "col_city" and inside a child job there is another column with id "col_city".

The error I get is when loading/reading the job file, but it seems to me that the real cause is when it is being saved - it should not have ID clashes.

tomaszguzialek:
Fixes #642

### DIFF
--- a/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobWriter.java
+++ b/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobWriter.java
@@ -124,7 +124,8 @@ public class JaxbJobWriter implements JobWriter<OutputStream> {
     }
 
     private void configureJobType(final AnalysisJob analysisJob, final JobType jobType, boolean includeMetadata) {
-        configureJobType(analysisJob, jobType, null, includeMetadata);
+        final BiMap<InputColumn<?>, String> columnMappings = HashBiMap.create(50);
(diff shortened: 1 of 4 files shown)


engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobWriter.java
```diff
@@ -162,7 +166,10 @@ private void configureJobType(final AnalysisJob analysisJob, final JobType jobTy
             sourceType.setDataContext(dataContextType);
         }
         // mappings for lookup of ID's
-        final BiMap<InputColumn<?>, String> columnMappings = HashBiMap.create(50);
+        if (columnMappings == null) {
+            columnMappings = HashBiMap.create(50);
```
kaspersorensen: would be a bit nicer to have this in line 127 instead of the null.

kaspersorensen:
LGTM... Only a minor cosmetical thingie

engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobWriter.java
tomaszguzialek: True, fixing now.

kaspersorensen:
LGTM

PR #645, opened 2015-09-08
REST documentation

jhorcicka:
This pull request contains a Swagger controller and it's simple web application (human readable user documentation) and a WADL controller (XML documentation).

### DIFF
--- a/monitor/services/pom.xml
+++ b/monitor/services/pom.xml
@@ -193,5 +193,23 @@
 			<artifactId>javax.el-api</artifactId>
 			<scope>provided</scope>
 		</dependency>
+
+        <!-- REST documentation -->
+        <dependency>
+            <groupId>org.jvnet.ws.wadl</groupId>
+            <artifactId>wadl-core</artifactId>
+            <version>1.1.6</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jvnet.ws.wadl</groupId>
+            <artifactId>wadl-client-plugin</artifactId>
+            <version>1.1.6</version>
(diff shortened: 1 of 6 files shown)


kaspersorensen:
I am testing this right now but don't think I understand it ...

I deploy it locally and go to http://localhost:8080/DataCleaner-monitor/swagger/
What I see is a generic swagger page with ```http://petstore.swagger.io/v2/swagger.json``` prefilled.
Nothing here about DataCleaner and it leads me to think that

1) probably there's an endpoint in my application that I should point Swagger to
2) probably we shouldn't put swagger _itself_ in the war file. Swagger seems to me more like an application that you would use externally if you wish.

Is that right? If so then I suggest not to add the whole uncompressed swagger directory to the .war file. Rather just say that we have (somewhere?) a swagger URL that you can use _if_ you use swagger. This stuff I would like to see in the datacleaner documentation.

kaspersorensen:
On the up-side... I do like the wadl part :-)

kaspersorensen:
OK I now figured out to point swagger UI to http://localhost:8080/DataCleaner-monitor/repository/api-docs

But the next problem seems to be that any request I try out with swagger is failing because of authentication issues.

jhorcicka:
OK. Swagger-UI webapp was removed from the war file. Problem with the failing requests was in the base URL which was set to "/" instead of "/repository". Now it is fixed.

kaspersorensen:
LGTM. I didn't do any manual testing this time around, but trust that you did and verified that Swagger is working now.

PR #652, opened 2015-09-09
Bug/634 connect single input component

### LINKED ISSUE #634: Using "Link to" to Phone correction doesn't work
Observation at workshop today: When using phone correction, the "link to" way of building relationships doesn't seem to work. I think this is because the Phone correction component has multiple input fields. But a default should be selected IMO.

tomaszguzialek:
Fixes #634


kaspersorensen:
LGTM :-)

PR #654, opened 2015-09-10
Feature/653 remote component provider

### LINKED ISSUE #653: Remote AnalyzerBeans provider in conf.xml
For the purpose of the remote-transformers epic, we need to support more ComponentDescriptorProviders configurable through conf.xml.

jakubneubauer:
Just to be clear - this PR contains just a "skeleton" for the "remote components descriptor provider". The real implementation will come later. fixes #653


engine/core/src/main/java/org/datacleaner/descriptors/CompositeDescriptorProvider.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Collection;
+import java.util.Set;
+import java.util.TreeSet;
+
+/**
+ * DescriptorProvider that provides a composite view of descriptors from 2 delegate providers.
+ */
+public class CompositeDescriptorProvider implements DescriptorProvider {
+
+    private DescriptorProvider delegate1, delegate2;
```
kaspersorensen: Let's make these final, for good sake :-)

engine/core/src/main/java/org/datacleaner/descriptors/CompositeDescriptorProvider.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    public TransformerDescriptor<?> getTransformerDescriptorByDisplayName(String name) {
+        TransformerDescriptor<?> result = delegate1.getTransformerDescriptorByDisplayName(name);
+        if(result != null) { return result; }
+        return delegate2.getTransformerDescriptorByDisplayName(name);
+    }
+
+    @Override
+    public Collection<FilterDescriptor<?, ?>> getFilterDescriptors() {
+        return new CompositeCollection(new Collection[] {delegate1.getFilterDescriptors(), delegate2.getFilterDescriptors()});
```
kaspersorensen: Just wondering: Should we do anything to get rid of duplicates here (and the two other methods for transformers and analyzers)? Maybe we should not worry about it yet, but I am just thinking that since the "get by name" methods will intrinsically only return a single component descriptor, maybe it's weird to provide multiple here?

engine/core/src/main/java/org/datacleaner/descriptors/RemoteDescriptorProvider.java
```diff
@@ -0,0 +1,97 @@
...
+    private final Map<String, FilterDescriptor<?, ?>> _filterBeanDescriptors = new HashMap<String, FilterDescriptor<?, ?>>();
+    private final Map<String, TransformerDescriptor<?>> _transformerBeanDescriptors = new HashMap<String, TransformerDescriptor<?>>();
+    private final Map<String, RendererBeanDescriptor<?>> _rendererBeanDescriptors = new HashMap<String, RendererBeanDescriptor<?>>();
+
+    public RemoteDescriptorProvider(TaskRunner taskRunner, String url, String username, String password) {
+        this.url = url;
+        this.username = username;
+        this.password = password;
+        latch = new CountDownLatch(1);
+        taskRunner.run(new Task() {
```
kaspersorensen: Instead of using TaskRunner here I would suggest using a LazyRef. Normally we only use the task runner at the job-execution level and in this case I guess you _always_ want multithreading (whereas TaskRunner allows e.g. singlethreaded implementations). The LazyRef also uses a shared thread pool and it has methods built-in for checking the status, so you can also get rid of the CountDownLatch in here.

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
```diff
@@ -52,56 +52,16 @@
 import org.apache.metamodel.util.SimpleTableDef;
 import org.apache.metamodel.xml.XmlSaxTableDef;
 import org.datacleaner.api.RenderingFormat;
-import org.datacleaner.configuration.jaxb.AbstractDatastoreType;
-import org.datacleaner.configuration.jaxb.AccessDatastoreType;
-import org.datacleaner.configuration.jaxb.BerkeleyDbStorageProviderType;
-import org.datacleaner.configuration.jaxb.CassandraDatastoreType;
-import org.datacleaner.configuration.jaxb.ClasspathScannerType;
+import org.datacleaner.configuration.jaxb.*;
```
kaspersorensen: While I do kind of "like" this different way of organizing imports, I fear that we will every time be committing back and forwards with new styles of organizing imports because we use different IDEs. I don't think eclipse has a way to do it the same way as IntelliJ does, so I guess I would like to ask you to revert to the Eclipse-way where every import is explicitly imported and

engine/xml-config/src/main/resources/configuration.xsd
```diff
@@ -124,6 +128,33 @@
...
+                    <documentation>
+                        Defines a custom component by a class name and properties values
+                    </documentation>
+                </annotation>
+            </element>
+            <element name="classpath-scanner" type="ab:classpathScannerType" minOccurs="0" maxOccurs="1">
+                <annotation>
+                    <documentation>
+                        Defines which java packages should be scanned for
+                        components to use in AnalyzerBeans.
```
kaspersorensen: Oh, I realize this is a copy-paste, but actually we should replace "AnalyzerBeans" with "DataCleaner" in this file.

FYI - AnalyzerBeans was the old name used for the "engine" of DataCleaner. Now it is fully apart of DC.

engine/xml-config/src/main/resources/configuration.xsd
```diff
@@ -52,6 +52,9 @@
 				</choice>
 				<element name="storage-provider" type="ab:storageProviderType"
 					minOccurs="0" maxOccurs="1" />
+				<element name="descriptor-providers" type="ab:descriptorProvidersType" minOccurs="1" maxOccurs="1" />
```
kaspersorensen: I guess ```minOccurs``` should be 0. Both to allow the deprecated variant, but also because in many cases (especially for conf.xml files on the monitor server) we would want the descriptor-provider not to be specified here.

engine/xml-config/src/main/resources/configuration.xsd
```diff
@@ -140,6 +171,27 @@
 		</sequence>
 	</complexType>
 
+    <complexType name="remoteComponentsType">
+        <sequence>
+            <element name="server" type="ab:serverType"/>
+        </sequence>
+    </complexType>
+
+    <complexType name="serverType">
```
kaspersorensen: Can we qualify this a bit more since this file contains many types of servers (database/datastore servers etc.). How about "removeComponentServer"?

engine/xml-config/src/main/resources/configuration.xsd
```diff
@@ -140,6 +171,27 @@
...
+        <sequence>
+            <element name="url" type="string" minOccurs="0" maxOccurs="1">
+                <annotation>
+                    <documentation>
+                        Defines URL to the Datacleaner monitor server
```
kaspersorensen: Actually I am already wondering if this is the base URL (such as http://localhost:8080/DataCleaner-monitor) or the URL to a specific web service or so. Maybe we should add that to this description.

And very minor: The 'c' in DataCleaner should be capitalized.

engine/core/src/main/java/org/datacleaner/descriptors/CompositeDescriptorProvider.java
jakubneubauer: In C++ I would use a template maybe. But how to do it in Java? Something like this comes to my mind - unify all "get by name" method into this one:
<T> T getComponentByName(String name, Class<T> componentClass)
The implementation can then use a map of lists (key=componentClass), or just simple bunch of "if"s by the class name...

engine/xml-config/src/main/resources/configuration.xsd
jakubneubauer: I believe that the services sub-paths cannot be customized by user on the server-side. If true, so is it fixed, than I would require just the base URL in the conf.xml, the rest should be added by the program.

engine/core/src/main/java/org/datacleaner/descriptors/RemoteDescriptorProvider.java
jakubneubauer: I wasn't aware of LazyRef.requestLoad(), which makes it suitable for this case (pre-loading in the background).

kaspersorensen:
LGTM!

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
MennoB: [cid:image001.png@01D0EEC4.D3B2B750]

Eclipse has this too, of course. See the two settings in the lower right corner of the dialogue above. This means that by default eclipse needs a 100 imports from the same package before it will automatically provide you with a wildcard (*) import statement. Personally I think it’s safer to rarely (if ever) use this wild card, because, although it’s cleaner, it’s more prone to lead to class name conflicts.

Menno.

PR #657, opened 2015-09-10
Pressing "Esc" key now behaves like Cancel button. fixes #656

### LINKED ISSUE #656: Wrong behavior of "Save changes?" dialog on "Esc" key
Steps to reproduce: Modify job, close the window. A dialog appears with "The job has unsaved changes. what would you like to do?". Pressing "Esc" key closes the windows without saving. It should behave like "Cancel" button, leaving the job window unclosed.

jakubneubauer:
fixes #656

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
@@ -751,7 +751,7 @@ public final class AnalysisJobBuilderWindowImpl extends AbstractWindow implement
 
                 if (unsavedChangesChoice == 0) { // save changes
                     _saveButton.doClick();
-                } else if (unsavedChangesChoice == 2) { // cancel closing
+                } else if (unsavedChangesChoice != 1) { // cancel closing
                     return false;
                 }
             }


desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -751,7 +751,7 @@ protected boolean onWindowClosing() {
 
                 if (unsavedChangesChoice == 0) { // save changes
                     _saveButton.doClick();
-                } else if (unsavedChangesChoice == 2) { // cancel closing
+                } else if (unsavedChangesChoice != 1) { // cancel closing
```
kaspersorensen: This looks right, but could we switch to using the symbolic constants from JOptionPane instead of the int values here? I know this is nothing changed in your PR, but since you're doing this improvement we might as well improve the code readabiity by comparing with ```JOptionPane.WHATEVER``` instead of 0 and 1

kaspersorensen:
Thanks for the improvement - certainly is valid. But added a minor request seeing that you're anyways taking a bit of care for this code :-)

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
jakubneubauer: The return value is index of the chosen option. So using constant doesn't makes sense here. Since JOptionPane javadoc is not really clear, I verified this in JOptionPane code:
```java
        for(int counter = 0, maxCounter = options.length;
            counter < maxCounter; counter++) {
            if(options[counter].equals(selectedValue))
                return counter;
        }
```

jakubneubauer:
Kasper, this is what makes you super product owner - focused on code quality, but in the same time so kind and polite to the committers :-)

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
LosD: If the dialog is closed without choosing anything (ESC or Cancel), the method will return JOptionPane.CLOSED_OPTION:
```
    /** Return value from class method if user closes window without selecting
     * anything, more than likely this should be treated as either a
     * <code>CANCEL_OPTION</code> or <code>NO_OPTION</code>. */
    public static final int         CLOSED_OPTION = -1;
```

Otherwise it will use the indexes from JOptionPane.YES_NO_OPTION:
```
    //
    // Return values.
    //
    /** Return value from class method if YES is chosen. */
    public static final int         YES_OPTION = 0;
    /** Return value from class method if NO is chosen. */
    public static final int         NO_OPTION = 1;
    /** Return value from class method if CANCEL is chosen. */
    public static final int         CANCEL_OPTION = 2;
```

jakubneubauer: Not in the case you provide your own list of options (which is our case).

LosD: That is true, but in our case we still maintain the meaning of yes/no/cancel.

jakubneubauer: Yes, we have the meaning in our mind. But the actual code of JOptionPane doesn't behave according to our mind, it returns the index in the array. Not the constant. See the full code of
```
public static int showOptionDialog(Component parentComponent,
        Object message, String title, int optionType, int messageType,
        Icon icon, Object[] options, Object initialValue)
```
If the caller provides the `options` parameter, the only constant it may return is `CLOSED_OPTION` (if user closes dialog without choice). Otherwise it returns an index in the `options` array. So NOT a YES/NO constant.
Then, interpreting an array index as a constant would be dangerous. It would be mixing apples and pears.

LosD: Yeah, I know that, but I still don't agree.

We match the indexes because we chose to match the indexes. Keeping the code less readable because of an implementation detail makes no sense to me.

I'd even argue that there's a good change that the use of numbers instead of constants helped this bug creep in in the first place.

jakubneubauer: I still believe you are making a mistake, sorry. Make a mental excercise which will (dis)prove your idea: What if Oracle developer decides to change the constant value? Normally, this shouldn't brake anything. But your code would be broken. Because the constant value in JOptionPane.YES_OPTION is totally unrelated to an index in our(!) array of options... You cannot use some constant in you comparison just because now it has - by coincidence - the same value as you need, but the semantic of the constant is totally different than the value you are comparing.

jakubneubauer: Or imagine this: what if the actual value of JOptionPane.YES_OPTION was for example 100, from the beginning of the JOptionPane existence? Then you probably wouldn't get the idea to use it in our case at all.

LosD: Hmmmmm... Good point about internal numbers.

I still do not like the bare numbers, though, as they should still convey meaning. Make our own constants?

kaspersorensen: Oh I thought this was trivial :-) Ehm, then I suggest we just merge as-is since there's at least no obvious better choice!

PR #660, opened 2015-09-11
Feature/647 value dist reducer

tomaszguzialek:
Ref #647 

This PR is competitive to #674, so have a look at both and merge only one!

I am posting this PR to gather feedback if I am on the right track and what I forgot to take care about. Especially Single vs GroupedValueDistributionResult.

The PR also adds a special test of reducing it in Spark + the new parameter to specify minimum number of partitions for the input data (in order to enforce more than one result that can be reduced later).

Also, ValueDistribution is not non-distributable anymore, so now using UniqueKeyCheck to test non-distributable jobs in Spark.

### DIFF
--- a/components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
+++ b/components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
@@ -22,7 +22,9 @@ package org.datacleaner.beans.valuedist;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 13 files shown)


components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
```diff
@@ -0,0 +1,176 @@
...
+    @Inject
+    @Provided
+    RowAnnotationFactory _rowAnnotationFactory;
+
+    @Override
+    public ValueDistributionAnalyzerResult reduce(Collection<? extends ValueDistributionAnalyzerResult> analyzerResults) {
+        final ValueCountList topValues = ValueCountListImpl.emptyList();
+        final Collection<String> uniqueValues = Collections.emptyList();
+        final Map<String, RowAnnotation> annotations = Collections.emptyMap();
+        final InputColumn<?>[] highlightedColumns = null;
```
tomaszguzialek: I was not sure what to use as the annotations, RowAnnotationFactory, highlighted columns as the first value and later as a reduced value. Feedback welcome here.

components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
```diff
@@ -0,0 +1,176 @@
...
+
+        SingleValueDistributionResult reducedResult = new SingleValueDistributionResult(null, topValues, uniqueValues,
+                0, 0, 0, annotations, new RowAnnotationImpl(), _rowAnnotationFactory, highlightedColumns);
+
+        for (ValueDistributionAnalyzerResult partialResult : analyzerResults) {
+            if (partialResult instanceof SingleValueDistributionResult) {
+                final SingleValueDistributionResult singleResult = (SingleValueDistributionResult) partialResult;
+                reducedResult = reduceSingleResult(annotations, highlightedColumns, reducedResult, singleResult);
```
kaspersorensen: I see a pattern here where we reduce only two results at a time. That might cause quite some performance overhead because of moving data out and around between many results. Would be better if the reducer would build one big result in a iterative way based on each result in a for-loop.

components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
```diff
@@ -0,0 +1,176 @@
...
+    private ValueCountList reduceTopValues(ValueCountList topValues1, ValueCountList topValues2) {
+        if (topValues1.getValueCounts().isEmpty()) {
+            return topValues2;
+        }
+
+        ValueCountListImpl topValuesImpl1 = (ValueCountListImpl) topValues1;
+        ValueCountListImpl topValuesImpl2 = (ValueCountListImpl) topValues2;
+
+        for (ValueFrequency valueFrequency2 : topValuesImpl2.getValueCounts()) {
+            for (ValueFrequency valueFrequency1 : topValuesImpl1.getValueCounts()) {
```
kaspersorensen: Many caveats here:

 * If there are unique values it seems that these get added to getValueCounts() as a single entry with key LabelUtils.UNIQUE_LABEL.
 * Since you are running two for-loops inside each other the performance characteristic here will be something like Big O = O^2. I think it should be doable to do it with a Big O = 2*O if you just append each value to a single new result.

tomaszguzialek:
Ready for review again, but have a look at a competitive PR: #674

tomaszguzialek:
Closing as #674 is preferred.

PR #662, opened 2015-09-14
Feature/636 value dist no grouping keep reduce

### LINKED ISSUE #636: Value distribution <count=X> grouping is mostly just annoying
Observation at workshop today (and at other times as well): Users don't seem to understand the way we group values by count in the Value Distribution swing result. We should disable that feature, or at the very least raise the thresholds for it significantly.

tomaszguzialek:
Fixes #636


LosD:
There isn't much new code to review here, so LGTM :)

Merge when Travis happy.

components/value-distribution/src/main/java/org/datacleaner/result/AbstractValueCountingAnalyzerResult.java
```diff
@@ -101,62 +94,9 @@ public int getInstanceCount(String instance) {
             }
 
             return result;
+        } else {
+        	return original;
```
kaspersorensen: Tabs?

Issue #620, opened 2015-09-02
Allow "Fuse / Coalesce" transformer to work on multiple data streams


kaspersorensen:
With the addition of OutputDataStreams in DataCleaner 4.1 we see an old need becoming more and more relevant: To effectively join two data streams producing something a la a (sql) "union" of streams. We already have that capability with the "Fuse / Coalesce fields" transformer but only for fusing filter conditions (and in that process mapping/combining fields to build a new set of master fields).

What I could imagine as a great thing would be that same ability, but working not only on filter conditions but also on output data streams, or even multiple "normal" table data streams.

Design considerations:

 * This could be used for 3 scenarios (or combination of them) then:
  * Two or more component (usually filter) requirements being coalesced.
  * Two or more source table data streams being coalesced.
  * Two or more output data streams being coalesced.
 * I would suggest to no longer model this as a Transformer, but to have it as a plain Component which produces a single OutputDataStream. Reason for this is that the lifecycle (init, close etc.) and visibility of InputColumns of any succeeding components will be specialized to cover this streams lifetime and available fields. The current transformer approach does "leak" the ability for succeeding components to also consume InputColumns which are not available for all the records that have been coalesced.
 * I think this is the only component we will ever build that should have this characteristic. But even though I think it's the only one, I would suggest to add an annotation in engine-core so that we have a way to "mark" the implementation with this characteristic as metadata instead of hardcoding a specific implementation. You never know what we will need and also for unittesting it is good to be able to plug in other implementations.

kaspersorensen:
Relevance points: Suggested by @wvholland @drexler42 as well as HI consultants.

kaspersorensen:
Actually another use case for this component characteristic might be in cross-set matching. Once it is stabilized we might move it to an official API package.

PR #665, opened 2015-09-15
Refactor of RowProcessingPublisher and related classes

kaspersorensen:
While working on #620 I ran into multiple roadblocks and I need to rework a bit how the row processing publishers work. First step in this is to clean up and separate the logic between different types of row processing (source tables are query based, output data streams are pushed), which is what this PR does a bit of.

RowProcessingPublisher was made into an interface. An abstract class is there now: ```AbstractRowProcessingPublisher``` and two implementations: ```SourceTableRowProcessingPublisher```
and ```OutputDataStreamRowProcessingPublisher```.

A minor thing was also fixed in the CloseTaskListener - the success flag that was passed into this class somehow was never looked at. Now if the RowProcessingPublisher itself marks the success flag as false (quite rare) then the CloseTaskListener is also aware of it.


components/basic-transformers/.gitignore
```diff
@@ -2,4 +2,6 @@
 /.settings
 /.classpath
 /target
-/.cache
\ No newline at end of file
+/.cache
+/.cache-main
+/.cache-tests
```
LosD: Newline is recommended at end of files :)

engine/core/src/main/java/org/datacleaner/job/concurrent/CompositeTaskListener.java
```diff
@@ -0,0 +1,56 @@
...
+    private final TaskListener[] _listeners;
+
+    public CompositeTaskListener(TaskListener ... listeners) {
+        _listeners = listeners;
+    }
+    
+    @Override
+    public void onBegin(Task task) {
+        for (TaskListener listener : _listeners) {
+            listener.onBegin(task);
```
LosD: We should probably wrap listeners in try/catch to avoid the whole chain dying (unless that is what we want to achieve). Same for onComplete() and onError()

kaspersorensen: I wouldn't know how to react on any exception so I feel it's better not to try and catch it then.

If it happens then the job will fail for sure, but that also seems correct (generalized since this is a general purpose class).

LosD: :+1:

engine/core/src/main/java/org/datacleaner/job/runner/AnalysisJobMetricsImpl.java
```diff
@@ -45,6 +46,7 @@ public ComponentMetrics getComponentMetrics(ComponentJob componentJob) {
         return new ComponentMetricsImpl(rowProcessingMetrics, componentJob);
     }
 
+    @SuppressWarnings("deprecation")
```
LosD: Any reason we're using a deprecated method?

engine/core/src/main/java/org/datacleaner/job/runner/AnalysisRunnerJobDelegate.java
```diff
@@ -170,15 +153,31 @@ private void scheduleRowProcessing(RowProcessingPublishers publishers, LifeCycle
      * Prevents that any row processing components have input from different
      * tables.
      * 
+     * @param job
+     */
+    private void validateSingleTableInput(AnalysisJob job) {
+        SourceColumnFinder sourceColumnFinder = new SourceColumnFinder();
```
LosD: final

engine/core/src/main/java/org/datacleaner/job/runner/AnalysisJobMetricsImpl.java
kaspersorensen: Unfortunately the AnalysisJobMetrics (and also RowProcessingMetrics btw) has methods on it's interface that is starting to bit-rot a bit. Getting tables __kinda__ makes sense, but is certainly not trivial to read because some tables might represent output data streams. That's why the method was deprecated on the publishers class, but not yet here on the metrics class because the metrics are more direct interface towards the outside world.

LosD: Ugh. Oh well.

LosD:
I'm no fan of all the duplication between the two concrete row publishers, especially in runRowProcessing

kaspersorensen:
Good point. I didn't notice that it has such overlap. I'll munge it a bit more

LosD:
Overall looks better better than before. I wont say "good", since it is still pretty hard to get ones head around, and to follow the flow... But I guess that is just the nature of a beast like this, at some point it's impossible to make complexity any more clear.

kaspersorensen:
Just pushed a refactor of the runRowProcessing method. Could you check the latest commit?

LosD:
Much better :)

LosD:
LGTM

PR #668, opened 2015-09-15
First working version of remote transformers

jakubneubauer:
Current limitations that will be addressed in future development:
 * data types must be available (on classpath) on both client and server side
 * only primitive data types and Lists and Maps are supported.
 * not optimized code for speed.


engine/core/pom.xml
```diff
@@ -81,6 +81,17 @@
...
 		</dependency>
 
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-jsonSchema</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.eobjects.datacleaner</groupId>
+            <artifactId>DataCleaner-rest-client</artifactId>
```
kaspersorensen: Shouldn't the dependencies be inverted here? Now "engine-core" is depending on "rest-client". I would expect that it was opposite: That "rest-client" depended on "engine-core" and that the Transformer implementations etc. are created in that module?

engine/core/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -0,0 +1,251 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.TextNode;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+/**
+ * @Since 9/1/15
+ */
+public class RemoteTransformer implements Transformer {
```
kaspersorensen: Maybe it's premature optimization to think this, but I would suggest to extend BatchTransformer instead of implement Transformer directly. That could also just be a matter of creating that as a separate story tho'.

engine/core/src/main/java/org/datacleaner/descriptors/JsonSchemaConfiguredPropertyDescriptorImpl.java
```diff
@@ -0,0 +1,162 @@
...
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Array;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * @Since 9/1/15
+ */
+public class JsonSchemaConfiguredPropertyDescriptorImpl implements ConfiguredPropertyDescriptor {
```
kaspersorensen: Naming-wise I would find it more natural to call this "RemoteConfiguredPropertyDescriptorImpl" since the "json schema" part is apparently "just" the name of the underlying Jackson module used to abstract it.

engine/core/src/main/java/org/datacleaner/descriptors/TypeBasedConfiguredPropertyDescriptorImpl.java
```diff
@@ -0,0 +1,151 @@
...
+
+import java.lang.annotation.Annotation;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @Since 9/9/15
+ */
+public class TypeBasedConfiguredPropertyDescriptorImpl implements ConfiguredPropertyDescriptor {
```
kaspersorensen: This is the second implementation of ConfiguredPropertyDescriptor in this PR. I guess I am surprised that is needed, but OK. I am at least missing some javadoc to understand what it is actually.

engine/core/pom.xml
jakubneubauer: No, rest-client is very simple client, not using any engine classes. Actually it should not depend on DC api (which I will repair in some next commit). On the other hand, the engine-core now has the remote transformer which uses the rest-client to call the server.

engine/env/spark/pom.xml
```diff
@@ -136,6 +136,22 @@
 					<groupId>org.eobjects.metamodel-extras</groupId>
 					<artifactId>MetaModel-extras-access</artifactId>
 				</exclusion>
+                <exclusion>
```
kaspersorensen: Would be good to not have the remote stuff as part of engine-core, then exclusions like this are not needed anymore. Since we have many projects and products depending on the core we should keep it as light weight as possible and this is a good example of why.

pom.xml
```diff
@@ -982,6 +982,11 @@
 				<artifactId>jackson-databind</artifactId>
 				<version>${jackson.version}</version>
 			</dependency>
+            <dependency>
```
kaspersorensen: In this and all other XML files I notice a different style of indentation. Could we correct that please?

engine/env/spark/pom.xml
jakubneubauer: Hmm... Hard to imagine how to get rid of jackson here. The remote transformer in the engine-core communicates with the server via JSON-based api. And we lay on the data serialization to JSON. Any idea?

engine/core/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
jakubneubauer: Yes, let's do it separately. We also want to compare benchmarks so surely it will be considered.

engine/env/spark/pom.xml
kaspersorensen: OK, I understand that rest-client is then agnostic to DC in general. So what I imagine is a dependency/module structure like this:

 * new module: engine/DataCleaner-rest-client (no DC related dependencies)
 * new module: components/DataCleaner-remote-component (dependencies: DataCleaner-rest-client and DataCleaner-engine-core)
 * new dependency: engine/DataCleaner-xml-config would need to depend on DataCleaner-remote-components, but we can set the dependency to be optional
 * new dependency: desktop/ui would need to depend on DataCleaner-remote-components
* new dependency: monitor/api would need to depend on DataCleaner-rest-client
* new dependency: monitor/services (or maybe only monitor/ui) would need to depend on DataCleaner-remote-components

That way if you build a UI-free application that depends on DC then you dont necesarily have to bring in the remote components (which is unlikely to be relevant in such a case IMO).

jakubneubauer: Seems ok, with question to the DtaCleaner-xml-config. How to beware compile time dependency to DataCleaner-remote-component... The JaxbConfigurationReader from there builds the components descriptor providers (so possibly the remote components descriptor provider, which must depend on DataCleaner-remote-component (or will be part of it)).

Is there any possibility that the JaxbConfigReader could build the Remote descriptor provider in some IoC style? So if it is on classpath, it uses it, if not, will be not available. And without any compile dependency? I can imagine it e.g. with Spring: the bean is injected by classpath scanning by recognizing an interface. Is something similar available in DataCleaner? If not, would be using bit of reflection here a big evil?

kaspersorensen: I think actually we have had the same concern with regard to stuff like the H2StorageProvider or other implementations that was optional. I guess those runtime errors would only occur if the conf.xml file being read actually used those objects. What I mean is that in many cases you would not want remote components in the conf.xml file anyway, and then I guess it would also not try to class load them because the code is never activated, or?

jakubneubauer: That is true, so the dependency would be with scope "provided"?

jakubneubauer: And if really provided by the application, it will work. If not, the remote components section in conf.xml will be not supported (will throw some strange ClassNotFound errors). Maybe these errors could be catched so that the application will not crash.

jakubneubauer: And yet one correction: monitor-services don't need remote-components.

jakubneubauer:
I believe we are ready to merge.

engine/rest-client/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>DataCleaner-rest-client</artifactId>
+
+	<dependencies>
+		<!-- REST client -->
+		<dependency>
+			<groupId>com.sun.jersey</groupId>
+			<artifactId>jersey-client</artifactId>
+			<version>1.19</version>
```
kaspersorensen: I suggest to put this version under <dependencyManagement> in the parent pom, or else it may get overridden in packaging modules (version resolution in Maven is always a dark art)

kaspersorensen:
LGTM except for the little dependency management hint.

engine/rest-client/pom.xml
jakubneubauer: In dependencyManagement? Wouldn't it then brake the hadoop dependency on jersey-client 1.9 ?

kaspersorensen: I hope that 1.9 is compatible with 1.19. But if not then we need to downgrade here because this functionality should also be compatible with the rest I guess.

jakubneubauer: Ok, so I will add it in dependencyManagement. Let us try 1.19 and downgrade only in case of issues, ok?

kaspersorensen: yes

jakubneubauer:
pushed the jersey version management. Should be ready to merge

kaspersorensen:
hehe there's a typo in the property ("jersy" instead of "jersey") but I'm gonna merge anyway :P

PR #669, opened 2015-09-15
Updated the enforcer plugin configurations to catch monitor dependencies

kaspersorensen:
It seems that our configuration of the enforcer plugin used the same id in parent and monitor pom. And unfortunately that means that a few banned dependencies have slipped into our build!

We need to do something about this asap since downstream builds are affected.


jhorcicka:
It should be OK now.

kaspersorensen:
Thanks

PR #670, opened 2015-09-16
Union of source tables

kaspersorensen:
Partially solves #620 ... But only the part of it that has to do with union of source tables.

image

Right now it is functional, so I would advise adding it for the DC 4.5 release. But I would point out a few strange things that I right now consider independent UI bugs:

 * Once configured I would expect a panel to show up in the dialog (see screenshot) showing the output data stream. This panel is only visible if I close the window and open it up again.
 * In the UI it seems that the order of the columns can easily become messed up. I think this is independent because this is the first output data stream we generate which easily changes depending on checks by the user. For instance, if you select first LASTNAME and FIRSTNAME, but then go to select CUSTOMERNUMBER (positioned before those two) then it changes greatly the output data stream metadata. But that does not seem to be reflected properly in UI. But it seems to work fine when saving as analysis.xml file.

Remaining TODOs to fix:

 - [x] Support "union scenario" of source tables.
 - [x] Support "merge scenario" of multiple output data streams.
 - [x] The sorting in AnalysisRunnerJobDelegate still leaves some potential
issues if there are nested OutputDataStreams which contains more than
one MultiStreamComponent.
 - [x] The AnalysisListener.rowProcessingBegin(...) method is invoked twice.
 - [x] Test in the desktop UI.
 - [x] Change analysis.xml format to be able to have the fused streams represented.
  * Represented in separate story: #689
 - [x] Hide "Output columns" panel in UI.
 - [x] Hide "Link to" (from transformer aspect) in UI.
 - [x] Better widget for coalesce units - in this version it seems that unioning of streams is the central aspect. I imagine having a list of available streams/tables and building a matrix of unioned fields with streams/tables on one axis and column on another.
 - [x] Consider naming: Should it actually be called "Union" instead of fuse?
 - [x] Remove the "consider empty string as null" property if this is going to be a union component.
 - [x] Prevent race condition (speculation): Initialization is skipped the second time leading to second stream initializing early. Maybe before initialization actually finishes.


kaspersorensen:
I think this is now ready for the first tests by someone else than myself. For sure a lot of things are NOT working like they should, but some are :-)

Everything related to saving and loading jobs with fusing of streams is NOT handled. But that should maybe also be handled separately to avoid scope creep of this PR. I have reported #689 anad #696 for that.

kaspersorensen:
Looking a bit deeper at the issue with the output data streams panel (OutputDataStreamsViewer) I can see two issues:

 * The columns in that panel seems to be sorted.
 * The task pane container containing it should have a call to .updateUI() when the OutputDataStreamsViewer is removed+added.

kaspersorensen:
New screenshot:

image

api/src/main/java/org/datacleaner/api/OutputColumns.java
```diff
@@ -30,10 +30,24 @@
...
     private final String[] columnNames;
     private final Class<?>[] columnTypes;
 
     /**
+     * Private constructor used to create an empty output columns objects.
+     */
+    private OutputColumns() {
+        columnNames = new String[0];
+        columnTypes = new Class[0];
+    }
```
LosD: What is this needed for?

kaspersorensen: This is for the constant above (NO_OUTPUT_COLUMNS). It is used by MultiStreamComponent which is technically a transformer, but without output columns.

LosD: .... I'm officially blind. It was 8 lines away...

components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceFunction.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public CoalesceFunction(boolean considerEmptyStringAsNull) {
+        _considerEmptyStringAsNull = considerEmptyStringAsNull;
+    }
+
+    public Object coalesce(List<? extends Object> values) {
+        for (final Object value : values) {
+            if (value != null) {
+                if (_considerEmptyStringAsNull) {
+                    if (!"".equals(value)) {
```
LosD: I first wanted to protest about not using String.isEmpty(), but this avoids a cast, and is only a little less readable in itself, so :+1:

components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceUnit.java
```diff
@@ -66,15 +66,31 @@ public CoalesceUnit(InputColumn<?>... inputColumns) {
...
-                        break;
+                        found = true;
                     }
                 }
-                if (_inputColumns[i] == null) {
+                
+                if (!found) {
+                    // try with trimming and case-insensitive matching
+                    for (int j = 0; j < allInputColumns.length; j++) {
+                        InputColumn<?> inputColumn = allInputColumns[j];
```
LosD: Final! :)

components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceUnit.java
```diff
@@ -66,15 +66,31 @@ public CoalesceUnit(InputColumn<?>... inputColumns) {
...
+                        if (name.trim().equalsIgnoreCase(inputColumn.getName().trim())) {
+                            _inputColumns[i] = inputColumn;
+                            found = true;
+                        }
+                    }
+                }
+
+                if (!found) {
                     List<String> names = CollectionUtils.map(allInputColumns, new HasNameMapper());
```
LosD: You can add a final here, now that you're adding finals anyway ;)

components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceUnit.java
```diff
@@ -66,15 +66,31 @@ public CoalesceUnit(InputColumn<?>... inputColumns) {
         if (_inputColumns == null) {
             _inputColumns = new InputColumn[_inputColumnNames.length];
             for (int i = 0; i < _inputColumnNames.length; i++) {
+                boolean found = false;
+
                 final String name = _inputColumnNames[i];
+
+                // first do an exact match round
                 for (int j = 0; j < allInputColumns.length; j++) {
                     InputColumn<?> inputColumn = allInputColumns[j];
```
LosD: Aaaaand here!

components/fuse/src/main/java/org/datacleaner/components/fuse/FuseStreamsComponent.java
```diff
@@ -0,0 +1,136 @@
...
+            final CoalesceUnit unit = _units[i];
+            final InputColumn<?>[] inputColumns = unit.getInputColumns(_inputs);
+            final List<Object> values = inputRow.getValues(inputColumns);
+            final Object value = _coalesceFunction.coalesce(values);
+            output[i] = value;
+        }
+
+        if (logger.isDebugEnabled()) {
+            logger.debug("Fused values for row: {}", Arrays.toString(output));
+        }
```
LosD: No need for the isDebugEnabled(). Just pass the array without toString, that is done by the logger, but only if it needs to output it.

LosD: No... I'm wrong. It's an object array, that will be seen as parameters

components/fuse/src/test/java/org/datacleaner/components/fuse/FuseStreamsComponentIntegrationTest.java
```diff
@@ -0,0 +1,238 @@
...
+                    dataContext.query().from("customers").selectCount().toQuery());
+            Assert.assertEquals(COUNT_CUSTOMERS, countCustomers.getValue(0));
+
+            final Row countEmployees = MetaModelHelper.executeSingleRowQuery(dataContext,
+                    dataContext.query().from("employees").selectCount().toQuery());
+            Assert.assertEquals(COUNT_EMPLOYEES, countEmployees.getValue(0));
+        }
+    }
+
+    @Test(timeout = 20 * 1000)
```
LosD: Consider using a global timeout rule instead:
```java
@Rule
public Timeout globalTimeout = new Timeout(20, TimeUnit.SECONDS);
```
or 
```java 
@Rule
public Timeout globalTimeout = Timeout.seconds(20);
```

desktop/ui/src/main/java/org/datacleaner/panels/fuse/ColumnListMultipleCoalesceUnitPropertyWidget.java
```diff
@@ -55,11 +55,14 @@
 /**
  * {@link PropertyWidget} for two properties at one time: An array of
  * {@link InputColumn}s and an array of {@link CoalesceUnit}s.
+ * 
+ * This widget displays a list of {@link CoalesceUnit}s with {@link InputColumn}
+ * s that the user can add an remove from.
  */
-public class MultipleCoalesceUnitPropertyWidget extends AbstractPropertyWidget<InputColumn<?>[]> implements
+public class ColumnListMultipleCoalesceUnitPropertyWidget extends AbstractPropertyWidget<InputColumn<?>[]> implements
```
LosD: This is a dreadful name... I can't come up with a better suggestion, though.

desktop/ui/src/main/java/org/datacleaner/panels/fuse/ColumnListMultipleCoalesceUnitPropertyWidget.java
```diff
@@ -142,17 +145,20 @@ public void removeCoalesceUnit() {
             _unitContainerPanel.updateUI();
             fireBothValuesChanged();
         }
+        updateUI();
```
LosD: This is a little bit weird call. updateUI() is normally used for a UI (metal, plastic, etc.) change. Don't you want to do a repaint() instead?

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -231,8 +231,7 @@ public AnalysisJobBuilder addSourceColumn(MetaModelInputColumn inputColumn) {
         if (!_sourceColumns.contains(inputColumn)) {
             _sourceColumns.add(inputColumn);
 
-            List<SourceColumnChangeListener> listeners = new ArrayList<>(
-                    _sourceColumnListeners);
+            List<SourceColumnChangeListener> listeners = new ArrayList<>(_sourceColumnListeners);
```
LosD: You touched it, you own it... Final! :)

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -312,14 +311,17 @@ public void importJob(AnalysisJob job) {
     public AnalysisJobBuilder removeSourceColumn(MetaModelInputColumn inputColumn) {
         boolean removed = _sourceColumns.remove(inputColumn);
         if (removed) {
-            List<SourceColumnChangeListener> listeners = new ArrayList<>(
-                    _sourceColumnListeners);
+            List<SourceColumnChangeListener> listeners = new ArrayList<>(_sourceColumnListeners);
```
LosD: Same here

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -312,14 +311,17 @@ public void importJob(AnalysisJob job) {
     public AnalysisJobBuilder removeSourceColumn(MetaModelInputColumn inputColumn) {
         boolean removed = _sourceColumns.remove(inputColumn);
```
LosD: Now that you're at it...

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -385,8 +387,7 @@ public boolean containsSourceColumn(Column column) {
 
         // make a copy since some of the listeners may add additional listeners
         // which will otherwise cause ConcurrentModificationExceptions
-        List<TransformerChangeListener> listeners = new ArrayList<>(
-                _transformerChangeListeners);
+        List<TransformerChangeListener> listeners = new ArrayList<>(_transformerChangeListeners);
```
LosD: More final listener

engine/core/src/main/java/org/datacleaner/job/runner/AbstractRowProcessingConsumer.java
```diff
@@ -254,4 +261,30 @@ public void registerOutputDataStream(OutputDataStreamJob outputDataStreamJob,
     public AnalysisJob getAnalysisJob() {
         return _analysisJob;
     }
+
+    @Override
+    public int onPublisherInitialized(RowProcessingPublisher publisher) {
```
LosD: It's a little bit weird with an listener-like method that returns a count.

LosD:
LGTM, but see comments

PR #672, opened 2015-09-17
Added logging about all CountDownLatch.await() calls in engine and api

kaspersorensen:
This is a small fix to add some logging whenever we are waiting for a long time on CountDownLatches. 

Background: Recently we've seen that the build sometimes stalls in Travis CI ("no output received in 10 minutes"). While I still don't know why that would happen, my guess is that somewhere we maybe have a race condition. And since we call the "normal" CountDownLatch.await() method which blocks infinitely here and there, I thought it would be good to add some logging instead for every 2 minutes that pass by while waiting. That way we gain two things:

 * A bit of logging in general on what is happening, why something is not responding.
 * Maybe good hints from Travis on where we may have some issues.

### DIFF
--- a/api/src/main/java/org/datacleaner/api/AnalyzerResultFutureImpl.java
+++ b/api/src/main/java/org/datacleaner/api/AnalyzerResultFutureImpl.java
@@ -24,6 +24,7 @@ import java.io.ObjectOutputStream;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.metamodel.util.LazyRef;
 import org.apache.metamodel.util.Ref;
@@ -73,7 +74,8 @@ public class AnalyzerResultFutureImpl<R extends AnalyzerResult> implements Analy
                 try {
                     _result = resultRef.get();
                     if (_result == null && resultRef instanceof LazyRef) {
-                        // TODO: workaround - reported as MM bug, remove when fixed.
+                        // TODO: workaround - reported as MM bug, remove when
+                        // fixed.
                         throw new RuntimeException(((LazyRef<?>) resultRef).getError());
                     }
                     onSuccess();
@@ -165,9 +167,17 @@ public class AnalyzerResultFutureImpl<R extends AnalyzerResult> implements Analy
     public R get() {
         if (_countDownLatch != null) {
             try {
-                _countDownLatch.await();
+                boolean finished = false;
+                int iteration = 0;
+                while (!finished) {
+                    iteration++;
+                    finished = _countDownLatch.await(2, TimeUnit.MINUTES);
+                    if (!finished) {
+                        logger.info("Awaited completion for " + (iteration * 2) + " minutes...");
+                    }
+                }
             } catch (InterruptedException e) {
-                // do nothing
+                throw new IllegalStateException("Awaiting completion of AnalyzerResultFuture was interrupted!", e);
             }
         }
         if (_error != null) {
--- a/api/src/main/java/org/datacleaner/api/AnalyzerResultFutureImpl.java
+++ b/api/src/main/java/org/datacleaner/api/AnalyzerResultFutureImpl.java
@@ -169,11 +169,12 @@ public class AnalyzerResultFutureImpl<R extends AnalyzerResult> implements Analy
             try {
                 boolean finished = false;
                 int iteration = 0;
+                final int minutesToWaitPerIteration = 2;
                 while (!finished) {
                     iteration++;
-                    finished = _countDownLatch.await(2, TimeUnit.MINUTES);
+                    finished = _countDownLatch.await(minutesToWaitPerIteration, TimeUnit.MINUTES);
                     if (!finished) {
-                        logger.info("Awaited completion for " + (iteration * 2) + " minutes...");
+                        logger.info("Awaited completion for " + (iteration * minutesToWaitPerIteration) + " minutes...");
                     }
                 }
             } catch (InterruptedException e) {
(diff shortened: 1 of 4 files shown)


kaspersorensen:
An example such stalled build is from #670 ... If you check the build link here: https://travis-ci.org/datacleaner/DataCleaner/jobs/80708914

kaspersorensen:
Wauv and it seems even this PR's build is getting into that mode now ...
```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.012 sec - in org.datacleaner.test.full.scenarios.QueryOptimizationAndAnalysisListeningTest
Running org.datacleaner.test.full.scenarios.JobWithOutputDataStreamsTest
INFO  SourceTableRowProcessingPublisher - Primary key column {} not added to source columns, not pre-selecting primary keys
INFO  RowProcessingPublishers - Registered 2 publishers: [SourceTableRowProcessingPublisher[stream=CUSTOMERS, consumers=1], OutputDataStreamRowProcessingPublisher[stream=foo bar records, consumers=1]]
INFO  AnalysisRunnerJobDelegate - Created 2 row processor publisher(s)
INFO  InfoLoggingAnalysisListener - Processing of 'CUSTOMERS' start
INFO  InfoLoggingAnalysisListener - Processing of 'foo bar records' start
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 2 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 2 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 4 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 4 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 6 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 6 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 8 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 8 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 10 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 10 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 12 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 12 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 14 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 14 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 16 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 16 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'job completion' for 18 minutes...
INFO  ConcurrencyUtils - Awaited completion of 'stream: PushOutputDataStream[name=foo bar records]' for 18 minutes...
```

api/src/main/java/org/datacleaner/api/AnalyzerResultFutureImpl.java
```diff
@@ -165,9 +167,17 @@ public boolean isReady() {
...
         if (_countDownLatch != null) {
             try {
-                _countDownLatch.await();
+                boolean finished = false;
+                int iteration = 0;
+                while (!finished) {
+                    iteration++;
+                    finished = _countDownLatch.await(2, TimeUnit.MINUTES);
+                    if (!finished) {
+                        logger.info("Awaited completion for " + (iteration * 2) + " minutes...");
```
LosD: Let's add a constant for the minute count. We may want to adjust it at some point, and it's easy to miss two values that must be in sync.

api/src/main/java/org/datacleaner/api/AnalyzerResultFutureImpl.java
```diff
@@ -165,9 +167,17 @@ public boolean isReady() {
...
+                    }
+                }
```
LosD: Hmmm... Any reason this isn't using your new ConcurrencyUtils.awaitCountDown?

kaspersorensen: Yes because this class is in the api module where the utils are not available. Kinda lame but I found only this one case :-/

LosD:
Hmmmm.... Hmmmmm... Duplication because poor organization is not really cool. How come the impl is in the API module? *Impl is usually a sign that it belongs nowhere near the API.

kaspersorensen:
The impl is there because from a "consumer of the api" point of view you need to return a concrete class. And I wouldn't want everybody implementing their own AnalyzerResultFuture implementation as it is non-trivial. So providing a standard-implemention should be part of the API I think.

Anyways, back to the PR itself - a solution could be to also move the ConcurrencyUtils to the API module, but that also would kinda suck.

LosD:
Yeah, that would polute the API module even further.

Isn't the analyzerresultfuture usually used by providers, not consumers (I mean in the semantic versioning sense)? Isn't there any package that providers would normally also need to include?

LosD:
(If there's no way around it, let's just merge as-is :))

kaspersorensen:
Currently there's no such package. I guess that's not a bad idea, but out of scope. Merging.

PR #673, opened 2015-09-17
Fixed UI's response to non-analyzers adding "executability" to a job

### LINKED ISSUE #671: "Execute without analyzers" dialog should not show up when there are output data streams or other result producers
The dialog for executing without analyzers (```ExecuteJobWithoutAnalyzersDialog```) is currently coming up any time that there are no analyzers in the job. But this premise is becoming less and less valid since at least the following situations could also be valid execution scenarios:

 * There are output data streams
 * There are other components than analyzers that produce results (such as CountryStandardizer)

kaspersorensen:
Fixes #671 

In short: Analyzers are no longer the only type of component that could make a job worth executing, so this PR relaxes some of the feedback you get in the UI about wanting to execute a job under certain circumstances.


LosD:
LGTM

LosD:
Merge when Travis is happy.

PR #674, opened 2015-09-17
Feature/647 value dist reducer non pairwise

### LINKED ISSUE #647: AnalyzerResultReducer for Value distribution results

tomaszguzialek:
This PR is competitive to #660, so when reviewing have a look at both approaches and merge only one. This approach is iterative reducing while #660 merges one partial result at a time.

Fixes #647

### DIFF
--- a/components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
+++ b/components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
@@ -22,7 +22,9 @@ package org.datacleaner.beans.valuedist;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -181,20 +183,36 @@ public class ValueDistributionAnalyzerResultReducer implements AnalyzerResultRed
     }
 
     private Collection<String> reduceUniqueValues(Collection<String> uniqueValues1, Collection<String> uniqueValues2) {
-        Collection<String> reducedUniqueValues = new ArrayList<>();
-
-        for (String value : uniqueValues1) {
-            if (!uniqueValues2.contains(value)) {
-                reducedUniqueValues.add(value);
+        Map<String, Integer> frequencyMap = new HashMap<String, Integer>();
+        
+        if (uniqueValues1 != null) {
+            for (String value : uniqueValues1) {
+                if (frequencyMap.containsKey(value)) {
+                    frequencyMap.put(value, frequencyMap.get(value) + 1);
+                } else {
+                    frequencyMap.put(value, 1);
+                }
             }
         }
-        for (String value : uniqueValues2) {
-            if (!uniqueValues1.contains(value)) {
-                reducedUniqueValues.add(value);
+        
+        if (uniqueValues2 != null) {
+            for (String value : uniqueValues2) {
+                if (frequencyMap.containsKey(value)) {
+                    frequencyMap.put(value, frequencyMap.get(value) + 1);
+                } else {
+                    frequencyMap.put(value, 1);
+                }
             }
         }
-
-        return reducedUniqueValues;
+        
+        List<String> uniqueList = new ArrayList<>();
+        for (Map.Entry<String, Integer> entry: frequencyMap.entrySet()) {
+            if (entry.getValue() == 1) {
+                uniqueList.add(entry.getKey());
+            }
+        }
+        
+        return uniqueList;
     }
 
 }
--- a/components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
+++ b/components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
@@ -19,6 +19,7 @@
  */
 package org.datacleaner.beans.valuedist;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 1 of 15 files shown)


components/value-distribution/src/main/java/org/datacleaner/result/ValueCountingAnalyzerResult.java
```diff
@@ -61,7 +61,7 @@
 
     public Integer getUnexpectedValueCount();
 
-    public boolean hasAnnotatedRows(String value);
+    public Boolean hasAnnotatedRows(String value);
```
kaspersorensen: Why the change? Do you ever want to allow a null here?

engine/env/spark/src/main/java/org/datacleaner/spark/SparkAnalysisRunner.java
```diff
@@ -54,9 +54,23 @@
...
+        this(sparkContext, sparkJobContext, null);
+    }
+
+    public SparkAnalysisRunner(JavaSparkContext sparkContext, SparkJobContext sparkJobContext, Integer minPartitions) {
         _sparkContext = sparkContext;
         _sparkJobContext = sparkJobContext;
+        if ((minPartitions != null) && (minPartitions > 0)) {
+            _minPartitions = minPartitions;
+        } else {
+            logger.warn(
```
kaspersorensen: Here it's going to raise a warning for something that is the default behaviour ... I would rather only throw an exception if minPartitions is non-null and less-or-equal to 0.

components/value-distribution/src/main/java/org/datacleaner/result/ValueCountingAnalyzerResult.java
tomaszguzialek: Yes, returning null if GroupedValueDistribution has more than one child result.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkAnalysisRunner.java
tomaszguzialek: True. The initial plan was to warn only if minPartitions parameter was specified, but an incorrect value. Will fix it.

components/value-distribution/src/main/java/org/datacleaner/result/ValueCountingAnalyzerResult.java
kaspersorensen: Hmm that's very subtle ... Hard to guess even. I would rather just return false then and require the code to call this on the child results.

kaspersorensen: Right now any calls like this will start throwing exceptions:

```
if (result.hasAnnotatedRows("foo")) {
 ...
}
```

We might better then rename the method so that we can at least catch it at compile time. But in general I don't care about trying to patch the interface here - it's a beast that eventually needs to be redone, but a lot of consideration needs to go into it because at the same time the analyzer has to return some thing from which metrics can automatically be gathered.

components/value-distribution/src/main/java/org/datacleaner/beans/valuedist/ValueDistributionAnalyzerResultReducer.java
```diff
@@ -0,0 +1,291 @@
...
+        Collection<ValueFrequency> flattenedTopValues = getOrInitializeFlattenedTopValues(flattenedTopValuesMap,
+                singlePartialResult);
+        Collection<String> flattenedUniqueValues = getOrInitializeFlattenedUniqueValues(flattenedUniqueValuesMap,
+                singlePartialResult);
+        Collection<ValueFrequency> flattenedDistinctValues = getOrInitializeFlattenedTopValues(
+                flattenedDistinctValuesMap, singlePartialResult);
+
+        flattenedUniqueValues.addAll(singlePartialResult.getUniqueValues());
+        flattenedTopValues.addAll(singlePartialResult.getTopValues().getValueCounts());
+        flattenedDistinctValues.addAll(singlePartialResult.getValueCounts());
```
kaspersorensen: What I am noticing is that ```SingleValueDistributionResult.getValueCounts()``` actually contains all the data you need here ... both all the "top values", the nulls and the uniques. Would it maybe be easier to just get that list of value counts and make up all the rest from it? That way we could reduce the complexity here I think and also avoid moving the same data around into multiple collections.

kaspersorensen:
I prefer this approach over #660 ... shall we close the other one?

tomaszguzialek:
Introducing ReducedValueDistributionResult simplified the code, but it solves the problem only for SingleValueDistributionResults. It does not cover GroupedValueDistributionResults...

kaspersorensen:
Not having time to look at the code, I could still imagine that a grouped result simply holds a list of underlying results. Now the type of the underlying results would be ReducedValueDistributionResult instead of the old SingleValueDistributionResult. No?

tomaszguzialek:
Internally, I could hold a collection of child results, but I need to expose them as collection to the UI. The interface currently does not have any method for this...

kaspersorensen:
I'm not sure we are talking about the same thing then? When I look at GroupedValueDistributionResult I see this method:
```
@Override
    public Collection<? extends ValueCountingAnalyzerResult> getGroupResults() { ... }
```
Since that's just "anything that extends ```ValueCountingAnalyzerResult``` that could also be the reduced ones?

tomaszguzialek:
OK, I haven't seen this. It is a different interface implementing this method. However, I feel we are back in the same place... I can internally hold ```Map<String, Map<String, Integer>>``` representing ```group -> value -> count```. But what should I do with methods like ```public Collection<String> getUniqueValues()?``` Flatten the map (some locally unique values will be then removed)?

kaspersorensen:
hmm you shouldn't re-implement the grouped one I think. It is already pretty OK. You just need some awareness of groups when you combine all the single results.

kaspersorensen:
But yes, in the reducer I think you will need to temporarily keep a ```Map<String, Map<String,Integer>>``` in order to keep the state before building first a ```Map<String, ReducedValueDistributionResult>``` and then turn that into a single ```GroupedValueDistributionResult```.

tomaszguzialek:
Ready for review again...

kaspersorensen:
LGTM! I will manually apply this after squashing it to avoid a confusing commit log :)

PR #677, opened 2015-09-17
bug/676 Changing the output column name now immediately refreshes the column name in the output columns table

### LINKED ISSUE #676: Automatic change in output column name doesn't show in the gui immediately
Create for example a "Build key/value map" transformer, select some columns and change a key names. If you enlarge the dialog so that the output column name is visible, its name is not change until you hover a mouse over it or until you change layout of the dialog for example.

jakubneubauer:


desktop/api/src/main/java/org/datacleaner/panels/MutableInputColumnListPanel.java
```diff
@@ -51,7 +52,18 @@
     private final AnalysisJobBuilder _analysisJobBuilder;
     private final OutputColumnVisibilityButton _visibilityButton;
 
-    public MutableInputColumnListPanel(AnalysisJobBuilder analysisJobBuilder, MutableInputColumn<?> inputColumn) {
+    public JComponent panelOwner;
```
kaspersorensen: Looks to me like this field is _always_ null. It is not set via the constructor.

jakubneubauer: fixed - I forgot the field setting. Was caused by copy&paste from project where I tried it. Not good approach...

kaspersorensen: OK that's fixed. But sorry it is still public and non-final. Any reason for that? Seems to me it should be private final.

jakubneubauer:
thanks, fixed the field visibility etc.

PR #679, opened 2015-09-18
Added banned dependencies that causes conflicts and duplicates

kaspersorensen:
Here's my suggestion for an updated list of banned dependencies. Build will fail, but once we agree on this list we then can resovle the issues.


kaspersorensen:
Merged into #691 - closing this PR.

PR #686, opened 2015-09-18
Feature/666 remote transformers menu items

### LINKED ISSUE #666: Remote transformers menu items
In desktop GUI of DC there are menus with components divided into categories. Remote components (transformers for now) should be added into those menus/categories.

jakubneubauer:
fixes #666


engine/rest-client/src/main/java/org/datacleaner/restclient/ComponentList.java
```diff
@@ -50,11 +51,13 @@ public void add(ComponentInfo componentInfo) {
...
     @JsonInclude(JsonInclude.Include.NON_NULL)
     public static class ComponentInfo {
-
         private String name = "";
         private String description = "";
         private String createURL = "";
+        private Set<String> categoryNames = new HashSet<>();
+        private String superCategoryName = null;
         private Map<String, PropertyInfo> properties = new HashMap<>();
+        private byte[] iconData = null;
```
kaspersorensen: How will this serialize into JSON? Wondering if we should instead expose a URL for the icon.

jakubneubauer: It will serialize as base64 string. If we look on the actual use-case: Desktop queries list of components and always wants an icon for each component. Doing it in one query is more effective than downloading icons one by one after the component list.
BTW, If somebody don't need icons, there is a boolean parameter to the component-list query which will prevent the icons data in the response (default is false).

kaspersorensen: OK that helps. I guess this is OK for now. Desktop might not always be _the_ use case but just one of many. But if the default will not include it then I think it is sensible.

PR #691, opened 2015-09-18
Feature/682 rest api documentation

### LINKED ISSUE #682: REST API documentation in a separate module
WADL and Swagger controllers should be in a separate maven module to be possible to exclude this functionality in particular products.

jhorcicka:
Fixes #682

WADL and Swagger controllers were moved into a separate module (monitor/documentation).


monitor/ui/pom.xml
```diff
@@ -211,7 +211,14 @@
 			<artifactId>DataCleaner-monitor-sugarcrm-datastore-wizard</artifactId>
 			<version>${project.version}</version>
 		</dependency>
-		
+
+        <!-- DataCleaner monitor documentation -->
+        <dependency>
+            <groupId>org.eobjects.datacleaner</groupId>
+            <artifactId>DataCleaner-monitor-documentation</artifactId>
```
kaspersorensen: This dependency means that we are still not in a good situation. Combined with PR #679 you can see that this still introduces the banned dependencies.

kaspersorensen:
I've made a merge with the PR #679 to show exactly where and why this is still an issue. A solution (IMO) is to not include monitor-documentation as a dependency for monitor-ui. But then we have a completely dangling module. Unfortunately the wadl dependency is simply in conflict with a lot of banned dependencies.

kaspersorensen:
I've "brutally" removed the documentation dependency from monitor-ui now. This should fix the issues downstream and for now that's most important. We will merge this once Travis is OK. Sorry if you don't like it @jhorcicka - I propose that we start searching for alternative wadl/swagger libraries with less dependencies.

PR #693, opened 2015-09-22
Feature/649 analyzer result reducer reference data matcher

### LINKED ISSUE #649: AnalyzerResultReducer for Reference data matcher

ClaudiaPHI:
Fixes #649

### DIFF
--- a/components/reference-data/src/main/java/org/datacleaner/beans/ReferenceDataMatcherAnalyzerReducer.java
+++ b/components/reference-data/src/main/java/org/datacleaner/beans/ReferenceDataMatcherAnalyzerReducer.java
@@ -1,0 +1,97 @@
+package org.datacleaner.beans;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import org.datacleaner.api.AnalyzerResultReducer;
+import org.datacleaner.result.Crosstab;
+import org.datacleaner.result.CrosstabDimension;
+import org.datacleaner.result.CrosstabNavigator;
+
+public class ReferenceDataMatcherAnalyzerReducer implements AnalyzerResultReducer<BooleanAnalyzerResult> {
+
+    @Override
+    public BooleanAnalyzerResult reduce(Collection<? extends BooleanAnalyzerResult> partialResults) {
+        if (partialResults.isEmpty()) {
+            return null;
+        }
+
+        final List<CrosstabDimension> columnStatisticCrosstabDimensions = new ArrayList<CrosstabDimension>();
+        final List<CrosstabDimension> columnValueCombinationCrosstabDimensions = new ArrayList<CrosstabDimension>();
+
+        for (BooleanAnalyzerResult partialResult : partialResults) {
+            final Crosstab<Number> partialColumnStatisticsCrosstab = partialResult.getColumnStatisticsCrosstab();
+            final Crosstab<Number> partialValueCombinationCrosstab = partialResult.getValueCombinationCrosstab();
+            createDimensions(columnStatisticCrosstabDimensions, partialColumnStatisticsCrosstab);
+            createDimensions(columnValueCombinationCrosstabDimensions, partialValueCombinationCrosstab);
+        }
+
+        final Crosstab<Number> newResultColumnStatistics = new Crosstab<Number>(Number.class,
+                columnStatisticCrosstabDimensions);
+        final Crosstab<Number> newResultColumnValueCombination = new Crosstab<Number>(Number.class,
+                columnValueCombinationCrosstabDimensions);
+
+        final CrosstabNavigator<Number> columnStatisticsNavigator = new CrosstabNavigator<Number>(
+                newResultColumnStatistics);
+        final CrosstabNavigator<Number> columnValueCombinationNavigator = new CrosstabNavigator<Number>(
+                newResultColumnValueCombination);
+        for (BooleanAnalyzerResult partialResult : partialResults) {
+            final Crosstab<Number> partialColumnStatisticsCrosstab = partialResult.getColumnStatisticsCrosstab();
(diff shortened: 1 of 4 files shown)


components/reference-data/src/main/java/org/datacleaner/beans/ReferenceDataMatcherAnalyzerReducer.java
```diff
@@ -0,0 +1,120 @@
...
+            final List<CrosstabDimension> dimensions = partialCrosstab.getDimensions();
+            for (CrosstabDimension dimension : dimensions) {
+                final List<String> categories = dimension.getCategories();
+                for (String category : categories) {
+                    final Number categoryValue = nav.where(dimension, category).safeGet(null);
+                    final CrosstabNavigator<Number> whereToPut = mainNavigator.where(dimension, category);
+                    if (categoryValue != null) {
+                        final Number oldValue = whereToPut.safeGet(null);
+                        if (oldValue != null) {
+                            final Number newValue = oldValue.doubleValue() + categoryValue.doubleValue();
```
kaspersorensen: Since we are here calling .doubleValue() we will be creating always a new double value. It's not that big a concern, but maybe we can quickly look to see if there is a utility function somewhere (or roll a basic one ourselves) for the sum function. What I have in mind is that if both numbers are e.g. of type Long, then we return a new Long. Having double as a "fall back" would make sense.

kaspersorensen:
Looks quite good to me.

I even think that there is a bit of it we might want to extract as a generic utility function. Because what you are doing here is really traversing the crosstabs and doing an operation (sum) on each position where there are two values. We might create a generic "crosstab reduction helper" class or so, in order to reuse this for other result reducers. But that's not to say you need to do that now - just keep it in mind if we have to reduce other crosstabs.

components/reference-data/src/main/java/org/datacleaner/beans/ReferenceDataMatcherAnalyzerReducer.java
ClaudiaPHI: oki doki.

ClaudiaPHI:
Ok. I will create the Helper so that it is done.

engine/core/src/main/java/org/datacleaner/util/CrosstabReducerHelper.java
```diff
@@ -0,0 +1,103 @@
...
+            final List<CrosstabDimension> dimensions = partialCrosstab.getDimensions();
+            for (CrosstabDimension dimension : dimensions) {
+                final List<String> categories = dimension.getCategories();
+                for (String category : categories) {
+                    final Number categoryValue = nav.where(dimension, category).safeGet(null);
+                    final CrosstabNavigator<Number> whereToPut = mainNavigator.where(dimension, category);
+                    if (categoryValue != null) {
+                        final Number oldValue = whereToPut.safeGet(null);
+                        if (oldValue != null) {
+                            final Number newValue = new BigDecimal(categoryValue.longValue()).add(new BigDecimal(
```
kaspersorensen: Now it seems we always get BigDecimal, where we before always got double?
My request was more to try and retain the original type - at least in the most usual scenarios where it is an Integer or Long. Switching to a decimal type has quite a semantic impact.

ClaudiaPHI: We never had always double. I am the one who  wrote the method and I chose double because I was thinking int is not enough and then you said you want to think about long as well.You said to search for a helper class that would add these numbers. The one that could add long numbers is BigDecimals.It returns a Number and then one could take which type prefers(int, float, double, long).  I will do the classical the adding by looking at the type then.

LosD:
How big is the chance that adding all these numbers would overflow? I actually liked the BigDecimal approach, as there is no such risk.

kaspersorensen:
I was mostly concerned with not adding together e.g. two integers, and getting back a decimal type... For me it is ok to return only BigDecimal and BigInteger, but I think it's semantically quite different to use integers (byte, short, int, long) or decimals (float, double).

kaspersorensen:
I don't think there's much chance of overflow, but hey I'm OK with preventing it anyway :-)

ClaudiaPHI:
Is it mergeable or do I try to add the overflow improvements ?

kaspersorensen:
Could we just fix it with something simple a la:
```
private Number sum(Number n1, Number n2) {
  if (isIntegerType(n1) && isIntegerType(n2)) {
    return new BigInteger...;
  }
  return new BigDecimal...
}

private boolean isIntegerType(Number n) {
  return (n instanceof Byte || n instanceof Short || n instanceof Integer || n instanceof Long ||  n instanceof BigInteger);
}
```

ClaudiaPHI:
I hope/think it is fixed.

engine/core/src/main/java/org/datacleaner/util/CrosstabReducerHelper.java
```diff
@@ -0,0 +1,114 @@
...
+                }
+            }
+        }
+    }
+
+    private static Number sum(Number n1, Number n2) {
+        if (isIntegerType(n1) && isIntegerType(n2)) {
+            return BigInteger.valueOf(n1.longValue()).add(BigInteger.valueOf(n2.longValue()));
+        } 
+        return new BigDecimal(n1.doubleValue()).add(new BigDecimal(n2.doubleValue()));
```
kaspersorensen: actually there's still a risk of precision loss here ... what if n1 or n2 is itself a BigInteger or BigDecimal - then I guess the .longValue and .doubleValue calls will loose precision or could cause overflow. Yuck. Smells a bit like gluing code that someone else has probably done a bunch of times before.

I can live with it like it is, because I think the risk is quite small, but ideally we should create or find a utility class that does this with more guarantees that we are not loosing precision etc.

ClaudiaPHI:
Ok. I will merge it.

Issue #692, opened 2015-09-18
Review component categories


kaspersorensen:
The following new library structure is suggested:

![image](https://cloud.githubusercontent.com/assets/291450/9959886/82b8afc2-5e17-11e5-93b5-3e2687e2c776.png)

![image](https://cloud.githubusercontent.com/assets/291450/9959896/92f7ced6-5e17-11e5-98aa-e74221edd207.png)

LosD:
We decided to rename _Standardization_ category to _Location_, and move **Address correction** in there.

LosD:
@kaspersorensen Currently DC disregards categories with a single component in them, which will place the component in the supercategory. We now created a few categories that will have only component (EAN number in Product data and Country Standardizer in the community edition). Do we want to continue moving those to the super category?

kaspersorensen:
Good question ... We should probably abolish the logic to remove "single component groups". It seems too weird that one moment you have a category and another you don't.

kaspersorensen:
Hi @LosD - one little thing I am missing is the ```ProductDataCategory``` within DataCleaner itself. I think we should put the category class in DC so that multiple extensions can use it. For instance, I just got hinted of a nice idea for an extension and would like to use the category.

LosD:
Seems like a good idea. I'll move it (after our sprint review).

LosD:
@kaspersorensen Would is also be relevant to move the national identifiers category into main DC?

kaspersorensen:
Sure, why noy

kaspersorensen:
Why not that is :-)

PR #694, opened 2015-09-22
Rearrange categories, hide deprecated components

LosD:
This changes the component categories, as the current categories has
proved confusing. This will also hide (but, for compatibility with old
jobs, not disable) components that isn't very useful.

Partial fix for #692


LosD:
Travis build for the PR is unrelated (another 10-minute timeout issue from outputsstream test)

kaspersorensen:
Looks good to me.

PR #695, opened 2015-09-22
Feature/664 show melon columns

### LINKED ISSUE #664: Show information about OutputDataStreams in configuration dialog
Currently the user has no way to learn about the columns of an output data stream, except if he immediately starts using it. This is a bit sparse informatoin - and I would compare my expectations to those of a transformers where there is a panel in the configuration dialog containing the output columns.

So: If a component has output data streams (consumed or not), they should show up in a panel in the configuration dialog of that component. For each stream we should list the columns and their types (but not allow renaming them like we do for transformed columns - that's not possible currently).

tomaszguzialek:
There is one issue I am struggling with. When a component is connected via "Link to...", the "Output data streams" appears as the first task pane. Every change in the input columns forces a refresh and a task pane is added at the end. I would like to have it always at the end, but cannot find an easy way to accomplish that.

Screenshot of the overall design
melons task pane
:

Fixes #664

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/panels/ColumnListTable.java
+++ b/desktop/api/src/main/java/org/datacleaner/panels/ColumnListTable.java
@@ -212,7 +212,7 @@ public final class ColumnListTable extends DCPanel {
         }
         _columnTable.setModel(model);
 
-        if ((hasPhysicalColumns) && (_editable)) {
+        if (hasPhysicalColumns && _editable) {
             final TableColumnExt columnExt = _columnTable.getColumnExt(2);
             columnExt.setMinWidth(26);
             columnExt.setMaxWidth(80);
--- a/desktop/api/src/main/java/org/datacleaner/panels/ColumnListTable.java
+++ b/desktop/api/src/main/java/org/datacleaner/panels/ColumnListTable.java
@@ -77,7 +77,7 @@ public final class ColumnListTable extends DCPanel {
     private final WindowContext _windowContext;
     private final boolean _addShadowBorder;
 
-    private boolean _editable;
+    private final boolean _editable;
 
     public ColumnListTable(Collection<? extends InputColumn<?>> columns, AnalysisJobBuilder analysisJobBuilder,
             boolean addShadowBorder, WindowContext windowContext) {
@@ -176,7 +176,7 @@ public final class ColumnListTable extends DCPanel {
         }
 
         final String[] headers;
-        if ((hasPhysicalColumns) && (_editable)) {
+        if (hasPhysicalColumns && _editable) {
             headers = HEADERS_WITH_ACTION_COLUMN;
         } else {
             headers = HEADERS_WITHOUT_ACTIONS;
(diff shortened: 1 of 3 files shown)


kaspersorensen:
Hmm that issue sounds pretty severe so I suggest we find a solution for it. How about always having the task pane, but just hiding/showing it when appropriate? That way we don't add/remove task panes all the time.

Otherwise it looks quite good. Remark on the looks of it: The icon for each stream (next to "Complete rows" and "Incomplete rows" in the screenshot) should be the icon found with IconUtils.OUTPUT_DATA_STREAM_PATH.

tomaszguzialek:
Always having a task pane does not solve the problem. I added "Output data streams" task pane in AbstractComponentBuilderPanel as all types of components can produce them, not only analyzers. The always-visible task pane is added in the constructor, so it is always the first task pane being added (so on top). I could try to make every class extending the abstract one reorder the task panes, but that does not sound clean...

tomaszguzialek:
Problem solved, @LosD gave me a precious hint. Moved adding of the task pane from the constructor to the end of "init" method, which I have not seen for some strange reason... Icon also fixed, so it is mergeable :).

melons task pane new icons

desktop/api/src/main/java/org/datacleaner/panels/ColumnListTable.java
```diff
@@ -77,22 +77,35 @@
     private final WindowContext _windowContext;
     private final boolean _addShadowBorder;
 
+    private boolean _editable;
```
LosD: This look like a candidate for final

desktop/api/src/main/java/org/datacleaner/panels/OutputDataStreamsViewer.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public void refresh() {
+        removeAll();
+        
+        setEnabled(false);
+        for (OutputDataStream outputDataStream : _componentBuilder.getOutputDataStreams()) {
+            setEnabled(true);
+            
+            List<InputColumn<?>> inputColumns = new ArrayList<>();
```
LosD: Final

desktop/api/src/main/java/org/datacleaner/panels/OutputDataStreamsViewer.java
```diff
@@ -0,0 +1,80 @@
...
+            for (Column column : outputDataStream.getTable().getColumns()) {
+                MetaModelInputColumn inputColumn = new MetaModelInputColumn(column);
```
LosD: Final... And no reason for the intermediate variable at all, really. No debugging advantage, and it doesn't help the source documenting itself, but if you prefer it like this, keep it (I hate unnecessary intermediate variables in all their forms, but @kaspersorensen loves 'em :))

desktop/api/src/main/java/org/datacleaner/panels/ColumnListTable.java
```diff
@@ -197,7 +212,7 @@ public void actionPerformed(ActionEvent e) {
         }
         _columnTable.setModel(model);
 
-        if (hasPhysicalColumns) {
+        if ((hasPhysicalColumns) && (_editable)) {
```
LosD: The parenthesis seems a bit overkill?

LosD:
LGTM except for very minor things.

PR #698, opened 2015-09-24
Applying output data streams of all component types

### LINKED ISSUE #696: JaxbJobWriter only writes output data streams from Analyzers
While working on #620 I noticed that when I save my job files they don't contain the output data stream of a new component I was building for fusing streams.

I dug into the issue a bit and see that this is because we have only implemented output data stream writing for analyzers in JaxbJobWriter (lines 567 - 576).

This _should_ work for all component types since output data streams are not a unique feature of analyzers.

kaspersorensen:
Fixes #696 which caused only analyzers to have their output data streams serialized.


LosD:
LGTM!

PR #700, opened 2015-09-25
Feature/remote components default values

jakubneubauer:
Created tranformer instance can initialize its properties in constructor. These initial values are then visible by user and he regards them as the default values. We want similar behaviour when using the remote components.
So server must provide these values in component info and client should use them when creating new instance of remote transformer.

### DIFF
--- a/engine/remote-components/src/main/java/org/datacleaner/descriptors/RemoteDescriptorProvider.java
+++ b/engine/remote-components/src/main/java/org/datacleaner/descriptors/RemoteDescriptorProvider.java
@@ -25,11 +25,15 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.commons.lang.ClassUtils;
 import org.apache.metamodel.util.LazyRef;
 import org.datacleaner.restclient.ComponentList;
 import org.datacleaner.restclient.ComponentRESTClient;
+import org.datacleaner.restclient.Serializator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.JsonNode;
 
 /**
  * @Since 9/8/15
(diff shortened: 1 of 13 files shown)


engine/remote-components/src/main/java/org/datacleaner/descriptors/RemoteDescriptorProvider.java
```diff
@@ -102,25 +110,27 @@ private void downloadDescriptors() {
...
                                         cl,
                                         propInfo.isRequired(),
                                         transformer,
-                                        initAnnotations(component.getName(), propertyName, propInfo.getAnnotations())));
+                                        initAnnotations(component.getName(), propertyName, propInfo.getAnnotations()), propInfo.getDefaultValue()));
                             } catch(Exception e) {
+                                logger.debug("Cannot initialize typed property descriptor '{}'.'{}'", component.getName(), propertyName, e);
                                 // class not available on this server.
-                                transformer.addPropertyDescriptor(new JsonSchemaConfiguredPropertyDescriptorImpl(
+                                transformer.addPropertyDescriptor(propDesc = new JsonSchemaConfiguredPropertyDescriptorImpl(
```
LosD: Any reason for adding propDesc? It's not used for anything (besides feeding directly into addPropertyDescriptor, of course).

engine/remote-components/src/main/java/org/datacleaner/descriptors/RemoteDescriptorProvider.java
```diff
@@ -151,4 +165,7 @@ private void downloadDescriptors() {
         return annotations;
     }
 
+    public static void main(String[] args) throws Exception {
+    }
+
```
LosD: Errrr? :)

LosD:
Looks good to me, but see comments

jakubneubauer:
The comments were incorporated.

LosD:
Super, let's merge when Travis is happy :)

LosD:
And Travis is happy now. Merging!

PR #702, opened 2015-09-29
Feature/681 pseudo module with standard components

### LINKED ISSUE #681: A pseudo-module for easily depending/including all the standard DC components as one dependency
A wish from @ankit2711 : It would be nice if we offered a single dependency on which embedding applications such as DataHub or List IQ could depend on and get all the standard components in DataCleaner such as Tokenizer, Value distribution, Table lookup and so on.

ClaudiaPHI:
Fixes #681 

- The psedo module contains some dependencies that are common in these subprojects:```engine/xml-config```, ```desktop-ui``` and ```monitor-services```.
- I have tested the build using this pseudo module in engine/xml-config. The build passes. 
- Are there any dependencies that should be added or removed from this pseudo module ?


standard-components/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.eobjects.datacleaner</groupId>
+		<artifactId>DataCleaner</artifactId>
+		<version>4.1-SNAPSHOT</version>
+	</parent>
+	<artifactId>DataCleaner-standard-components</artifactId>
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<project.version>4.1-SNAPSHOT</project.version>
```
kaspersorensen: This isn't needed, actually should not be here because it overrides the real property called "project.version" (which is dynamic - provided by Maven itself).

kaspersorensen:
I would expect also an update to desktop/ui/pom.xml and to monitor/ui/pom.xml ... To remove all the dependencies in there that can be replaced by this new module.

ClaudiaPHI:
Sure thing.

ClaudiaPHI:
Issue addressed.

kaspersorensen:
Looks good to me

PR #705, opened 2015-09-29
Feature/updatable datahub

MennoB:
Update functionality for golden records in DataHub. Ready for integration.


components/writers/src/main/java/org/datacleaner/beans/writers/UpdateDataHubAnalyzer.java
```diff
@@ -0,0 +1,572 @@
...
+        + "\nTo understand the configuration of the Update table component, consider a typical SQL update statement:"
+        + "\n<blockquote>UPDATE table SET name = 'John Doe' WHERE id = 42</blockquote>"
+        + "\nHere we see that there is a condition (WHERE id=42) and a value to update (name should become 'John Doe'). This is what the two inputs are referring to. But obviously you are not dealing with constant values like 'John Doe' or '42'. You have a field in your DC job that you want to map to fields in your database."
+        + "\nUsually the 'condition value' would be a mapping of the key that you have in your job towards the key that is in the database. The 'values to update' property would include the columns that you wish to update based on the values you have in your job.")
+@Categorized(superCategory = WriteSuperCategory.class)
+@Concurrent(true)
+public class UpdateDataHubAnalyzer implements Analyzer<WriteDataResult>, Action<Iterable<Object[]>>, HasLabelAdvice,
```
kaspersorensen: Is this component in any way different from the regular "Update table" component? I don't see anything DataHub-specific about it? Am I missing something?

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+            if (value != null || explicitNulls[i]) {
+                String columnName = columns[i].getName();
+                final UpdateField field = new UpdateField(columnName, value.toString());
+                fields.add(field);
+            }
+        }
+
+        List<FilterItem> whereItems = getWhereItems();
+        if (!(whereItems.size() == 1 && whereItems.get(0).getSelectItem().getColumn().getName().equals("gr_id"))) {
+            throw new IllegalArgumentException("Only gr_id is allowed as a select item");
```
kaspersorensen: Maybe we can phrase this a bit more user-friendly "Updates are only allowed on individual records, identified by gr_id" or something like that. Referring to a select item is at least a bit of an internal detail to the user.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateConnection.java
```diff
@@ -0,0 +1,70 @@
...
+    public final static String UPDATE_PATH = "/goldenrecords/batch";
+
+    DataHubConnection _connection;
+
+    public DataHubUpdateConnection(DataHubConnection connection) {
+        _connection = connection;
+    }
+
+
+  public String getUpdateUrl() {
```
kaspersorensen: Strange indentation going on here and below.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/update/UpdateField.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * POJO for transferring updated records from DataCleaner
+ * to the DataHub update REST service.
+ *
+ */
+public class UpdateField {
+
+    String name;
+    String value;
```
kaspersorensen: add "private final" to these two.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateDataBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class JsonUpdateDataBuilder {
+
+    public static String buildJsonArray(List<UpdateData> updateData) {
+
+        final OutputStream os = new ByteArrayOutputStream();
+        final ObjectMapper mapper = new ObjectMapper();
+        String json = "";
```
kaspersorensen: instead just write
```
final String json;
```

components/writers/src/main/java/org/datacleaner/beans/writers/UpdateDataHubAnalyzer.java
MennoB: This was still the description copied from the update table component. A new, more specific, description has been added.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateBuilder.java
MennoB: Done.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateConnection.java
MennoB: Changed indentation.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/update/UpdateField.java
MennoB: Agreed. Done.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateDataBuilder.java
MennoB: Also done.

components/writers/src/main/java/org/datacleaner/beans/writers/UpdateDataHubAnalyzer.java
```diff
@@ -0,0 +1,570 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Named("Update DataHub")
+@Description("Update Golden Records in the DataHub. This component allows you to update fields defined in the golden records metadata. The component always updates a golden record based"
+        + "\non its golden record id."
+        + "\nThe 'values to update' property lets you specify which values must be changed. This component can not update fields of the metadata, like traffic lights or severities or track codes.")
+@Categorized(superCategory = WriteSuperCategory.class)
+@Concurrent(true)
+public class UpdateDataHubAnalyzer implements Analyzer<WriteDataResult>, Action<Iterable<Object[]>>, HasLabelAdvice,
```
kaspersorensen: But I mean in general ... is there anything different in the code at all? If not, then we could just use the existing UpdateTableAnalyzer?

components/writers/src/main/java/org/datacleaner/beans/writers/UpdateDataHubAnalyzer.java
```diff
@@ -0,0 +1,570 @@
...
+    @Configured(order = 4)
+    @Description("Names of columns in the target table, which form the conditions of the update.")
+    @ColumnProperty
+    @MappedProperty(PROPERTY_NAME_CONDITION_VALUES)
+    String[] conditionColumnNames;
+
+    @Inject
+    @Configured(order = 5)
+    @Description("Datastore to write to")
+    UpdateableDatastore datastore;
```
kaspersorensen: Looking at the type here, it seems to be generically working for any updateable datastore. That's fine, but I would prefer not having two analyzers that does the same thing.

MennoB: You're right. We should be able to make this functionality work with just the generic UpdateTableAnalyzer. We'll give this a shot. However, we also need to make functionality to delete records in the datahub. We'll make a separate analyzer for that.

kaspersorensen: Okay, sounds fine with me, but that could again be a totally general "Delete rows" analyzer also for other datastores. Would be best not to inflate components just because of one new connector (except if the function is really specific to that datastore, like e.g. We have specific search functions for ElasticSearch datastores only).

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateConnection.java
```diff
@@ -0,0 +1,69 @@
...
+import java.net.URISyntaxException;
+
+import org.apache.http.client.utils.URIBuilder;
+import org.datacleaner.util.http.MonitorHttpClient;
+
+public class DataHubUpdateConnection {
+    public final static String CONTEXT_PATH = "/service/cdi/v1";
+    public final static String UPDATE_PATH = "/goldenrecords/batch";
+
+    DataHubConnection _connection;
```
kaspersorensen: private final?

engine/core/src/main/java/org/datacleaner/metamodel/datahub/update/UpdateData.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * POJO for transferring updated records from DataCleaner
+ * to the DataHub update REST service. 
+ *
+ */
+public class UpdateData {
+    
+    private String grId;
+    private UpdateField[] fields;
```
kaspersorensen: final?

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateDataBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class JsonUpdateDataBuilder {
+
+    public static String buildJsonArray(List<UpdateData> updateData) {
+
+        final OutputStream os = new ByteArrayOutputStream();
+        final ObjectMapper mapper = new ObjectMapper();
+        String json;
```
kaspersorensen: final?

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateQueryBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Serializes a batch of update statements into a JSON object that can be sent
+ * to DataCleaner monitor.
+ *
+ */
+@Deprecated
```
kaspersorensen: Adding a new class that is already deprecated?

engine/core/src/main/java/org/datacleaner/util/http/CASMonitorHttpClient.java
```diff
@@ -65,6 +65,10 @@
 
     private static final Logger logger = LoggerFactory.getLogger(CASMonitorHttpClient.class);
 
+    private static final String CDI_TICKET_HEADER = "CDI-ticket";
+    private static final String CDI_SERVICE_URL_HEADER = "CDI-serviceUrl";
+    private static final String CDI_USERID = "CDI-userId";
```
kaspersorensen: This is not proper CAS compliant. This is _HI specific_ CAS code. I don't think that's the proper way to do it.

engine/core/src/test/java/org/datacleaner/metamodel/datahub/DataHubUpdateCallbackTest.java
```diff
@@ -0,0 +1,92 @@
...
+    @InjectMocks
+    private DataHubUpdateCallback callback;
+
+    private UpdateData updateData;
+
+    private UpdateField field1;
+    private UpdateField field2;
+
+    private static int maxBatchSize = DataHubUpdateCallback.INSERT_BATCH_SIZE;
+    private static String query = "query";
```
kaspersorensen: I think these two static fields should also be final.

engine/core/src/test/java/org/datacleaner/metamodel/datahub/DataHubUpdateScriptTest.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.metamodel.datahub;
+
+public class DataHubUpdateScriptTest {
```
kaspersorensen: Empty class?

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateConnection.java
MennoB: Yup, done.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/update/UpdateData.java
MennoB: Done.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateDataBuilder.java
MennoB: Done.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateQueryBuilder.java
MennoB: Ehm, how did that get there? Removed.

engine/core/src/test/java/org/datacleaner/metamodel/datahub/DataHubUpdateScriptTest.java
MennoB: Test class for a no longer existing class. Deleted.

engine/core/src/main/java/org/datacleaner/util/http/CASMonitorHttpClient.java
MennoB: You're right. Unfortunately that means that making something generically CAS-compliant is not feasible at the moment. CASMonitorHttpClient should become DataHubMonitorHttpClient.

kaspersorensen:
Oh but the old CASMonitorHttpClient _was_ CAS compliant. When it was built it was tested in a setup without DataHub, just with DC-monitor behind a custom CAS server!

kaspersorensen:
That was the reason for the more chatty protocol that was there before. Because the ticket was created for the request by the CAS server.

drexler42:
We need to have 2 classes then. A basic one with CAS functionality and a special one for DataHub. I refactored the code to have an abstract class and 2 non-abstract  implementations.

kaspersorensen:
This works but maybe we could even avoid the abstract and just let DataHubMonitorHttpClient extend CASMonitorHttpClient. That way we don't break API for anyone who would be instantiating CASMonitorHttpClient in extensions or so.

I was also wondering a bit regarding the change from private to protected fields. I usually prefer to keep them private and just add getters if they could be relevant to subclasses.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubRepoConnection.java
```diff
@@ -0,0 +1,89 @@
...
+import org.datacleaner.util.http.MonitorHttpClient;
+
+public class DataHubRepoConnection  {
+    public final static String DATASTORES_PATH = "/datastores";
+    public final static String CONTEXT_PATH = "/ui";
+    public final static String REPOSITORY_PATH = "/repository";
+    public final static String SCHEMA_EXTENSION = ".schemas";
+    public final static String QUERY_EXTENSION = ".query?";
+ 
+    DataHubConnection _connection;
```
kaspersorensen: private final?

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateCallback.java
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.metamodel.schema.Schema;
+import org.apache.metamodel.schema.Table;
+import org.apache.metamodel.update.RowUpdationBuilder;
+import org.datacleaner.metamodel.datahub.update.UpdateData;
+
+public class DataHubUpdateCallback extends AbstractUpdateCallback implements UpdateCallback, Closeable {
+
+    private final DataHubDataContext _dataContext;
+    //private DataHubConnection _connection;
+    public static final int INSERT_BATCH_SIZE = 100;
```
kaspersorensen: normally static stuff should be above non-static stuff

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateQueryBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+ * to DataCleaner monitor.
+ *
+ */
+public class JsonUpdateQueryBuilder {
+
+    public static String buildJsonArray(List<String> queries) {
+
+        final OutputStream os = new ByteArrayOutputStream();
+        final ObjectMapper mapper = new ObjectMapper();
+        String json = "";
```
kaspersorensen: final and no assignment?

engine/core/src/test/java/org/datacleaner/util/http/CasMonitorHttpClientTest.java
```diff
@@ -30,7 +30,7 @@
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.impl.client.HttpClients;
 
-public class CASMonitorHttpClientTest {
+public class CasMonitorHttpClientTest {
```
kaspersorensen: Since the name is still CASMonitorHttpClient I think we should also keep that in the test class name.

kaspersorensen:
Only cosmetic remarks left :-) LGTM other than that.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateCallback.java
drexler42: OK.

drexler42:
I'll wait for Travis, then I'll merge. OK?

engine/core/src/test/java/org/datacleaner/util/http/CasMonitorHttpClientTest.java
drexler42: How do you spot all of that.... Anyways, I renamed the class.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/utils/JsonUpdateQueryBuilder.java
drexler42: final and no assignment.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubRepoConnection.java
drexler42: private final.

kaspersorensen:
yup!

Issue #659, opened 2015-09-11
Remote transformers - Hide / publish transformers, configuration of default properties.


khouzvicka:
I want to add settings for the remote transformers to the configuration file.
Admin can specify the list of published (included) remote transformers by their names. The rest of the transformers will be hidden.
Or he can specify the hidden list (excluded) and the rest will be allowed.
Only one element can be here, includes or excludes.
In the second part of the configuration there will be default property values for each remote transformer.

New part of configuration file:

``` html
<published-components>
    <includes>
        <include name="Address Correction"/>
    </includes>

    <excludes>
       <exclude name="Address Correction"/>
    </excludes>

    <defaults>
        <component name="Address Correction">
            <properties>
                <property name="Connection">ilona://localhost:4711/TlNMb2NhdG9y/</property>
            </properties>
        </component>
    </defaults>
</published-components>
```

kaspersorensen:
Suggestion from my side on the design:

 * Make an interface in DC monitor-API which will be called something like "RemoteComponentsConfiguration". Let this interface have a method that will provide the list of components to be exposed and a method to apply default properties to any given component.
 * In DC community edition we could have a default implementation that just exposes all components and does not apply any default settings. This implementation will be plugged as a spring bean via application-context.xml.
 * In DC commercial editions we could provide another implementation that reads a configuration file like what you just put here. This implementation will override the default implementation in a spring file just like we override e.g. the Repository implementation.

khouzvicka:
Hi Kasper. I am implementing this feature and I have question.
I want to add the configuration to conf.xml. Each tenant has one specific configuration (conf.xml).
Is there any global configuration file for this function? Or can this be tenant dependent?

kaspersorensen:
Hi Karel,

I don't think this is good to keep in conf.xml actually, because it's quite specific to the monitor configuration whereas conf.xml is used in so many places where this stuff would not make sense.

To begin with I would suggest to add this in our spring xml files. In the community/open codebase we would just have a dummy bean. In commercial edition we have ways to override this default/dummy bean with something smarter. And to provide input to it via ```datacleaner-monitor.properties```. That would be my preferred way of configuring it.

jakubneubauer:
[...]

PR #712, opened 2015-09-30
Feature/659 remote tramsformer publishing

khouzvicka:
Implementation of #659


kaspersorensen:
According to #659 we decided not to put this in the open core of DC but to only make an interface here with a dummy/simplistic implementation that we replace in Enterprise Edition.

khouzvicka:
OK.
I will make simpler implementation to this project, without any logic.
And all logic will be in other place.
Where is the right place? Some other github place or HumanInference SVN?

kaspersorensen:
That would be the DC EE repo on github. I'll send over the details via mail to keep this forum free of that.

PR #714, opened 2015-09-30
Feature/713 performance optimization

### LINKED ISSUE #713: Performance optimizations (wrt remote transformers)
We run the DC monitor and bombarded it with requests to run String concatenator on 1 row. The requests were sent in 15 threads. During this tests we monitored DC monitor with JProfiler.
Within this issue we would like to discuss the observations and possible solutions in a related branch/pull request.

jakubneubauer:
Fixes #713

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/descriptors/ConfiguredPropertyDescriptorImpl.java
+++ b/engine/core/src/main/java/org/datacleaner/descriptors/ConfiguredPropertyDescriptorImpl.java
@@ -35,22 +35,21 @@ import org.datacleaner.util.StringUtils;
 final class ConfiguredPropertyDescriptorImpl extends AbstractPropertyDescriptor implements ConfiguredPropertyDescriptor {
 
     private static final long serialVersionUID = 1L;
+    private final String name;
+    private final boolean isInputColumn;
+    private String[] aliases;
 
     protected ConfiguredPropertyDescriptorImpl(Field field, ComponentDescriptor<?> componentDescriptor)
             throws DescriptorException {
         super(field, componentDescriptor);
+        name = resolveName();
+        isInputColumn = resolveIsInputColumn();
+        aliases = resolveAliases();
     }
 
     @Override
     public String getName() {
-        Configured configured = getAnnotation(Configured.class);
-        if (configured != null) {
-            String value = configured.value();
-            if (!StringUtils.isNullOrEmpty(value)) {
-                return value.trim();
-            }
-        }
-        return ReflectionUtils.explodeCamelCase(super.getName(), true);
+        return name;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


engine/core/src/main/java/org/datacleaner/descriptors/ConfiguredPropertyDescriptorImpl.java
```diff
@@ -35,22 +35,21 @@
 final class ConfiguredPropertyDescriptorImpl extends AbstractPropertyDescriptor implements ConfiguredPropertyDescriptor {
 
     private static final long serialVersionUID = 1L;
+    private final String name;
+    private final boolean isInputColumn;
```
kaspersorensen: My feeling is that it would be better to cache the baseType.

monitor/api/src/main/java/org/datacleaner/monitor/configuration/TenantContext.java
```diff
@@ -162,4 +163,6 @@
      * @return
      */
     public ComponentStore getComponentStore();
+
+    public StringConverter getStringConverter();
```
kaspersorensen: This is not a nice thing to add to this interface. It's a very low-level thing to request from a very high-level interface.

kaspersorensen:
Hmm I'm not really a big fan of this PR. It adds some sporadic caching here and there but I don't think it really solves the right issues. At least give me a little time to check it out in more detail because I have a feeling we are fighting symptoms here.

kaspersorensen:
I would rather consider to optimize ```SimpleComponentDescriptor.getConfiguredProperty(String)``` because it has an inefficient way of resolving the property (iterates over all properties - would be smarter to have a (transient - it is serializable) hashmap/cache backing it).

The caching of the String converter is a good idea but only for each batch of records IMO. I wouldn't want it to be long-living since some injections that it can make could be dynamically changing over time.

jakubneubauer:
>> It adds some sporadic caching here and there but I don't think it really solves the right issues. At least give me a little time to check it out in more detail because I have a feeling we are fighting symptoms here.

Yes, I focused on the REST API, no other parts. And the proposed changes lowered the overhead time for the stateless service cca from 25 to 4%. That is not bad I guess. And I believe that it helps also to batching (it must also deserialize the row data), I just didn't measure it...
I'm much more concerned about the overhead of the spring framework. When 90% time is spent there, there is not much to do for better performance in our part... Do you have any idea here?

monitor/api/src/main/java/org/datacleaner/monitor/configuration/TenantContext.java
jakubneubauer: I also feel not comfortable with this part. Maybe you can propose another way how to make the StringConverter more performant? Especially the deserialization. I feel my lack of a global overview of DataCleaner parts.

kaspersorensen:
Please see #719

kaspersorensen:
Superceeded by #719

PR #715, opened 2015-09-30
added the missing components dependencies to pseudo module standard components

ClaudiaPHI:
added the missing components dependencies and removed the unnecessary
ones from desktop ui

### DIFF
--- a/desktop/ui/pom.xml
+++ b/desktop/ui/pom.xml
@@ -59,39 +59,12 @@
 		<!-- DataCleaner core / environment -->
 		<dependency>
 			<groupId>org.eobjects.datacleaner</groupId>
-			<artifactId>DataCleaner-desktop-api</artifactId>
-			<version>${project.version}</version>
-		</dependency>
-		<dependency>
-			<groupId>org.eobjects.datacleaner</groupId>
 			<artifactId>DataCleaner-xml-config</artifactId>
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.eobjects.datacleaner</groupId>
 			<artifactId>DataCleaner-env-berkeleydb</artifactId>
-			<version>${project.version}</version>
-		</dependency>
-
-		<!-- DataCleaner components -->
-		<dependency>
-			<groupId>org.eobjects.datacleaner</groupId>
-			<artifactId>DataCleaner-standard-components</artifactId>
-			<version>${project.version}</version>
-		</dependency>
-		<dependency>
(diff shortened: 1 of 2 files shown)


desktop/ui/pom.xml
```diff
@@ -72,28 +67,6 @@
 			<artifactId>DataCleaner-env-berkeleydb</artifactId>
 			<version>${project.version}</version>
 		</dependency>
-
-		<!-- DataCleaner components -->
-		<dependency>
-			<groupId>org.eobjects.datacleaner</groupId>
-			<artifactId>DataCleaner-standard-components</artifactId>
```
kaspersorensen: removing the standard-components?

standard-components/pom.xml
```diff
@@ -80,27 +90,27 @@
 		</dependency>
 		<dependency>
 			<groupId>org.eobjects.datacleaner</groupId>
-			<artifactId>DataCleaner-json</artifactId>
+			<artifactId>DataCleaner-sample-extension</artifactId>
```
kaspersorensen: We don't want the sample-extension in here. It's only a coding example, not something to actually use!

kaspersorensen:
Hmm the diff is pretty impossible to read because it seems to have shifted a lot of the dependencies down or up. I cannot figure out fully what changed.

desktop/ui/pom.xml
ClaudiaPHI: it is included in xml-config dependency

ClaudiaPHI:
I added the components you asked me about in the email.

kaspersorensen:
I only asked why, didn't say you had to :) Anyways I think this also adds sample-extension? And the diff makes it hard for me to see other effects.

desktop/ui/pom.xml
kaspersorensen: OK but I think it's best to explicitly name it here. The transitive dependency of xml-config is more a coincidence than design, whereas "ui" depending on "api" is certainly a strong contract.

ClaudiaPHI:
I also changed the order in order to match the order in the one in project tree.  If this is not necessary, we can close the branch.

desktop/ui/pom.xml
ClaudiaPHI: Ok. it seems the original thing is better.

ClaudiaPHI:
If you do not mind I will close this pull request without merging.

kaspersorensen:
Ok with me :-)

PR #717, opened 2015-09-30
Categories in main DataCleaner

LosD:
This will move the previous commercial-only categories to community.


kaspersorensen:
LGTM, merge when Travis agrees.

PR #719, opened 2015-09-30
Feature/713 performance optimization alternative

kaspersorensen:
Another shot at solving/improving #713 

I had to do quite some code munging to get into this part of the codebase. As a side note I would like to raise that we should really sync on coding guidelines here, since I see a lot of ideoms that doesn't seem to be shared. Most of it is cosmetic, but I did also find some semantic. Most worrying is the hardcoded tenant name in ```RemoteDescriptorProvider``` - we should really do something about that.

Anyways - back to the real issue here. I saw that a new StringConverter is created for each property to be deserialized. At least we can safely reuse the converter for the complete component. I think this should already give a good improvement.

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
@@ -72,238 +72,238 @@ import com.fasterxml.jackson.databind.node.TextNode;
  */
 public class ComponentHandler {
 
-	private static final Logger LOGGER = LoggerFactory.getLogger(ComponentHandler.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(ComponentHandler.class);
 
-	public static ObjectMapper mapper = Serializator.getJacksonObjectMapper();
+    public static ObjectMapper mapper = Serializator.getJacksonObjectMapper();
 
-	private final String componentName;
-	private final DataCleanerConfiguration dcConfiguration;
+    private final String componentName;
+    private final DataCleanerConfiguration dcConfiguration;
 
-	private ComponentDescriptor<?> descriptor;
-	private Map<String, MutableColumn> columns;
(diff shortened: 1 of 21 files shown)


monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
```diff
@@ -74,43 +76,48 @@
 
     public static ObjectMapper mapper = Serializator.getJacksonObjectMapper();
 
-    private final String componentName;
-    private DataCleanerConfiguration dcConfiguration;
+    private final String _componentName;
+    private final DataCleanerConfiguration _dcConfiguration;
+    private final StringConverter _stringConverter;
```
jakubneubauer: Please note, that simply reusing the StringConverter inside ComponentHandler does not solve problem of performance, if the mehod "deserialize" of StringConverter will be not changed too. Because current performance problem is that the "deserialize" method creates and initializes a DelegateConverter every time again.

kaspersorensen: You're right. Hmm maybe we should internally in StringConverter create a kind of a "base converter delegate" and then only wrap it with further converters when needed (basically when the collection in ```StringConverter.deserialize(String, Class<E> type, Collection<Class<? extends Converter<?>>>)``` is not empty).

kaspersorensen:
OK I've added an attempt to improve the performance on the StringConverter itself.

kaspersorensen:
@jakubneubauer do we agree on the approach here now?

kaspersorensen:
The test failure that seems to have happened on Java 7 is unrelated. Must be a race condition in the assertion.

jakubneubauer:
Yes, I agree. Although I expect some debate with Manuel about the stateless API optimization ;-)

kaspersorensen:
OK, let's have that then :-)

PR #720, opened 2015-10-01
Feature/697 own swagger json provider

jhorcicka:
There is a special controller (in monitor/documentation module) that searches for other controllers (from monitor/services module) and provides the JSON document required by swagger-ui web application. 

The main reason for this was to get rid off undesired dependencies within swagger library.


monitor/documentation/pom.xml
```diff
@@ -31,40 +31,30 @@
...
+            <artifactId>spring-webmvc</artifactId>
+        </dependency>
 
         <!-- Servlet+JSP dependencies -->
-		<dependency>
-			<groupId>javax.servlet</groupId>
-			<artifactId>javax.servlet-api</artifactId>
-			<scope>provided</scope>
-		</dependency>
+        <dependency>
```
kaspersorensen: In XML documents we have tabs instead of spaces (yes - it's an inconsistent rule, but let's keep the diffs consistent until the point where we decide to update the rule once and for all :))

monitor/ui/pom.xml
```diff
@@ -172,6 +172,11 @@
 			<artifactId>DataCleaner-monitor-services</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+        <dependency>
```
kaspersorensen: also funny indentation here.

kaspersorensen:
I couldn't find much else to complain about than indentations ;-)

monitor/documentation/pom.xml
jhorcicka: OK, thanks for mentioning it. It should be fixed now.

kaspersorensen:
Looks good, merging.

Issue #716, opened 2015-09-30
Hidden Property Annotation extension


khouzvicka:
Hidden Property Annotation is now for hiding specific property from component in gui.
After my improvement, annotation  will have two boolean parameters. One for local hiding (same behavior like before) and second for remote hiding. Programmer can specify hiding for local access or remote access or both.

kaspersorensen:
I'm closing this now.

@khouzvicka a little tip: When you post a PR make sure to somewhere in it's description put the text
```
Fixes #716
```
This will ensure that if the PR is merged, then the issue is also closed.

PR #722, opened 2015-10-01
Feature/716 hidden property annotation extension

khouzvicka:
Implementation of #716


kaspersorensen:
LGTM. Let's merge when Travis is OK with it.

PR #730, opened 2015-10-03
Icons for the new component categories

kaspersorensen:
My initial attempt at least ... Feedback very welcome.


kaspersorensen:
One observation: When inspecting the library in the UI I see that CountryStandardizer is placed directly in the "Improve" category, not within the "Location" subcategory. Didn't we disable that feature @LosD (removal of categories with size=1)?

LosD:
Damn. Nope, forgot about that one

LosD:
What I _meant to say_ was "Yes, yes we did" :)

PR #735, opened 2015-10-06
Feature/650 reducer for boolean analyzer

### LINKED ISSUE #650: AnalyzerResultReducer for Boolean analyzer

ClaudiaPHI:
Fixes #650 

- Created an BooleanAnalyzeReducer + unit test 
- Refractored ReferenceDataMatcherAnalyzerReducer (almost identical code)


kaspersorensen:
travis failed

components/basic-analyzers/src/main/java/org/datacleaner/beans/BooleanAnalyzer.java
```diff
@@ -57,6 +57,7 @@
     public static final String MEASURE_ROW_COUNT = "Row count";
     public static final String DIMENSION_COLUMN = "Column";
     public static final String DIMENSION_MEASURE = "Measure";
+    public static final String DIMENSION_COMBINATION = "Combination ";
```
kaspersorensen: maybe call it DIMENSION_COMBINATION_PREFIX because it is more a prefix than an actual dimension name.

components/reference-data/src/main/java/org/datacleaner/beans/ReferenceDataMatcherAnalyzerReducer.java
```diff
@@ -35,37 +30,9 @@ public BooleanAnalyzerResult reduce(Collection<? extends BooleanAnalyzerResult>
...
-            final Crosstab<Number> partialValueCombinationCrosstab = partialResult.getValueCombinationCrosstab();
-            CrosstabReducerHelper.addData(newResultColumnStatistics, partialColumnStatisticsCrosstab);
-            CrosstabReducerHelper.addData(newResultColumnValueCombination, partialValueCombinationCrosstab);
-        }
-
-        return new BooleanAnalyzerResult(newResultColumnStatistics, newResultColumnValueCombination);
+        // the implementation of this reducer would be almost identical with
+        // BooleanAnalyzerReducer. 
+        final BooleanAnalyzerReducer booleanAnalyzerReducer = new BooleanAnalyzerReducer();
+        return booleanAnalyzerReducer.reduce(partialResults);
```
kaspersorensen: maybe delete this class entirely and simply annotate ReferenceDataMatcherAnalyzer so that it uses the BooleanAnalyzerReducer?

kaspersorensen: That would be fine since it also implements Analyzer<BooleanAnalyzerResult>

engine/core/src/main/java/org/datacleaner/util/CrosstabReducerHelper.java
```diff
@@ -1,114 +0,0 @@
...
-
-import org.datacleaner.result.Crosstab;
-import org.datacleaner.result.CrosstabDimension;
-import org.datacleaner.result.CrosstabNavigator;
-
-/**
- * Helper class for reductions of crosstabs
- *
- */
-public class CrosstabReducerHelper {
```
kaspersorensen: Why was this moved from engine-core to basic-analyzers? Right now it seems to be shared between boolean analyzer and reference-data matcher. And potentially any analyzer that uses crosstabs. So shouldn't it be in a common place instead of in basic-analyzers which is a particular set of analyzers?

ClaudiaPHI: because it seemed to me I made it more for boolean analyzers. As far I could see the Boolean Analyzers have only 2 dimensions. I was not sure about other type of crosstabs.

kaspersorensen:
Looks quite alright when I do a quick scan, but something fails in Travis. Seems to only be on Java 7 ... Did you code in Java 8 ? :)

components/basic-analyzers/src/main/java/org/datacleaner/beans/BooleanAnalyzerReducer.java
```diff
@@ -0,0 +1,228 @@
...
+                        columnDimension, measureDimension);
+            }
+            // gather the sum of all possible value combinations found in the
+            // partial crosstabs
+            if (partialValueCombinationCrosstab != null) {
+                addValueCombinationsCrosstabDimension(valueCombinations, partialValueCombinationCrosstab);
+            }
+        }
+        // create a new measure dimension for Value Combination crosstab
+        if (valueCombinations != null && columnValueCombinationCrosstabDimensions != null) {
```
LosD: None of these can ever be null (both are created explicitly with new)

engine/core/src/main/java/org/datacleaner/util/CrosstabReducerHelper.java
```diff
@@ -76,38 +76,69 @@ private static boolean dimensionExits(Collection<CrosstabDimension> list, Crosst
...
+
+    public static Number sum(Number n1, Number n2) {
         if (isIntegerType(n1) && isIntegerType(n2)) {
             return BigInteger.valueOf(n1.longValue()).add(BigInteger.valueOf(n2.longValue()));
-        } 
+        }
         return new BigDecimal(n1.doubleValue()).add(new BigDecimal(n2.doubleValue()));
     }
 
+    public static Number substract(Number n1, Number n2) {
```
LosD: Small typo: It's subtract

LosD:
LGTM!

PR #737, opened 2015-10-06
Issue #736: Fixes unsafe cast in CreateStagingTableAnalyzer

### LINKED ISSUE #736: Unsafe cast in CreateStagingTableAnalyzer
line 107: 
```
final DatastoreCreationDelegate creationDelegate = new DatastoreCreationDelegateImpl(
                (MutableDatastoreCatalog) datastoreCatalog);
```

This is highly unsafe since datastore catalog might be e.g. DatastoreCatalogImpl.

Instead let the DatastoreCreationDelegate class do a check and adapt it's behaviour accordingly.

kaspersorensen:
Fixes #736 - unsafe cast in CreateStagingTableAnalyzer


desktop/ui/src/main/java/org/datacleaner/output/datastore/DatastoreCreationDelegateImpl.java
```diff
@@ -29,23 +30,27 @@
...
+    public void createDatastore(Datastore datastore) {
+        final String name = datastore.getName();
+        synchronized (_datastoreCatalog) {
+            if (_datastoreCatalog.containsDatastore(name)) {
+                logger.warn("Datastore '{}' already exists. No new datastore will be created!", name);
+            } else {
+                if (_datastoreCatalog instanceof MutableDatastoreCatalog) {
+                    ((MutableDatastoreCatalog) _datastoreCatalog).addDatastore(datastore);
+                } else {
+                    logger.error("Tried to create datastore '{}', but the datastore catalog is not mutable", name);
```
LosD: To the user, nothing really happens, as far as I can see. Shouldn't this throw an exception so the user can get a warning?

ClaudiaPHI:
LGTM!

desktop/ui/src/main/java/org/datacleaner/output/datastore/DatastoreCreationDelegateImpl.java
kaspersorensen: Yes ... It should never happen actually. So better throw exception.

kaspersorensen:
Let's merge then when Travis is done with it.

LosD:
It is now. Merging!

PR #740, opened 2015-10-07
Feature/658 hdfs file browser

### LINKED ISSUE #658: HDFS file browser.
Choosing HDFS files is a little cumbersome. A dialog where you can choose the file would be very nice.

A minor UI complication here may be that sometimes the user may want to choose a folder instead of a file, which the normal "file chooser" paradigm doesn't really combine, it's usually opened either to choose a file or folder, and the UI reflects that.

tomaszguzialek:
The HDFS file chooser can only select files, not directories. However, I guess it is good enough for now.
hdfs_browser_1
hdfs_browser_2

Fixes #658


desktop/api/src/main/java/org/datacleaner/util/WidgetFactory.java
```diff
@@ -323,6 +333,41 @@ public static JXTextField createTextField(String promptText, int columns) {
         return tf;
     }
 
+    public static JXFormattedTextField createFormattedTextField(String promptText, final int columns, final Format format) {
+        JXFormattedTextField tf = new JXFormattedTextField(promptText);
+        // Stupid JXFormattedTextField will not pass along a formatter to the constructor.
```
kaspersorensen: Maybe we should consider creating a separate class (DCFormattedTextField) for it then? Not a blocker for me, but maybe to consider if we touch the code anyway.

desktop/ui/src/main/java/org/datacleaner/widgets/HdfsUrlTextFieldImpl.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package org.datacleaner.widgets;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+import org.apache.metamodel.util.HdfsResource;
+import org.datacleaner.windows.HdfsUrlChooser;
+
+public class HdfsUrlTextFieldImpl extends AbstractFilenameTextField<HdfsResource> {
```
kaspersorensen: Naming-wise I think this is a bit busy and both adds prefix + suffix to the abstract classname. Impl isn't needed when also a prefix is there IMO. And it replaces "filename" with "url" which seems to violate inheritance (to inherit, you should be able to say that "url is a filename" which I don't think is the case).

I would normally name this: HdfsFilenameTextField.

desktop/ui/src/main/java/org/datacleaner/widgets/SelectionListener.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.widgets;
+
+public interface SelectionListener<N> {
```
kaspersorensen: This sounds very very generic, yet it is bound to FilenameTextField. So it gives me a wrong impression IMO.

desktop/ui/src/main/java/org/datacleaner/windows/ServerAddressDialog.java
```diff
@@ -0,0 +1,144 @@
...
+import org.datacleaner.util.NumberDocument;
+import org.datacleaner.util.WidgetFactory;
+import org.datacleaner.util.WidgetUtils;
+import org.datacleaner.widgets.Alignment;
+import org.datacleaner.widgets.DCLabel;
+import org.jdesktop.swingx.JXFormattedTextField;
+import org.jdesktop.swingx.JXTextField;
+import org.jdesktop.swingx.VerticalLayout;
+
+public class ServerAddressDialog extends JComponent {
```
kaspersorensen: Again, this gives me the impression of something very generic, yet it is HDFS/Hadoop specific. So I would certainly have that in the name.

desktop/ui/src/main/java/org/datacleaner/widgets/SelectionListener.java
```diff
@@ -0,0 +1,24 @@
...
+    void onSelected(FilenameTextField filenameTextField, N selected);
```
kaspersorensen: Looking at what this listener interface does ... Would it make sense to instead just (re)use ```ResourceTypePresenter.Listener``` instead?

tomaszguzialek:
@LosD, can you have a look at the review comments as you know this part better than me?

desktop/api/src/main/java/org/datacleaner/util/WidgetFactory.java
LosD: I don't think so. It is nicely contained within the WidgetFactory, so the pain has been taken care of the the end user.

desktop/ui/src/main/java/org/datacleaner/widgets/HdfsUrlTextFieldImpl.java
LosD: I'd rather rename AbstractFilenameTextField, then. It is not a filename.

desktop/ui/src/main/java/org/datacleaner/windows/ServerAddressDialog.java
LosD: That is actually on purpose. Everything HDFS-specific in contained within showHdfsNameNodeDialog(). I imagined it to be used for any kind of server selection.

However, it is not really a complete idea. We may want to do it by passing classes that does the server validation to avoid it depending on the actual service.

desktop/ui/src/main/java/org/datacleaner/widgets/SelectionListener.java
LosD: Agreed.

desktop/ui/src/main/java/org/datacleaner/widgets/SelectionListener.java
LosD: I'm not actually sure. It may very well be.

desktop/api/src/main/java/org/datacleaner/util/WidgetFactory.java
kaspersorensen: OK.

desktop/ui/src/main/java/org/datacleaner/windows/ServerAddressDialog.java
kaspersorensen: Yeah, I mean the class imports hadoop things and is concrete/non-abstract ... In it's current form I would qualify it's name a bit or else I would get confused. Also since it is in a rather large java package (org.datacleaner.windows) already.

desktop/ui/src/main/java/org/datacleaner/widgets/SelectionListener.java
tomaszguzialek: ```ResourceTypePresenter.Listener``` has a little different signature:
```
public void onResourceSelected(ResourceTypePresenter<?> presenter, Resource resource)
```
Also, it has one more method to implement. It would be a little annoying to use it IMHO.

LosD: Oh yeah. I actually used this before. ResourceTypePresenter.Listener is the listener interface where you have two different methods. One called if there is a valid resource, and one called if there is not. It's... Not pretty.

kaspersorensen: OK then it leaves just that I think the naming here is overly generic. SelectionListener can be any "selection" but the method signature tells that it at least something to do with a FilenameTextField.

tomaszguzialek: Maybe we should get rid of that instead of renaming it. FileSelectionListener is the only interface that implements SelectionListener<N>. It is exactly how it was before this PR...

desktop/ui/src/main/java/org/datacleaner/widgets/HdfsUrlTextFieldImpl.java
tomaszguzialek: My proposal: AbstractFileTextField with HdfsUrlTextField and FilenameTextField as subclasses.

LosD: Hmmm... Hmmmm... I still don't really agree. It's still not really a file (at least not in the Java sense of the word).We could call it AbstractResourceTextField, with HdfsResourceTextField, and FileResourceTextField.

... Of course, that will mean that we lose some of the actual meaning (it can no longer be inferred that it is the actual Url and name that it is being used for). HdfsUrlResourceTextField and FilenameResourceTextField, maybe?

LosD: OTOH, they are not resources, just aliases for resources...

tomaszguzialek:
Arrgh, my commit killed the discussion about naming of AbstractFilenameTextFieldClass. Agreed that AbstractFileTextField might suggest it is related to java.io.File, but on the other hand a file is also a generic thing notion. However, I am fine with renaming it to AbstractFileResourceFieldClass. URL is after all a Uniform RESOURCE Locator :). I will go ahead with such naming.

LosD:
:+1:

LosD:
No wait. AbstractFileResource makes it sound like it's referring to FileResource, a very concrete class.

tomaszguzialek:
Sorry, I meant AbstractResourceTextField. I used this name also in the commit.

kaspersorensen:
Just be aware that I know a few extensions that use the class FilenameTextField. So they should be able to keep using that one. Of course renaming it's parent class should be doable without impacting backwards compatibility.

LosD:
LGTM, but it is probably @kaspersorensen that should do the final check.

desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleFilePropertyWidget.java
```diff
@@ -94,9 +95,9 @@ public SingleFilePropertyWidget(ConfiguredPropertyDescriptor propertyDescriptor,
             _filenameField.setFile(currentValue);
         }
 
-        _filenameField.addFileSelectionListener(new FileSelectionListener() {
+        _filenameField.addSelectionListener(new FileSelectionListener() {
```
kaspersorensen: This method name change will break compatibility I guess. And now that the interface was reverted to name "FileSelectionListener" then I guess also the old method name makes sense.

kaspersorensen:
And here's 31 new warnings:

image

kaspersorensen:
The white area in the dialog makes it look pretty bare.

If I compare with the normal file chooser, then it has a silver border all the way around, and an "Open" and "Cancel" button in the bottom. I think we should add the same here to make it look a bit more integrated.

Adding screenshot of normal file chooser for reference:

image

kaspersorensen:
One the remarks above are handled, then I am good with merging this. Thanks for the patience :)

kaspersorensen:
Another strange thing. After selecting the HDFS file it doesn't seem that a datastore is registered. When I go back to Datastore Mgmt, there are no new entries at least.

tomaszguzialek:
Fixed the warnings and the datastore is now registered. As a bonus, the type of files in HDFS can be now automatically inferred.

kaspersorensen:
image

tomaszguzialek:
LGTM. Merge after Travis' confirmation.

PR #749, opened 2015-10-08
Composite datastore scrollbar

### LINKED ISSUE #733: Missing scrollbar in CompositeDatastoreDialog
If you have a lot of datastores and try to make a composite dialog, then the vertical size (height) of the dialog might be too big for your screen. And the dialog has no scrollbars, so there is no way to go to the bottom of the list.

tomaszguzialek:
Fixes #733

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/CompositeDatastoreDialog.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/CompositeDatastoreDialog.java
@@ -20,6 +20,7 @@
 package org.datacleaner.windows;
 
 import java.awt.BorderLayout;
+import java.awt.Dimension;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
@@ -28,6 +29,7 @@ import java.util.Set;
 import javax.swing.JCheckBox;
 import javax.swing.JComponent;
 import javax.swing.JLabel;
+import javax.swing.JScrollPane;
 import javax.swing.event.DocumentEvent;
 
 import org.datacleaner.bootstrap.WindowContext;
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/windows/CompositeDatastoreDialog.java
```diff
@@ -173,7 +175,11 @@ protected JComponent getDialogContent() {
             checkBoxPanel.add(checkBox);
         }
 
-        WidgetUtils.addToGridBag(checkBoxPanel, formPanel, 0, 1, 2, 1);
+        JScrollPane checkBoxScrollPane = WidgetUtils.scrolleable(checkBoxPanel);
```
kaspersorensen: final

desktop/ui/src/main/java/org/datacleaner/windows/CompositeDatastoreDialog.java
```diff
@@ -173,7 +175,11 @@ protected JComponent getDialogContent() {
...
+        checkBoxScrollPane.setPreferredSize(new Dimension(checkBoxPanel.getWidth(), 300));
+        checkBoxScrollPane.setVerticalScrollBarPolicy(JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED);
+        checkBoxScrollPane.setBackground(_outerPanel.getBackground());
```
kaspersorensen: Is this really needed? Normally a scrollpane does not fill the content, it only adds the scroll bars on top.

tomaszguzialek: Without setting the background color, the panel became white...

kaspersorensen:
OK. Will merge then :-)

PR #751, opened 2015-10-08
Column name always showing as "none" in Ref. Integrity analyzer dialog

### LINKED ISSUE #747: Referential Integrity check "Column name" property widget always initialized with "none"
If you configure the _Referential Integrity_ analyzer in DC desktop and then close the dialog, then the property value of the "Column name" property is properly saved.

BUT if you reopen the dialog it shows up with a value of "none".

kaspersorensen:
Fixes #747 

This was basically an issue of making sure that the combobox would accept a column name before it had been instrumented with a table. The order of instrumenting it with the table is out of the property widget's hands.

I also added a fireValueChanged(...) in the end to ensure that UI is updated if the setValue(...) call is coming from the back instead of the front end.


ClaudiaPHI:
LGTM!

PR #753, opened 2015-10-08
Feature/659 remote tramsformer publishing

khouzvicka:
Simple community implementation for  #659


engine/rest-client/src/main/java/org/datacleaner/restclient/ComponentController.java
```diff
@@ -75,7 +75,7 @@ public String createComponent(final String tenant, final String name, final Stri
      * @return
      */
     public ProcessOutput processComponent(final String tenant, final String id, final ProcessInput processInput)
-            throws ComponentNotFoundException;
+            throws RuntimeException;
```
LosD: Not really a problem, but keeping ```throws RuntimeException``` instead of just deleting the throws declaration seems a bit unnecessary: ```throws ComponentNotFoundException``` didn't have any real effect, but it could help for documentation purposes. ```throws RuntimeException``` doesn't have any effect, and doesn't document anything.

jakubneubauer:
will create new PR to merge to release/5.0 branch

PR #754, opened 2015-10-08
check if there are sample rows for "null" values of value distribution.

### LINKED ISSUE #743: Address correction wizard's drill-to-details panel is empty
The value distribution step: 

The panel that was supposed to show the records that are null is empty: 

screen shot 2015-10-07 at 2 41 31 pm

P.S. If you find a better title for this issue please update it :-P

ClaudiaPHI:
Fixes #743  
check if there are sample rows for "null" values of value distribution.
The method did not check if there are annotated Rows for null values. 
It returned true and it created an window with no row sample examples.
 
screen shot 2015-10-08 at 1 32 05 pm


kaspersorensen:
Good fix!

PR #759, opened 2015-10-08
Made search work on whole tree panel. Search alias+category.

### LINKED ISSUE #758: Search/filter textbox in 'Library' should be default key listener in job window
The search/filter textbox at the bottom of the 'Library' tree should automagically be engaged when entering text in the job editing UI of DataCleaner. Of course it should not catch key events on other text boxes or so, but when e.g. the graph or the tree is in focus then this search box should be the main spot for it.

kaspersorensen:
Fixes #758 

In addition to making the search box catch keystrokes from the whole tree panel, I also made the searching more broad because I think it was being too strict before. So now the search is ignoring whitespaces and it searches also aliases and category names. Thus if you search on e.g. 'Text' you get:

image

Or searching "refere":

image

I also made it easy to clear the text field by clicking Esc.


tomaszguzialek:
LGTM!

PR #761, opened 2015-10-09
Enable ElasticSearchDatastore SSL without keystore set

### LINKED ISSUE #688: Make it possible to _not_ select keystore for an SSL ElasticSearch datastore
We have made keystore selection mandatory for an ElasticSearch datastore that uses SSL, as the Shield extension by default requires an SSL client certificate. However, it is possible to disable that requirement (see here: https://www.elastic.co/guide/en/shield/current/_using_elasticsearch_java_clients_with_shield.html#disabling-client-auth and here: https://www.elastic.co/guide/en/shield/current/separating-node-client-traffic.html). That means that we should preferably keep the keystore selection optional to support those setups.

LosD:
This will ensure a user can enable SSL without having a keystore set.

Fixes #688

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
+++ b/engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
@@ -33,6 +33,8 @@ import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.transport.InetSocketTransportAddress;
 import org.elasticsearch.node.Node;
 
+import com.google.common.base.Strings;
+
 /**
  * Datastore providing access to an ElasticSearch index.
  */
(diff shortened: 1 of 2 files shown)


engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
```diff
@@ -113,8 +113,10 @@ public PerformanceCharacteristics getPerformanceCharacteristics() {
             if (!StringUtils.isNullOrEmpty(_username) && !StringUtils.isNullOrEmpty(_password)) {
                 settingsBuilder.put("shield.user", _username + ":" + _password);
                 if (_ssl) {
-                    settingsBuilder.put("shield.ssl.keystore.path", _keystorePath);
-                    settingsBuilder.put("shield.ssl.keystore.password", _keystorePassword);
+                    if(_keystorePath != null && !_keystorePath.isEmpty()) {
```
kaspersorensen: Small improvement would be to use ```Strings.isNullOrEmpty(_keystorePath)```.
But not a blocker at all.

kaspersorensen:
LGTM!

engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
LosD: Let's do it!

PR #762, opened 2015-10-09
Bug/732 promoted datastores

### LINKED ISSUE #732: Use database-priority based approach for inclusion/exclusion of database icons in DatastoreManagementPanel
Right now I am missing here the icons for MS SQL server and MySQL. But the available databases is a dynamic thing and we seem to have introduced a static rule about how many databases we list in panel 2.

Here's how it looks today:

image

While I would want the following databases to always be shown if their drives are installed:

 * Hive
 * PostgreSQL
 * Oracle
 * MySQL
 * MS SQL Server

tomaszguzialek:
Fixes #732


kaspersorensen:
LGTM. Advise to merge when Travis is successful

PR #763, opened 2015-10-09
Feature/enum sorting

jakubneubauer:
For remote components: values of enumeration properties shown in combo-box were not sorted. So now either the enum class implements Comparable, or it is sorted by name.


engine/core/src/main/java/org/datacleaner/descriptors/EnumerationValue.java
```diff
@@ -59,13 +59,14 @@ public EnumerationValue(Enum<?> value) {
...
             name = ((HasName) enumValue).getName();
+            if(name == null) { name = ""; }
         } else {
             name = value.toString();
         }
     }
 
     /** Available only if this object represents a Java enum value */
-    public Enum<?> asJavaEnum() {
+    public Enum asJavaEnum() {
```
kaspersorensen: Why degrading to raw type here?

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
```diff
@@ -89,7 +91,6 @@
     public ComponentHandler(DataCleanerConfiguration dcConfiguration, String componentName) {
         _dcConfiguration = dcConfiguration;
         _componentName = componentName;
-        _stringConverter = new StringConverter(dcConfiguration);
```
kaspersorensen: I kinda liked that it was final. What's keeping us going back and forward here?

engine/core/src/main/java/org/datacleaner/descriptors/ClasspathScanDescriptorProvider.java
```diff
@@ -611,7 +611,7 @@ public ClasspathScanDescriptorProvider addAnalyzerClass(Class<? extends Analyzer
             try {
                 descriptor = Descriptors.ofAnalyzer(clazz);
                 _analyzerBeanDescriptors.put(clazz.getName(), descriptor);
-            } catch (Exception e) {
+            } catch (Throwable e) {
```
kaspersorensen: Catching throwable is something to be careful with... Difference from Exception is just errors which you cannot recover from (OOM, StackOverflow etc). Why is it needed?

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
jakubneubauer: using non-final so it can be lazy-created in getter. Avoids creation of the string converter for simple cases, e.g. when property is simple string value.

engine/core/src/main/java/org/datacleaner/descriptors/ClasspathScanDescriptorProvider.java
jakubneubauer: Well, I encountered a case where analyzing a transformer/analyzer class with reflection caused NoClassDefFoundError. It was actually because some .jar files on classpath were in older version because of dependencies conflicts.
Your notice  is generally good, but in this case no recovery is done. DataCleaner just logs a problem that it cannot create descriptor for a component and goes on. Logging also errors is ok in this case in my opinion.

jakubneubauer: will change the catch block to catch specific error class only.

jakubneubauer:
Reviewed with @jhorcicka

PR #764, opened 2015-10-09
Feature/728 service for tenant name

jakubneubauer:
1. New rest service to provide info about logged-in user. For now only a tenant name is in the response.
2. This servise is used in the desktop client to automatically recognize tenant to be used for the remote components api client calls.


kaspersorensen:
LGTM!

PR #765, opened 2015-10-09
Feature/726 remote transformer batching

jakubneubauer:
Using batching transformer as base class for remote transformer. Allows grouping rows into single query. For string concatenator it is cca 2-3x faster than "1 row = 1 query" approach.


monitor/pom.xml
```diff
@@ -132,6 +132,12 @@
 					</exclusion>
 				</exclusions>
 			</dependency>
+			<!-- spring-security-cas imports version 3.1, while hadoop-common imports version 3.2.1. -->
+			<dependency>
+				<groupId>commons-collections</groupId>
+				<artifactId>commons-collections</artifactId>
+				<version>3.2.1</version>
```
kaspersorensen: I believe that we already have commons collections under dependency management, so it's version should not be specified here.

jakubneubauer: No yet. This new addition is in the <dependencyManagement> part of monitor pom.xml

kaspersorensen: What I mean is that commons-collections is already a transitive dependency of both desktop and monitor:

image

So if we start doing management of it, it should be at the parent POM level to avoid anomalies between desktop and monitor.

jakubneubauer: Now I understand your point. Will move it to the parent pom.

jakubneubauer:
obsoleted by PR Bug/828 remote transformers output row collector

PR #766, opened 2015-10-09
Feature/delete source records

HettyK:


engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDeleteBuilder.java
```diff
@@ -25,11 +26,22 @@
 import org.apache.metamodel.query.FilterItem;
 import org.apache.metamodel.schema.Table;
 
-public class DataHubDeleteBuilder extends AbstractRowDeletionBuilder{
+public class DataHubDeleteBuilder extends AbstractRowDeletionBuilder {
+    private static final String GR_ID_COLUMN_NAME = "gr_id";
+    private static final String SOURCE_ID_COLUMN_NAME = "source_id";
+    private static final String SOURCE_NAME_COLUMN_NAME = "source_name";
+
+    enum RecordType {
```
kaspersorensen: Can't we avoid having this somehow? It's one thing that it should eventually go away in DataHub itself, but now it's also leaking to the client. Maybe we can make it so that it's transmitted to the client as a string via some of the services and thus it becomes a dynamic parameter that can eventually also be made dynamic on the server?

HettyK: I guess you mean that we don't want the names of the columns hardcoded in the client? I checked the schema were are fetching from DataHub and that has a primary key property set to true for the gr_id (golden records) and the source_id and source_name (source records). So maybe we can use that to check the where clause. Would that be a good solution?

kaspersorensen: I mean that the enum is a very big restriction around what DataHub can do (only work on person or organization). This restriction is unfortunately there in the backend but shouldn't slip out to the frontend IMO. We can just ask the backend to tell us what record types it has and keep that at the client. But not at compile time. It should be something we keep on request, making it possible for the server to improve and handle other record types.

kaspersorensen:
LGTM

PR #773, opened 2015-10-12
Utility for masking file paths when comparing against the benchmark

### LINKED ISSUE #772: Masking file paths when asserting against benchmarks in unit tests
In unit tests often there are asserts of DataCleaner job XML files against benchmarks. Sometimes, fx in Create CSV Analyzer there is an absolute file path in the test causing the build to fail on other machines. Let's introduce a utility to mask file paths for easier testing.

tomaszguzialek:
Fixes #772


ClaudiaPHI:
LGTM!

PR #774, opened 2015-10-12
Made metric extraction work on results wrapped in AnalyzerResultFuture

kaspersorensen:
We came across this issue that metrics are not extracted from analyzer results that are wrapped in a future object.


tomaszguzialek:
LGTM.

Issue #779, opened 2015-10-12
NullPointerException on remote component PUT request 


jhorcicka:
When submitting a PUT request to get stateless result of a transformer without specifying "data" part of the input there is a NullPointerException. 

### Steps to reproduce
1. Run monitor. 
2. Submit a valid PUT request to check the correct functionality. 
  * URL: http://localhost:8888/repository/admin/components/Concatenator
  * BODY: 
  ```
  {
    "configuration": {
      "columns": ["c1", "c2"],
      "properties": {
        "Columns": ["c1", "c2"],
        "Separator": "-"
      }
    },
    "data": [
      ["v11", "v12"],
      ["v21", "v22"]
    ]
  }
  ```

3. Submit a non-valid PUT request to reproduce an error. 
  * URL: http://localhost:8888/repository/admin/components/Concatenator
  * BODY: 
  ```
  {
    "configuration": {
      "columns": ["c1", "c2"],
      "properties": {
      "Columns": ["c1", "c2"],
      "Separator": "-"
      }
    }
  }
  ``` 

### Actual outcome
500, No content, NullPointerException. 

### Expected outcome
200, Empty JSON result ({}).

PR #780, opened 2015-10-12
Bug #779

jhorcicka:
NullPointerException was fixed within a PUT request to get stateless result of a transformer without specifying "data" part of the input.


kaspersorensen:
LGTM

PR #783, opened 2015-10-12
Avoid potentially losing lines

### LINKED ISSUE #667: Graph UI lines of the OutputStream disappear
First Step: 
image

Second step: 
image

Steps to reproduce: 

1. Build new job.
2. Choose _Customers_ datastore.
3. Add _Completeness analyzer_.
4. Link _customers.csv_ to _Completeness analyzer_.
5. Add _Create Excel spreadsheet_.
6. Link "Complete rows" to _Create Excel spreadsheet_
7. Set sheet name to "test" in dialog (Why is this needed? Shouldn't there be a default?)
8. Add _String analyzer_.
9. Link "incomplete rows" to _String analyzer_
10. Add _Convert to number_.
11. Double click _Convert to number_ and choose all rows.
12. Close

Expected behavior: Everything mapped correctly.

Actual behavior: One or more components may lose their mapping. I've seen zero loss, both lost their mapping and only _Create Excel spreadsheet_ lose it's mapping. At no point did I see _String analyzer_ lose its mapping, that may be because it was mapped last. All components seems to keep their scope setting.

LosD:
This is more of a workaround than a proper fix, but a proper fix is a bit limited by our interface. The main issue is that the property widgets all have their own listeners directly connected to the builders, making it hard to control listeners without yanking the entire dialog window (which is actually still not bullet-proof). A proper fix would probably only allow (or strongly encourage) listeners to listen to their dialog, which would then have a general component listener (also not currently supported) on the component and/or analysisJobListener.

Luckily, only MultipleInputColumnsPropertyWidget seems to have been directly looking at the input from the listener, instead of just updatingfrom the general state, which should make the other TransformerChangeListener-implementing PropertyWidgets immune. Heavy testing is still recommended, though.

Fixes #667
Also contains a fix for re-adding listeners, as well as a wrong addTransformerChangeListener on panel close (none of them actually .

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/panels/fuse/ColumnListMultipleCoalesceUnitPropertyWidget.java
+++ b/desktop/ui/src/main/java/org/datacleaner/panels/fuse/ColumnListMultipleCoalesceUnitPropertyWidget.java
@@ -25,6 +25,7 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


kaspersorensen:
LGTM and everything in here is quite safe improvements, so thumbs up and from my side no need to do extensive testing before merging. It looks good and _at least_ an improvement to previous code.

PR #787, opened 2015-10-13
Incompletely configured components in output data streams does not show in window

### LINKED ISSUE #781: Incompletely configured components in output data streams does not show in window
The main window looks happy even if a component in an output data stream has problems.
image
(Note the Excel component with an issue, but the text in the bottom says "Ready to execute")

ClaudiaPHI:
Fixes #781 
check if the sub-jobs are configured as well.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
@@ -46,14 +46,12 @@ import javax.swing.TransferHandler;
 
 import org.apache.metamodel.schema.Column;
 import org.apache.metamodel.schema.Table;
-import org.datacleaner.api.OutputDataStream;
 import org.datacleaner.bootstrap.WindowContext;
 import org.datacleaner.data.MetaModelInputColumn;
 import org.datacleaner.descriptors.ComponentDescriptor;
 import org.datacleaner.descriptors.ConfiguredPropertyDescriptor;
 import org.datacleaner.job.AnalysisJob;
 import org.datacleaner.job.builder.AnalysisJobBuilder;
-import org.datacleaner.job.builder.AnalyzerComponentBuilder;
 import org.datacleaner.job.builder.ComponentBuilder;
 import org.datacleaner.job.builder.UnconfiguredConfiguredPropertyException;
 import org.datacleaner.panels.DCPanel;
@@ -305,7 +303,8 @@ public final class JobGraph {
                     imagePath = null;
 
                     try {
-                        if (checkAnalysisJobBuilderIsConfigured()) {
+                        if (_analysisJobBuilder.isConfigured(true)
+                                && _analysisJobBuilder.isConsumedOutDataStreamsJobBuilderConfigured(true)) {
                             title = "Ready to execute";
                             subTitle = "Click the 'Execute' button in the upper-right\ncorner when you're ready to run the job.";
                             imagePath = "images/window/canvas-bg-execute.png";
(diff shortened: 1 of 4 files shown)


ClaudiaPHI:
Before: 

screen shot 2015-10-13 at 2 02 37 pm


After: 


screen shot 2015-10-13 at 1 57 13 pm

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
```diff
@@ -456,4 +458,28 @@ private void setScrollbarValue(JScrollBar scrollBar, int value) {
...
+    /**
+     * Checks if a job including its child jobs are correctly configured.
+     * 
+     * @return
+     */
+    private boolean checkAnalysisJobBuilderIsConfigured() {
+
+        boolean isConfigurated = true;
+        final List<AnalyzerComponentBuilder<?>> analyzerComponentBuilders = _analysisJobBuilder
+                .getAnalyzerComponentBuilders();
```
kaspersorensen: Output data streams are not exclusive to analyzers, so this is the wrong collection to look at. Don't we have a method to get all components that have active output data streams?

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
```diff
@@ -456,4 +458,28 @@ private void setScrollbarValue(JScrollBar scrollBar, int value) {
...
+        for (AnalyzerComponentBuilder<?> analyzerComponentBuilder : analyzerComponentBuilders) {
+            final List<OutputDataStream> outputDataStreams = analyzerComponentBuilder.getOutputDataStreams();
+            if (!outputDataStreams.isEmpty()) {
+                for (OutputDataStream outputDataStream : outputDataStreams) {
+                    final AnalysisJobBuilder outputDataStreamJobBuilder = analyzerComponentBuilder
+                            .getOutputDataStreamJobBuilder(outputDataStream);
```
kaspersorensen: First check if the output data stream is being consumed - or else you will create new job builders on the fly with this call, which is not what we want.

kaspersorensen: getConsumedOutputDataStreamsJobBuilders() should be what you use I think.

ClaudiaPHI: I will look.

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -689,6 +685,36 @@ public void updateStatusLabel() {
         _executionAlternativesButton.setEnabled(executeable);
     }
 
+    /**
+     * Checks if a job including its child jobs are correctly configured.
+     **/
+    private void checkAnalyzerJobsAreConfigured() {
```
kaspersorensen: I see some code duplication here ... Could we push this logic down to the AnalysisJobBuilder instead of having it duplicated in the UI? I think the validate method that we had before should be a good place to have it.

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -1368,4 +1368,17 @@ public AnalysisJobBuilder getRootJobBuilder() {
         }
         return consumedOutputDataStreamJobBuilders;
     }
+
+    /**
+     * Checks if a job including its child jobs are correctly configured.
+     **/
+    public boolean isConsumedOutDataStreamsJobBuilderConfigured(final boolean throwException) {
+        final List<AnalysisJobBuilder> consumedOutputDataStreamsJobBuilders = getConsumedOutputDataStreamsJobBuilders();
+        boolean isConfigurated = true;
```
kaspersorensen: should probably be "isConfigured" but actually I think you can avoid it completely and just return false the moment you find something that is not configured, or return true otherwise.

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -1368,4 +1373,17 @@ public AnalysisJobBuilder getRootJobBuilder() {
...
         return consumedOutputDataStreamJobBuilders;
     }
+
+    /**
+     * Checks if a job children are configured.
+     **/
+    private boolean isConsumedOutDataStreamsJobBuilderConfigured(final boolean throwException) {
+        final List<AnalysisJobBuilder> consumedOutputDataStreamsJobBuilders = getConsumedOutputDataStreamsJobBuilders();
+        for (final AnalysisJobBuilder analysisJobBuilder : consumedOutputDataStreamsJobBuilders) {
+            if (!analysisJobBuilder.checkConfiguration(throwException)) {
```
kaspersorensen: also call isConfigured here so that you also get children-of-children

LosD:
I believe we can merge this now (as far as I remember, we were just waiting for Travis to get around to building/testing the PR)?

ClaudiaPHI:
Yes. We can!

PR #789, opened 2015-10-13
Added tooltip to the file selector

### LINKED ISSUE #788: Filename fields should show the filename in the tooltip
Filename fields should show the filename in the tooltip, which would be a lot more helpful than the current approach of just saying "filename".

2015-10-13 16_00_53-create csv file _ datacleaner

stefanjanssen:
So that you can actually see what you are entering. Fixes #788 .

2015-10-13 16_14_29-orderdb _ analysis job _ datacleaner


stefanjanssen:
Worst timing ever.

PR #790, opened 2015-10-13
Added tooltip to the filename textfield

### LINKED ISSUE #788: Filename fields should show the filename in the tooltip

stefanjanssen:
Fixes #788


kaspersorensen:
LGTM

PR #798, opened 2015-10-14
Issue #611: Added 'additional output fields' to Completeness Analyzer

### LINKED ISSUE #611: Add an "additional output fields" property to Completeness Analyzer
Now that we've enriched Completeness Analyzer to have output data streams, it seems strange that it only provides the fields that are being analyzed, and cannot yield e.g. a primary key or any other value which will be really useful for post-processing.

I see three solutions:

 * Ideally we would want some way to tell the output data stream builder that "all visible fields" should be added. But I have no clue how to implement that so I'm gonna skip that possibility for now. If anyone has a bright idea then I suggest making a separate issue about it.
 * We could add another criteria for the completeness check like "Disregard". That way you can add columns to the analysis without including them in any evaluation.
 * We could add another (optional) ```InputColumn[]``` property which simply adds to the fields that are being made available in the output.

I guess this story implies a decision to take the last option here. But I am also open towards people preferring other solutions. Basically I think they are equally imperfect ;-)

kaspersorensen:
Fixes #611 

Here's a screenshot where I am doing analysis only on firstname+lastname, but then following it with processing of jobtitle (an additional field).

image

### DIFF
--- a/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
+++ b/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
@@ -90,22 +90,22 @@ public class CompletenessAnalyzer implements Analyzer<CompletenessAnalyzerResult
     }
 
     @Inject
-    @Configured(PROPERTY_VALUES)
+    @Configured(order=1, value=PROPERTY_VALUES)
     @Description("Values to check for completeness")
     InputColumn<?>[] _valueColumns;
 
     @Inject
-    @Configured(PROPERTY_CONDITIONS)
+    @Configured(order=1, value=PROPERTY_CONDITIONS)
     @Description("The conditions of which a value is determined to be filled or not")
     @MappedProperty(PROPERTY_VALUES)
     Condition[] _conditions;
 
     @Inject
-    @Configured(PROPERTY_EVALUATION_MODE)
+    @Configured(order=3, value=PROPERTY_EVALUATION_MODE)
     EvaluationMode _evaluationMode = EvaluationMode.ANY_FIELD;
 
     @Inject
-    @Configured(value = PROPERTY_ADDITIONAL_OUTPUT_VALUES, required = false)
+    @Configured(order=100, value= PROPERTY_ADDITIONAL_OUTPUT_VALUES, required = false)
     @Description("Optional additional values to add to output data streams")
     InputColumn<?>[] _additionalOutputValueColumns;
 
--- a/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
+++ b/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
@@ -19,7 +19,10 @@
  */
 package org.datacleaner.beans;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.inject.Inject;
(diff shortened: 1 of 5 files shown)


components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
```diff
@@ -86,21 +90,26 @@ public String getName() {
...
 
     @Inject
-    @Configured(PROPERTY_VALUES)
+    @Configured(order=1, value=PROPERTY_VALUES)
     @Description("Values to check for completeness")
     InputColumn<?>[] _valueColumns;
 
     @Inject
-    @Configured(PROPERTY_CONDITIONS)
+    @Configured(order=1, value=PROPERTY_CONDITIONS)
```
tomaszguzialek: Both configured properties have order = 1.

components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
```diff
@@ -191,7 +215,7 @@ public void fillAllConditions(Condition condition) {
         final OutputDataStreamBuilder incompleteStreamBuilder = OutputDataStreams
                 .pushDataStream(OUTPUT_STREAM_INCOMPLETE);
 
-        for (InputColumn<?> column : _valueColumns) {
+        for (InputColumn<?> column : createOutputDataStreamColumns()) {
```
tomaszguzialek: Is that necessary to create them here instead of iterating _outputDataStreamsColumns? I thought they are created once in the init method.

components/basic-analyzers/src/test/java/org/datacleaner/beans/CompletenessAnalyzerTest.java
```diff
@@ -241,4 +248,107 @@ public void testOutputDataStream() throws Throwable {
...
+            analyzer2.addInputColumns(completeDataStreamColumns);
+            analyzer2.setName("analyzer2");
+            assertTrue(analyzer2.isConfigured());
+            assertTrue(analyzer1.isOutputDataStreamConsumed(completeStream));
+
+            final AnalyzerComponentBuilder<MockAnalyzer> analyzer3 = incompleteDataStreamJobBuilder
+                    .addAnalyzer(MockAnalyzer.class);
+            analyzer3.addInputColumns(incompleteDataStreamColumns);
+            analyzer3.setName("analyzer2");
+            assertTrue(analyzer2.isConfigured());
```
tomaszguzialek: Shouldn't that be analyzer3 instead of analyzer2?

components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
kaspersorensen: will fix.

components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
kaspersorensen: the getOutputDataStreams() method is called many times again and again, and if reusing same instance we become stale easily as configuration changes. I only wanted to use _outputDataStreamsColumns in the run(...) method because that method is called after build and initialization.

components/basic-analyzers/src/test/java/org/datacleaner/beans/CompletenessAnalyzerTest.java
kaspersorensen: Probably yes. It's a copy of the similar test, but I will fix both instances.

components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
tomaszguzialek: Makes sense.

components/basic-analyzers/src/test/java/org/datacleaner/beans/CompletenessAnalyzerTest.java
```diff
@@ -203,7 +210,7 @@ public void testOutputDataStream() throws Throwable {
             final AnalyzerComponentBuilder<MockAnalyzer> analyzer3 = incompleteDataStreamJobBuilder
                     .addAnalyzer(MockAnalyzer.class);
             analyzer3.addInputColumns(incompleteDataStreamColumns);
-            analyzer3.setName("analyzer2");
+            analyzer3.setName("analyzer3");
             assertTrue(analyzer2.isConfigured());
```
tomaszguzialek: Again, this should be analyzer3.

tomaszguzialek:
Found one more instance of probable copy-paste error. Otherwise, mergeable!

PR #799, opened 2015-10-14
Issue #771: Tweaked the search text field to not select upon focus

### LINKED ISSUE #771: Search-anywhere sometimes miss first character
In some cases, if you just start typing when searching for components, the first character may be lost.

It looks like the first character is selected, and therefore removed when typing the next character.

### Steps to reproduce
1. Create new job (using whatever datastore).
2. Click datastore/library pane.
3. Start typing "completeness analyzer". A few letters will do, so type "comp".

### Actual outcome
First character lost, so "omp" will be in the search field.

### Expected outcome
"comp" should be in the search field.

kaspersorensen:
Fixes #771 

During testing I did have a 1-out-of-hundred chance that sometimes I would get double-begnning letters. I'm not sure why, but to me this seems now at least to be improved a lot and a lot less annoying.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/panels/SchemaTreePanel.java
+++ b/desktop/ui/src/main/java/org/datacleaner/panels/SchemaTreePanel.java
@@ -37,6 +37,10 @@ import javax.swing.JScrollPane;
 import javax.swing.SwingWorker;
 import javax.swing.border.EmptyBorder;
 import javax.swing.event.DocumentEvent;
+import javax.swing.text.AttributeSet;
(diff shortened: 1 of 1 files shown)


LosD:
Hmmm... Just tested. First two tries yielded double-first-letter.

Why the code looks good, I don't think we can call it ready.

LosD:
Tested a bit more. After working in the main graph, it seems to be around every 2 or 3 times (click in main panel, click in toolbox panel, type). It also seems to be dependent on writing speed, but it's wildly inconsistent, so I can't really say for sure.

PR #802, opened 2015-10-14
Feature/datahubdatastore xml serialization

ankit2711:
Name of the branch does not really give a good insight of the contents/improvements. 

In short -
1. Changes to make the DataHub connector work for default security mode.
2. Changes to store the datahub datastore in the conf.xml 
3. Contractual changes associated with the DataHub REST contract.
4. Bug fixes in the datahub connector.


engine/core/src/test/java/org/datacleaner/metamodel/datahub/DataHubUpdateCallbackTest.java
```diff
@@ -44,48 +46,50 @@
...
         }
-        verify(dataContext, times(1)).executeUpdates(Matchers.anyListOf(UpdateData.class)); //Tests if it was called a second time
+        verify(dataContext, times(1)).executeUpdates(Matchers.anyListOf(UpdateData.class)); // Tests
+                                                                                            // if
+                                                                                            // it
+                                                                                            // was
+                                                                                            // called
+                                                                                            // a
+                                                                                            // second
+                                                                                            // time
```
kaspersorensen: Not important, but I would suggest to put this comment above the line so we dont get this line wrapping issue.

engine/xml-config/src/main/java/org/datacleaner/configuration/DatastoreXmlExternalizer.java
```diff
@@ -415,6 +422,31 @@ public Element toElement(SalesforceDatastore datastore) {
...
+        if (!StringUtils.isNullOrEmpty(datastore.getDescription())) {
+            ds.setAttribute("description", datastore.getDescription());
+        }
+
+        appendElement(ds, "host", datastore.getHost());
+        appendElement(ds, "port", datastore.getPort());
+        appendElement(ds, "username", datastore.getUsername());
+        appendElement(ds, "password", encodePassword(datastore.getPassword()));
+        appendElement(ds, "tenantname", datastore.getTenantName());
+        appendElement(ds, "https", datastore.https());
```
kaspersorensen: Not related to this PR, but the method name https() should probably be "isHttps()" if it returns a boolean?

engine/xml-config/src/main/java/org/datacleaner/configuration/DatastoreXmlExternalizer.java
```diff
@@ -415,6 +422,31 @@ public Element toElement(SalesforceDatastore datastore) {
...
+        appendElement(ds, "acceptunverifiedsslpeers", datastore.acceptUnverifiedSslPeers());
```
kaspersorensen: ... and also here I would prefer "isAcceptUnverifiedSslPeers()"

kaspersorensen:
I had only remarks that are minor and/or not related to this specific PR (just stuff I noticed while reading the PR).

So all in all: LGTM, but for extra points also consider the remarks I made in an additional update :-)

PR #804, opened 2015-10-15
Feature/767 value dist unique last

### LINKED ISSUE #767: Value distribution: Put <unique> at the bottom of the value list
Right now the composite value ```<unique>``` is arranged in a similar way to other values. This usually means that it will show up like this maybe:
```
 * Green -> 230
 * Orange -> 178
 * <unique> -> 5
 * Red -> 2
```

But since uniques are basically of count 1 we should have them at the bottom as the least frequent values are put there.

tomaszguzialek:
Fixes #767

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/result/ValueDistributionResultSwingRendererGroupDelegate.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/result/ValueDistributionResultSwingRendererGroupDelegate.java
@@ -26,8 +26,6 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/result/ValueDistributionResultSwingRendererGroupDelegate.java
```diff
@@ -229,6 +231,26 @@ public void actionPerformed(ActionEvent e) {
...
+        ValueFrequency uniqueValueFrequency = null;
+        for (ValueFrequency valueFrequency : valueCounts) {
+            if ("<unique>".equals(valueFrequency.getName())) {
+                uniqueValueFrequency = valueFrequency;
+                break;
+            }
+        }
+        
+        if (uniqueValueFrequency != null) {
+            List<ValueFrequency> valueCountsList = new ArrayList<>(valueCounts);
```
ClaudiaPHI: final. just to be mean :-P

kaspersorensen:
Another approach that would be a bit more general is to simply update the compareTo(...) method of ValueFrequency!

tomaszguzialek:
Tried that approach now, but cannot make it work for some reason. Maybe it is something to do with equals vs compareTo inconsistency? Do you want compareTo approach or the existing one is good enough?

kaspersorensen:
If it shows to be cumbersome then let's not do it that way. Then I am fine with this approach.

PR #805, opened 2015-10-15
Fixes #769. added custom installed drivers to the panel

### LINKED ISSUE #769: Custom JDBC drivers not shown on driver list
In the Options dialog we have a list of database drivers available. You can also register your own driver using the buttons available.

But if you add a driver that we don't know, nothing shows up on the list, even if the driver is installed. We should at least list the driver there so that you can see it is installed.

ClaudiaPHI:
Fixes #769 

This is how it looks now: 

screen shot 2015-10-15 at 9 59 46 am

Would you prefer it without image ? 
screen shot 2015-10-15 at 9 55 54 am


tomaszguzialek:
LGTM. Generic datastore icon is OK, IMO. I would merge it when Travis allows.

ClaudiaPHI:
Let's see what @kaspersorensen thinks.

kaspersorensen:
Hmm in this case it's a bit weird because the Oracle driver is also in the "known drivers" list! I would actually then just expect it to be marked as installed then.

The scenario we are looking to fix is when you install a driver that we do NOT know already in the list.

ClaudiaPHI:
Then I will also add a check if it in the known list

LosD:
Wow, Github doesn't play so nice when commenting on the commit instead of the PR. Sorry, about that.

LosD:
LGTM, but see commit comment.

ClaudiaPHI:
can we merge this ?

LosD:
YES! :)

PR #810, opened 2015-10-15
Fix listener loop

### LINKED ISSUE #809: Listener loop in new FilenameTextField
Seems like I overlooked a pretty serious drawback to the refactoring. There's a chance of a listener loop on filename update (change -> listener react with updating the filename -> change -> listener react, etc.).

LosD:
This should make sure FilenameTextField listeners are protected.

Fixes #809


LosD:
Manual testing was happy

stefanjanssen:
This solves the issues I was having.

PR #812, opened 2015-10-15
Feature/785 refresh of descriptor providers

jhorcicka:


desktop/ui/src/main/resources/datacleaner-home/conf.xml
```diff
@@ -67,10 +67,18 @@
...
-		<package recursive="true">com.neopost</package>
-	</classpath-scanner>
-
+    <descriptor-providers>
+        <classpath-scanner>
+            <package recursive="true">org.datacleaner</package>
+            <package recursive="true">com.hi</package>
+            <package recursive="true">com.neopost</package>
+        </classpath-scanner>
+        <remote-components>
```
jakubneubauer: For now I would comment the <remote-components> section, until datacleaner.org service will be available.

engine/core/src/main/java/org/datacleaner/configuration/DataCleanerEnvironmentImpl.java
```diff
@@ -27,7 +27,7 @@
 import org.datacleaner.storage.StorageProvider;
 
 /**
- * Default (immutable) implementation of {@link DataCleanerEnvironment}.
+ * Default (immutable) implementation of {@liDataCleanerEnvironmentImplnk DataCleanerEnvironment}.
```
jakubneubauer: typo

engine/rest-client/pom.xml
```diff
@@ -15,6 +15,11 @@
 			<artifactId>DataCleaner-api</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+        <dependency>
+            <groupId>org.eobjects.datacleaner</groupId>
+            <artifactId>DataCleaner-engine-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
jakubneubauer: I would try to avoid this dependency

PR #815, opened 2015-10-16
Feature/760 preview alternatives

### LINKED ISSUE #760: Ability to select how many records to preview in Transformer preview
Right now we always get a fixed number of records (200 to be precise) for doing a Transformer preview.

Some transformers take longer than others, and some transformers may cost money/credits to use. And sometimes you may need a slightly bigger preview to see what you're looking for.

So we should add a way to select how many records to include in the preview. My suggestion would be to add a down-arrow next to the preview-button, like in the Execute button, which would allow you to select the amount of records.

tomaszguzialek:
Fixes #760

preview_alternatives

preview_alternatives_open

preview_alternatives_maxrowsdialog


desktop/api/src/main/java/org/datacleaner/panels/TransformerComponentBuilderPanel.java
```diff
@@ -88,14 +94,52 @@ protected TransformerComponentBuilderPanel(Image watermarkImage, int watermarkHo
...
+                    final JMenuItem defaultPreviewMenutItem = WidgetFactory.createMenuItem("Preview " + defaultPreviewRows + " rows",
+                            IconUtils.ACTION_PREVIEW);
+                    defaultPreviewMenutItem.addActionListener(defaultPreviewTransformedDataActionListener);
+
+                    final JMenuItem maxRowsPreviewMenuItem = WidgetFactory.createMenuItem("Run first N records",
+                            IconUtils.ACTION_PREVIEW);
+                    maxRowsPreviewMenuItem.addActionListener(new ActionListener() {
+                        
+                        @Override
+                        public void actionPerformed(ActionEvent e) {
```
LosD: image

LosD:
LGTM

desktop/api/src/main/java/org/datacleaner/panels/TransformerComponentBuilderPanel.java
tomaszguzialek: Well... yeah :D.

ClaudiaPHI:
LGTM!

kaspersorensen:
There's a double border I think ...

kaspersorensen:
What I mean is that it looks 2px thick inbetween the two buttons.

tomaszguzialek:
preview_alternatives_open

ClaudiaPHI:
:+1:

LosD:
A beaut'!

kaspersorensen:
Tried it out - works very well :-) My only remark is that I notice we're now moving quite some stuff into WidgetFactory lately (not just this PR) and maybe we should consider if that's a good spot for it or not. But I don't want to let that block this PR ... MERGING! :-) :fireworks:

PR #821, opened 2015-10-16
Issue #817: Copied CUSTOMERS table from desktop to monitor

### LINKED ISSUE #817: OrderDB's customer table is missing JOBTITLE in the DC monitor
Some time ago we updated the CUSTOMERS table in DC desktop. But I think we forgot to also update it in monitor, which means that it's easy to make a job that's not portable between the two. It should be easy, so let's update the one in monitor too! :-)

kaspersorensen:
Fixes #817


ClaudiaPHI:
:scream_cat:

ClaudiaPHI:
:sob:

ClaudiaPHI:
LGTM!

PR #822, opened 2015-10-16
Check connection always before saving.

HettyK:


desktop/ui/src/main/java/org/datacleaner/windows/DataHubDatastoreDialog.java
```diff
@@ -264,6 +286,11 @@ protected boolean validateForm() {
             return false;
         }
 
+        if (_tenantName == null) {
+            setStatusError("Please test the connection first");
+            return false;
```
kaspersorensen: I think this then prevents you from saving a datastore without testing it? Isn't that a bit annoying? What if you're simply editing an existing datastore or maybe you temporarily don't have access to the datahub? Testing the datastore can IMO be postponed till the moment it is actually connecting. I guess also username/passwords may change and all that, so verifying it at this time is only giving a false sense of verification.

drexler42: That is true, but the alternatives looked worse. We tried getting the alias when you hit the Save button. But when you then fail to get the tenant name (e.g. because the credentials are wrong), then you are faced with an exception in the Save button handling.
Retrieving the tenant name requires most of the other settings to be already correct. So in the end the best option seemed to be to always require testing the connection before saving.

kaspersorensen: How about not saving the tenant name at all, but simply requesting it when it is needed in the datacontext?

drexler42:
The tenant is now no longer stored as part of the datahub datastore properties. Instead, it is retrieved when a data context is created,

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDataContext.java
```diff
@@ -166,8 +174,43 @@ public void executeGoldenRecordDelete(String goldenRecordId) {
...
+                    userInfo = new ObjectMapper().readValue(content, UserInfo.class);
+                    return userInfo.tenant;
+                } finally {
+                    FileHelper.safeClose(content);
+                }
+            } else {
+                final String reasonPhrase = statusLine.getReasonPhrase();
+                throw new RuntimeException("Failed to retrieve the tenant name: " + reasonPhrase);
+            }
+        } catch (Exception e) {
```
kaspersorensen: Minor issue: This will also catch the RuntimeException thrown just above, so you'll have an additional level of exception causes to unravel. It would probably be good to check here if it is instanceof RuntimeException and if so, rethrow it instead of wrapping it.

kaspersorensen:
LGTM! I'll merge because I understand the time constraints we have right now. But maybe check the remark anyway if you have time to fix it later today that could be an option.

PR #823, opened 2015-10-17
Equals filter outcome naming

### LINKED ISSUE #797: Rename outcomes of EqualsFilter to EQUAL and NOT_EQUAL instead of valid/invalid
The fact that the equals filter is using an outcome enum with values VALID and INVALID is wrong. From the standpoint of the filter we cannot judge what is valid or not, only if it is equal or not. So the outcome values should be EQUAL or NOT_EQUAL

Note: Do add ```@Alias("VALID")``` and  ```@Alias("INVALID")``` to the new enum values in order to retain backwards compatibility whenever the outcomes are referred to by string names.

kaspersorensen:
This fixes #797 and in general I am quite happy with the change.

BUT I would be lying if I said this PR also didn't have any negative consequences. That would be on compatibility of code that is programmatically using the Equals filter. Basically such code needs to be adapted, as can be seen by the diff. I know we have it in use here and there in commercial editions of DC, so that would need to be updated as well.

The ```@Alias``` on the new enum ensures that at least .analysis.xml files are still backwards compatible. The only non-backward compatible change here is that programmatic instantiation of the component has changed.

### DIFF
--- a/components/basic-filters/src/main/java/org/datacleaner/beans/filter/EqualsFilter.java
+++ b/components/basic-filters/src/main/java/org/datacleaner/beans/filter/EqualsFilter.java
@@ -54,7 +54,13 @@ import com.google.common.base.Joiner;
 @Description("A filter that excludes values that are not equal (=) to specific set of valid values")
 @Categorized(FilterCategory.class)
 @Distributed(true)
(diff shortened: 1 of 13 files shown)


kaspersorensen:
Checked DataHub's codebase for potential conflicts. We have 4 locations where this would hit them... Have to talk with @ankit2711 if he thinks it's OK to change.

PR #824, opened 2015-10-18
Issue #806: Nano-second precision in CaptureChangedRecords filter

### LINKED ISSUE #806: CaptureChangedRecordsFilter loses the microseconds when a timestamp based column is used
The CaptureChangedRecordsFilter loses the microseconds while reading back the long date value stored in the state file. This causes some jobs which use this filter for incremental runs to keep running through the same records. This becomes worse when the jobs are scheduled to run with a fixed time, even when no new records are added these scheduled jobs keep running through the same records again and again. 

We need to find a way that the date conversion from Long value (stored in the state file) does not loose on the microsecond value.

This is a blocker.

kaspersorensen:
Fixes #806 

I would love a review by @ankit2711 - can you check it?


ankit2711:
The fix works with MM change with Postgres, but somehow I cannot make a new job, the Resource arguments for the State file in CaptureChangedRecordsFilter and the file resource for Write to CSV are somehow lost. Do I miss something in this branch..

PR #827, opened 2015-10-19
Feature/add documentation composite datastore

### LINKED ISSUE #794: Add documentation about composite datastore

ClaudiaPHI:
Fixes  #794 

2 sentences that look like this: 
screen shot 2015-10-19 at 10 44 16 am


LosD:
I'm not sure I understand the image. Isn't this in a section that describes datastores in general? It's a little weird that the only image is of a very special datastore type.

ClaudiaPHI:
I was thinking to add one for datastores in general. I did not see other suitable section where I can add the definition. Moreover, I was thinking that keeping things in the same place is easier to maintain.

kaspersorensen:
The position for this is OK, but then I am missing an intro and a little header inbetween because this right now looks like composite datastores are the most important type of datastores while it is in deed for the advanced users. What I mean is that this chapter is very early in the documentation so we need to introduce simple concepts first. Therefore I suggest to add before this new content:

 * A paragraph that tells the reader that the various datastore types can be created via the UI or via the configuration file. Make a link to the chapter about the configuration file.
 * A header (subheader or whatever - smaller than "What is a datastore?") saying "Composite datastores"

I would like also an XML example of the composite datastore. But not in this chapter, in the chapter about configuration file (which contains also a lot of other example xml datastore definitions).

LosD:
I would also suggest an example image of a simpler datastore type, like CSV, then.

ClaudiaPHI:
screen shot 2015-10-19 at 1 31 39 pm

screen shot 2015-10-19 at 1 33 47 pm
screen shot 2015-10-19 at 1 34 01 pm

screen shot 2015-10-19 at 1 34 09 pm

ClaudiaPHI:
Not exactly the best screenshots, but you get the idea

LosD:
Hmmm... Now there's instructions in the concepts part... And those are already in chapter two.

ClaudiaPHI:
screen shot 2015-10-19 at 2 17 24 pm

ClaudiaPHI:
screen shot 2015-10-19 at 2 43 07 pm


screen shot 2015-10-19 at 2 45 44 pm


"Just say yes" - (band Snow Patrol ).

LosD:
:+1: from me

documentation/src/docbkx/chapter11-concepts.xml
```diff
@@ -117,6 +117,33 @@
...
 			of these datastores as well.
 		</para>
+		<para>
+			A datastore can be created in the UI or via
+			<link linkend="chapter_configuration_file"> the configuration file</link>
+			.
+			You can create a datastore from any type of source such
+			as:
+			CSV,
+			EXCEL, Oracle Database,MySQL, etc.
```
kaspersorensen: "Excel" in normal case.
Space after comma, before "MySQL".

documentation/src/docbkx/chapter41-configuration-file.xml
```diff
@@ -368,17 +375,34 @@
 				first 1000 documents of each database.
 			</para>
 		</section>
+		<section>
+			<title>Composite datastore</title>
+			<para> This is an example of a composite datastore. It contains data
+				from 2 other datastores: Datastore1 and Datastore 2.
```
kaspersorensen: Missing space in "Datastore1".

documentation/src/docbkx/chapter41-configuration-file.xml
```diff
@@ -437,15 +461,15 @@
 				&lt;reference-data-catalog&gt;
 				&#160;&#160;&lt;synonym-catalogs&gt;
 				&#160;&#160;&#160;...
-				&#160;&#160;&#160;&#160;&lt;text-file-synonym-catalog name="textfile_syn" description="My text file synonyms"&gt;
+				&#160;&#160;&#160;&#160;&lt;text-file-synonym-catalog
+				name="textfile_syn" description="My text file synonyms"&gt;
```
kaspersorensen: This is an unfortunate reformatting because it happens in <programlisting> where newlines are actually significant!

Normally I use &#160; instead of space to avoid this. Not sure why it wasn't applied here in the first place.

documentation/src/docbkx/chapter41-configuration-file.xml
```diff
@@ -453,7 +477,8 @@
 				&lt;reference-data-catalog&gt;
 				&#160;&#160;&lt;synonym-catalogs&gt;
 				&#160;&#160;&#160;...
-				&#160;&#160;&#160;&#160;&lt;datastore-synonym-catalog name="datastore_syn" description="My datastore synonyms"&gt;
+				&#160;&#160;&#160;&#160;&lt;datastore-synonym-catalog
+				name="datastore_syn" description="My datastore synonyms"&gt;
```
kaspersorensen: Same thing here

documentation/src/docbkx/chapter41-configuration-file.xml
```diff
@@ -474,14 +501,14 @@
 				&lt;reference-data-catalog&gt;
 				&#160;&#160;&lt;string-patterns&gt;
 				&#160;&#160;&#160;...
-				&#160;&#160;&#160;&#160;&lt;regex-pattern name="regex danish email" description="Danish email addresses"&gt;
+				&#160;&#160;&#160;&#160;&lt;regex-pattern name="regex danish email"
+				description="Danish email addresses"&gt;
```
kaspersorensen: Same thing here

kaspersorensen:
Sorry :imp: :smile:

kaspersorensen:
Superceeded by #834

PR #831, opened 2015-10-19
Fix file resource property widget

### LINKED ISSUE #830: Resource property widget does not save it's value to the property

kaspersorensen:
Fixes #830

This was essentially a mixup from our previous story of refactoring FilenameTextField. This time I made the code a bit more robust by adding some unittests to ensure we don't mess up the listener expectations again.


LosD:
LGTM!

PR #833, opened 2015-10-19
Feature/795 add union documentation

### LINKED ISSUE #795: Add documentation about Union component
(and give example with a composite datastore - see #794 )

ClaudiaPHI:
Fixes #795


kaspersorensen:
Under the image I would like to have a quick explanation of the UI for selecting columns. It should explain how aligned columns will combined into one column in the output stream of the Union.

PR #846, opened 2015-10-21
Bug/845 nlcc transformer link problem

jhorcicka:
RemoteConfiguredPropertyDescriptor: Object.class is returned as default type argument instead of null. Linking now uses the first input column.


kaspersorensen:
LGTM (although more type information in the future will be better).

PR #849, opened 2015-10-23
Make ImmutableOutputDataStreamJob serializable

### LINKED ISSUE #842: AnalyzerResults produced within an Output Data Streams are not serializable
It seems that when we use output data streams and produce results from components within that stream, then saivng the result fails because of a non-transient reference somewhere to ImmutableAnalysisJob (I bet this is the child job referenced somewhere).

Example job:
image

Error scenario:
image

LosD:
It used to be impossible to serialize jobs with output datastreams, because ImmutableOutputDataStreamJob contains a non-transient reference to AnalysisJob. Making it transient fixes that without breaking the serialized result.

Fixes #842

### DIFF
--- a/engine/core/src/test/java/org/datacleaner/test/full/scenarios/JobWithOutputDataStreamsTest.java
+++ b/engine/core/src/test/java/org/datacleaner/test/full/scenarios/JobWithOutputDataStreamsTest.java
@@ -27,6 +27,7 @@ import static org.junit.Assert.assertTrue;
 
 import java.util.List;
 
+import org.apache.commons.lang.SerializationUtils;
 import org.datacleaner.api.OutputDataStream;
 import org.datacleaner.configuration.DataCleanerConfiguration;
 import org.datacleaner.configuration.DataCleanerConfigurationImpl;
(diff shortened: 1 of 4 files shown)


ClaudiaPHI:
LGTM!

engine/core/src/test/java/org/datacleaner/test/full/scenarios/JobWithOutputDataStreamsTest.java
```diff
@@ -146,14 +148,18 @@ public void testSimpleBuildAndExecuteScenario() throws Throwable {
 
         assertEquals(2, resultFuture.getResults().size());
 
+        final byte[] serialized = SerializationUtils.serialize(new SimpleAnalysisResult(resultFuture.getResultMap()));
+
+        SimpleAnalysisResult deSerializedResult = (SimpleAnalysisResult) SerializationUtils.deserialize(serialized);
```
kaspersorensen: final ;-)

LosD: Damn!

kaspersorensen:
Build break is strange, but I cannot see how it can be related. Merge?

PR #850, opened 2015-10-23
Do not use injected RowAnnotations

### LINKED ISSUE #837: Completeness analyzer should always be able to build row annotation sample set
If you use the Completeness analyzer in a job that acquires many row annotations, you might run into a situation where the annotation acquired by the Completeness analyzer is not getting a sample set in the RowAnnotationFactory. This is unfortunate because it is not just any sample, but the only sample that the result uses. The result screen will appear empty, except for the incomplete record count.

Somehow we should ensure that the row annotation injected in Completeness analyzer takes priority and is always granted a sample set.

LosD:
As the CompletenessAnalyzer must always get row annotations for the result to show, we cannot use the shared row annotations.

Fixes #837

### DIFF
--- a/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
+++ b/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
@@ -111,18 +111,13 @@ public class CompletenessAnalyzer implements Analyzer<CompletenessAnalyzerResult
     InputColumn<?>[] _additionalOutputValueColumns;
 
     // Do not inject the shared RowAnnotations, available rows are always needed.
-    final RowAnnotationFactory _annotationFactory = new InMemoryRowAnnotationFactory2();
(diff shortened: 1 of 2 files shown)


components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
```diff
@@ -109,13 +110,9 @@ public String getName() {
...
 
-    @Inject
-    @Provided
-    RowAnnotation _invalidRecords;
-
-    @Inject
-    @Provided
-    RowAnnotationFactory _annotationFactory;
+    // Do not inject the shared RowAnnotations, available rows are always needed.
+    final RowAnnotationFactory _annotationFactory = new InMemoryRowAnnotationFactory2();
```
kaspersorensen: private ? :)

And wouldn't it be prettier to instantiate/assign this variable in the constructor - like the _invalidRecords variable?

PR #851, opened 2015-10-23
Stop when returned rows is less than requested rows

drexler42:
This is a small change that saves 1 REST call at the end of the execution of the job. Until now, the job ended when the REST call returned an empty result. Now, each returned batch of records is counted. The job stops when the number of returned records is less than the requested number of records.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDataSet.java
+++ b/engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDataSet.java
@@ -68,6 +68,8 @@ public class DataHubDataSet extends AbstractDataSet {
     private Iterator<Object[]> _resultSetIterator;
     private Row _row;
 
+    private boolean _endReached;
+
     /**
      * Constructor
      * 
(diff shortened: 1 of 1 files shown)


engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubDataSet.java
```diff
@@ -122,20 +125,22 @@ public boolean next() {
...
-        HttpGet request = new HttpGet(uri);
+        final HttpGet request = new HttpGet(uri);
         request.addHeader(ACCEPT, JSON_CONTENT_TYPE);
-        HttpResponse response = executeRequest(request);
-
-        return getResultSet(response.getEntity());
+        final HttpResponse response = executeRequest(request);
+        final List<Object[]> resultSet = getResultSet(response.getEntity());
+        final int resultSetSize = resultSet.size();
+        _endReached = (resultSetSize < maxRows);
```
kaspersorensen: Shouldn't it be ```PAGE_SIZE``` rather than ```maxRows```?

drexler42: I don't think so. Look at line 124. The maxRows is taken from the query, or becomes _nextPageMaxRows when not set in the query. By using maxRows, my intention is to also make it work when the query is setting a maxRows by itself.

kaspersorensen: OK I think you're right!

kaspersorensen:
LGTM! Merging.

PR #852, opened 2015-10-23
Feature/metering

jakubneubauer:
Basic API for components metering.

### DIFF
--- a/api/src/main/java/org/datacleaner/api/ExecutionMeteringMessage.java
+++ b/api/src/main/java/org/datacleaner/api/ExecutionMeteringMessage.java
@@ -1,0 +1,53 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
+ * copy, or redistribute it subject to the terms and conditions of the GNU
+ * Lesser General Public License, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.api;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Added by jakub on 22.10.15
+ */
+public class ExecutionMeteringMessage implements ComponentMessage {
+
+    private static Iterable<Map.Entry<String, String>> EMPTY_ITERABLE = new ArrayList<>();
+
+    private final String type;
+    private Map<String, String> additionalInfo;
+
+    public ExecutionMeteringMessage(String type) {
+        this.type = type;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public ExecutionMeteringMessage putAdditionalInfo(String key, String value) {
+        if(additionalInfo == null) {additionalInfo = new HashMap<>();}
+        additionalInfo.put(key, value);
+        return this;
+    }
+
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/datacleaner/api/ExecutionMeteringMessage.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public ExecutionMeteringMessage putAdditionalInfo(String key, String value) {
+        if(additionalInfo == null) {additionalInfo = new HashMap<>();}
+        additionalInfo.put(key, value);
+        return this;
+    }
+
+    public Iterable<Map.Entry<String, String>> getAdditionalInfo() {
+        return additionalInfo == null ? EMPTY_ITERABLE : additionalInfo.entrySet();
```
kaspersorensen: Instead of this null checking logic, why not just ensure that the map is always created in the constructor. Then you can also make it final and less prone to concurrency issues etc. And avoid the nasty EMPTY_ITERABLE which is just making the code more involved to read.

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentAware.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.datacleaner.monitor.server.components;
+
+import org.datacleaner.api.Component;
+import org.datacleaner.descriptors.ComponentDescriptor;
+
+/**
+ * Added by jakub on 23.10.15
+ */
+public interface ComponentAware {
```
kaspersorensen: What is this for? No javadoc and I cannot see the relevance for metering?

api/src/main/java/org/datacleaner/api/ExecutionMeteringMessage.java
```diff
@@ -0,0 +1,53 @@
...
+        return type;
+    }
+
+    public ExecutionMeteringMessage putAdditionalInfo(String key, String value) {
+        if(additionalInfo == null) {additionalInfo = new HashMap<>();}
+        additionalInfo.put(key, value);
+        return this;
+    }
+
+    public Iterable<Map.Entry<String, String>> getAdditionalInfo() {
```
kaspersorensen: Why this difficult return type instead of just returning a Map?

Then you can also use Collections.emptyMap() instead of EMPTY_ITERABLE if you disagree on my remark below :)

api/src/main/java/org/datacleaner/api/ExecutionMeteringMessage.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public ExecutionMeteringMessage(String type) {
+        this.type = type;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public ExecutionMeteringMessage putAdditionalInfo(String key, String value) {
```
kaspersorensen: I would discourage any mutation for a message type since these will be passed around in a highly concurrent consumer system. Rather I would expect any additional info to be provided in a constructor and made final from the start.

kaspersorensen:
One final remark (sorry I know this is a lot for such a small PR :-)): I would love to have the message class also on master branch so that it can already now be incorporated into components.

But that also means I would like to have it in a commit without anything else - so that it can be cherry picked for multiple branches. The ComponentAware stuff is problematic in that regard.

api/src/main/java/org/datacleaner/api/ExecutionMeteringMessage.java
```diff
@@ -0,0 +1,53 @@
...
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.api;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Added by jakub on 22.10.15
```
kaspersorensen: Anything in the API _must_ have proper javadoc IMO

api/src/main/java/org/datacleaner/api/ExecutionMeteringMessage.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class ExecutionMeteringMessage implements ComponentMessage {
```
kaspersorensen: "Execution" seems a bit irrelevant in this regard because everything is execution. I would rather use words that really pertain to this, or none at all. For instance: UsageMeteringMessage, ServiceMeteringMessage, ComponentUsageMessage?

jakubneubauer:
Will try it again :-)

PR #854, opened 2015-10-26
Upgraded to MetaModel 4.4.1 using the staged repository

kaspersorensen:
Do not merge yet - we would want the MM 4.4.1 release to be final first. But this PR confirms that it is building well and all that.


kaspersorensen:
Failed due to https://issues.apache.org/jira/browse/METAMODEL-204

LosD:
:+1:

PR #859, opened 2015-10-27
Added support for analysis listener in remote components (server side).

jakubneubauer:
This PR looks bigger than is in real. Some classes were made to be a Spring beans and so references are now auto-wired. The main change is in ComponentHandler - a ComponentContext is now provided and for it an AnalysisJob and ComponentJob implementation was provided.

First I wanted to avoid to use the AnalysisJob&comp. interfaces but then I decided to not introduce new types of listeners just for metering messages. Now, the AnalysisListener can be provided as spring bean and it will work also with the Remote components API. So metering handling can be done with it and it will work not only for remote components API but also for the existing job API.


monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
```diff
@@ -448,4 +487,105 @@ public void onValues(Object[] values) {
...
+        }
+
+        public <E> E getInstance(InjectionPoint<E> injectionPoint) {
+            E obj;
+            final Class<E> baseType = injectionPoint.getBaseType();
+            if (baseType == OutputRowCollector.class) {
+                obj = (E) new ThreadLocalOutputRowCollector();
+            } else if(baseType == ComponentContext.class) {
+                obj = (E) componentContext;
+            } else {
```
jakubneubauer: @kaspersorensen I would like to ask you to comment this approach. I mean with providing (injecting) the standard ComponentContextImpl for the executed component. The ComponentContextImpl is created with a AnalysisListener, which is actually a chain of listeners provided by Spring container. The same approach as is used inside the core engine.
This means more burden is needed here (implementation of the AnalysisJob and TransformerJob), but the "pro" is that the AnalysisListeners, provided as spring beans by configuration, will work as well as in the core as here in the remote components API in the same way, which I like very much.

kaspersorensen: I am not 100% sure because it's a bit far away from me, but I recall we have stuff like PlaceholderAnalysisJob and PlaceholderComponentJob. Maybe those can be reused? We need this in a number situations whenever we want to call components outside of a job, or sometimes in slightly different situations than the normal job. So maybe it would be better to reuse/enhance those? I'm affraid we start recreating them every time we find a new situation.

kaspersorensen:
This PR is very hard to decipher now that it merges in master. Maybe you can separately merge in master to the 5.0 branch. That would make the diff on this PR readable again I think.

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandler.java
jakubneubauer: PlaceholderAnalysisJob unfortunately cannot be reused nor refactored to be usable here. Here in ComponentHandler we have not "full" job - there is no real datastore, etc... So in this case I implemented some lighter AnalysisJob, using data directly from ComponentHandler...

Issue #858, opened 2015-10-27
REST client cache map (key should not be an integer)


jhorcicka:
REST client uses an internal map for storing instances for combination of username+password. Currently an integer (hash code) is used as a key which can lead to many collisions. Encoded username+password combination seems to be a better solution.

PR #860, opened 2015-10-27
Bug #858 REST client cache map

jhorcicka:
In REST client there is a cache for instances which is implemented by a map. Buggy version contained a map with an Integer as a key which was computed by hashCode() method. This could cause collisions so now SHA-1 is used for key creation.


jakubneubauer:
LGTM

Issue #867, opened 2015-10-30
Nicer remote components tab in the configuration dialog


jhorcicka:
In DC desktop options dialog for remote server configuration the description text is aligned together with the text input fields for credentials. This is not so nice.

Expected Behavior:
The text should be extended to the width as much as possible.

Observed Behavior:
The text is aligned with the input fields which creates optically one narrow column in the wider otherwise empty area.

Replication Steps:
Open options dialog in DC desktop and select remote components tab.

PR #868, opened 2015-10-30
Bug #867 Nicer remote components tab

jhorcicka:
OptionsDialog, remote components tab. Description text is spread to the full width of the panel. In the code, methods for creating the tab were moved into a separate private class in the same file.


kaspersorensen:
A process note - In cases like this it always helps with a screenshot on the PR, allowing people to quickly review the actual UI change from a user perspective.

jhorcicka:
Old version:
oldversion

New version:
newversion

kaspersorensen:
The layout change looks good to me. We should probably look more at the actual text in there too - I think it gives me a bit too technical feeling right now and I would rather focus the text on what I as a user will get by entering my credentials here - but that can wait till we get closer to launching this so that messaging is convered in a cohesive way.

+1

jhorcicka:
I fully agree. For now it enables remote components but I guess more functionality is coming.

jakubneubauer:
reviewed, fixes #867

PR #871, opened 2015-10-30
Feature/839 country standardization html rendering

### LINKED ISSUE #839: HTML rendering of Country standardizer result
We don't seem to have any HTML rendering for the Country standardizer.

I would suggest just to make a simple one with a barchart or so - nothing fancy is needed at this point.

ClaudiaPHI:
Fixes #839 
screen shot 2015-10-30 at 2 47 23 pm

### DIFF
--- a/components/standardizers/src/main/java/org/datacleaner/result/renderer/CountryStandardizationResultHtmlRenderer.java
+++ b/components/standardizers/src/main/java/org/datacleaner/result/renderer/CountryStandardizationResultHtmlRenderer.java
@@ -63,7 +63,7 @@ public class CountryStandardizationResultHtmlRenderer implements Renderer<Countr
             sb.append("<td>");
             sb.append(StringEscapeUtils.escapeHtml(category));
             sb.append("</td>");
-            final DrillToDetailsBodyElement countryFrangment = new DrillToDetailsBodyElement(createElementId(),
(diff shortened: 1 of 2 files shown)


components/standardizers/src/main/java/org/datacleaner/result/renderer/CountryStandardizationResultHtmlRenderer.java
```diff
@@ -0,0 +1,88 @@
...
+import org.datacleaner.beans.standardize.CountryStandardizationResult;
+import org.datacleaner.result.AnnotatedRowsResult;
+import org.datacleaner.result.html.DrillToDetailsBodyElement;
+import org.datacleaner.result.html.HtmlFragment;
+import org.datacleaner.result.html.SimpleHtmlFragment;
+
+@RendererBean(HtmlRenderingFormat.class)
+public class CountryStandardizationResultHtmlRenderer implements Renderer<CountryStandardizationResult, HtmlFragment> {
+
+    private final AtomicInteger _elementCounter = new AtomicInteger(0);
```
kaspersorensen: This should not be a field since that makes it shared between many renderings. I would simply make it a local variable of the render() method.

kaspersorensen:
LGTM except for the item/remark that I added.
The build failure seems to be unrelated.

kaspersorensen:
LGTM

Issue #874, opened 2015-11-02
Provide dynamic advise for components to answer the "is distributable" question


kaspersorensen:
Today we have the ```Distributed``` annotation which lets the DC framework know which components can be distributed or not. But as it turns out, it sometimes depends on configuration. For instance the "Create CSV file" analyzer is only distributed when sorting of the output file is not enabled.

We should thus create an optional interface which components can implement in order to answer whether or not they are distributable.

We've done similar things before, e.g. with ```HasLabelAdvice```. I suggest we keep to that pattern and implement an ```HasDistributableAdvice``` interface or so.

kaspersorensen:
Fixed via #882

Issue #873, opened 2015-11-02
Provide API/awareness for components about the distributed environment they might be in


kaspersorensen:
With the introduction of the Spark module we now sometimes have the components running in a cluster where they need to make certain decisions based on the cluster information.

We should provide an API for components to get information about this. I suggest to do it like this (from component perspective):

 * Leverage the existing ```ComponentContext``` interface, but add a method to get information about the cluster environment. This information-class should answer questions such as:
  * Am I running in a cluster?
  * How many partitions of data are there in distributed run?
  * What is my partition number?
  * Maybe access to the lower-level delegate as an Object that the code would have to cast (such as RDD in spark or so).
 * In the existing ```DataCleanerEnvironment``` we should probably also have some information available. Thinking more about that ... :)
  * Maybe access to the lower-level delegate as an Object that the code would have to case (such as SparkContext or so).

kaspersorensen:
Wrong reference used in the PR! Reopening.

PR #875, opened 2015-11-02
Spark create CSV file

### LINKED ISSUE #862: Optimze how data is inserted in DC-spark jobs

kaspersorensen:
Fixes #862, but with a few remarks.

 * To make this work I had to do some tricks in the spark module which makes it aware of CreateCsvFileAnalyzer. I would prefer that to have that done in a different manner where the component can do all the work and the spark module just makes available some API for the component to figure out all the details. See #873 for more details. Specifically two things are now managed in Spark which should rather be managed by the component:
  * Creation of part-files instead of a single file.
  * Writing of CSV header line
 * Furthermore I would like to raise the need for a better "is distributable" API than just the ```@Distributable``` annotation. In the case of the "Create CSV file" analyzer the distributable flag is dependent on configuration, specifically whether or not output file sorting is needed. We have some assumptions here at the moment. See #874 for more details.

### DIFF
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
@@ -41,6 +41,7 @@ import org.datacleaner.job.FilterJob;
 import org.datacleaner.job.JaxbJobReader;
 import org.datacleaner.job.OutputDataStreamJob;
 import org.datacleaner.job.TransformerJob;
+import org.datacleaner.job.builder.AnalysisJobBuilder;
 
 /**
  * A container for for values that need to be passed between Spark workers. All
@@ -61,7 +62,7 @@ public class SparkJobContext implements Serializable {
 
     // cached/transient state
     private transient DataCleanerConfiguration _dataCleanerConfiguration;
-    private transient AnalysisJob _analysisJob;
+    private transient AnalysisJobBuilder _analysisJobBuilder;
     private transient List<ComponentJob> _componentList;
 
     public SparkJobContext(JavaSparkContext sparkContext, final String dataCleanerConfigurationPath,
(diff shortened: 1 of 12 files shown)


engine/env/spark/src/main/java/org/datacleaner/spark/functions/TuplesToTuplesFunction.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.spark.api.java.function.PairFlatMapFunction;
+
+import scala.Tuple2;
+
+/**
+ * A utility {@link PairFlatMapFunction} for conversion between a
+ * {@link JavaRDD} of {@link Tuple2} into a {@link JavaPairRDD} of that same
+ * tuple type.
+ */
+public class TuplesToTuplesFunction<K, V> implements PairFlatMapFunction<Iterator<Tuple2<K, V>>, K, V> {
```
LosD: It seems to me the primary idea is wrapping iterators in iterables, and I don't gather that from neither the name nor the JavaDoc.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -101,20 +103,39 @@ private static Resource createResource(String path) {
...
+                    return jobBuilder;
                 }
             });
+            _analysisJobBuilder = jobBuilder;
+        }
+        interceptJobBuilder(_analysisJobBuilder, new AtomicInteger(0));
+        return _analysisJobBuilder;
+    }
+
+    private void interceptJobBuilder(AnalysisJobBuilder analysisJobBuilder, AtomicInteger currentComponentIndex) {
```
LosD: Not the best name here, intercepting (at least to me) is usually when catching calls of some sort, i.e. AOP, or proxying, or putting itself inside a list of callbacks.

engine/env/spark/src/main/java/org/datacleaner/spark/functions/RowProcessingFunction.java
```diff
@@ -52,11 +83,136 @@ public RowProcessingFunction(final SparkJobContext sparkJobContext) {
...
+
+        final AnalysisJob analysisJob = jobBuilder.toAnalysisJob();
+
+        final List<Tuple2<String, NamedAnalyzerResult>> analyzerResults = executePartition(inputRowIterator,
+                analysisJob);
+
+        return analyzerResults.iterator();
+    }
+
+    private void interceptComponentsBeforeBuilding(AnalysisJobBuilder jobBuilder, int partitionNumber) {
```
LosD: "intercept" maybe also doesn't make so much sense here

LosD:
Generally looks good to me, but see comments.

Also, I'm not quite sure how Spark perceives "Function", but to me, functions are small lightweight things, that has a single... Well, function, but RowProcessingFunction is quite a big bastard. :)

engine/env/spark/src/main/java/org/datacleaner/spark/functions/TuplesToTuplesFunction.java
kaspersorensen: Well actually that's not the primary idea ;-) The iterator-to-iterable conversion is just something that happens to be necasary. The main purpose of the class is to enable the conversion of a ```RDD<Tuple2<K, V>>``` into a ```PairRDD<K, V>```.

kaspersorensen:
Fixed the method naming.
Regarding the RowProcessingFunction... I get what you say but eh ... Yeah ... Don't also big functions deserve the right to call themselves functions? ;-)

engine/env/spark/src/main/java/org/datacleaner/spark/functions/TuplesToTuplesFunction.java
LosD: Hmmm... While it might allow the calling class to do that, it doesn't seem to be what _this class_ does.

Issue #878, opened 2015-11-02
QueryOptimizationAndAnalysisListeningTest failing on Mac OS?


kaspersorensen:
On my iMac at home this test almost consistently fails every time:

```
Failed tests: 
  QueryOptimizationAndAnalysisListeningTest.testScenario:91 expected:<[1]0> but was:<[]0>
```

kaspersorensen:
Upon further investigation the root cause was found: I had a custom-built but bad MetaModel build in my local maven repo. This would cause the "getExpectedRows()" call to throw an exception because the query could not run properly.

But this revealed another (smaller) issue: That in case of such failures we do not properly handle propagate it. The composite analysis listener catches the exception and only logs it. I am unsure if this is good or if it should throw the first exception it catches. What do you think @LosD ?

LosD:
It's damn hard, is what it is. :)

In the cases that made us just log it, it was some minor issue that took down the whole chain, so that was not cool. OTOH, if it makes detecting bugs harder than it needs to be, then...

kaspersorensen:
Hmm I will close this issue because the Mac OSX failure is certtainly fixed. Let's have the composite listener on our radars ... if this makes issues for us again, I think we have a reason to change it.

LosD:
:+1:

PR #879, opened 2015-11-02
Improves ability to diagnose issues like #878

kaspersorensen:
This PR addresses #878 by making the logging and assertions easier to decipher. But I guess the issue itself isn't "solveable" since it was a local build issue that would never happen except for stupid guys like me who have corrupted local MetaModel JARs :)


LosD:
LGTM!

Issue #865, opened 2015-10-29
Encoded password with prefix


jhorcicka:
In conf.xml the password for a remote server is in encoded form which is not always simply recognizable from a not encoded one. That's why there should be a clear prefix that would differentiate it. 

Current state:
oFm53KJ9K+w= is encoded "admin" but could as well be the value itself. It is hard to tell for sure. 

Suggested state:
"encoded:oFm53KJ9K+w=" for the encoded variant
"admin" for a plain text variant

kaspersorensen:
I would recommend just "enc:" as a prefix since this is used also for datastore password storage etc. Let's make sure it's as consistent as we can.

jhorcicka:
OK

PR #880, opened 2015-11-03
Bug #865 encoded password with prefix

jhorcicka:
Change of a password reading behavior (from conf.xml).  Encoded value with a prefix and a plain text value are loaded correctly. Encoded value without a prefix is incorrect.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/SecurityUtils.java
+++ b/engine/core/src/main/java/org/datacleaner/util/SecurityUtils.java
@@ -167,8 +167,7 @@ public class SecurityUtils {
     public static String decodePasswordWithPrefix(String encodedPasswordWithPrefix) {
         if (hasPrefix(encodedPasswordWithPrefix)) {
             return decodePassword(encodedPasswordWithPrefix.substring(PREFIX.length()));
-        }
-        else {
+        } else {
             return encodedPasswordWithPrefix;
         }
     }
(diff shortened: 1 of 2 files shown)


engine/core/src/main/java/org/datacleaner/util/SecurityUtils.java
```diff
@@ -165,11 +165,12 @@ public static String encodePasswordWithPrefix(String passwordInPlainText) {
...
     public static String decodePasswordWithPrefix(String encodedPasswordWithPrefix) {
-        String encodedPasswordWithoutPrefix = hasPrefix(encodedPasswordWithPrefix) ?
-                encodedPasswordWithPrefix.substring(PREFIX.length()) :
-                encodedPasswordWithPrefix;
-
-        return decodePassword(encodedPasswordWithoutPrefix);
+        if (hasPrefix(encodedPasswordWithPrefix)) {
+            return decodePassword(encodedPasswordWithPrefix.substring(PREFIX.length()));
+        }
+        else {
```
kaspersorensen: strange to have linebreak before else, but this is not a blocker.

kaspersorensen:
LGTM, but for extra sugar on top I would remove that strange linebreak ;-)

kaspersorensen:
The failing Travis build is not related. It looks good to me. Merging! :)

PR #882, opened 2015-11-03
Issue #873: Added HasDistributionAdvice interface + check in job builder

### LINKED ISSUE #873: Provide API/awareness for components about the distributed environment they might be in

kaspersorensen:
Fixes #873

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/descriptors/AnnotationBasedTransformerComponentDescriptor.java
+++ b/engine/core/src/main/java/org/datacleaner/descriptors/AnnotationBasedTransformerComponentDescriptor.java
@@ -21,6 +21,7 @@ package org.datacleaner.descriptors;
 
 import org.datacleaner.api.ComponentSuperCategory;
 import org.datacleaner.api.Distributed;
+import org.datacleaner.api.HasDistributionAdvice;
 import org.datacleaner.api.Transformer;
 import org.datacleaner.components.categories.TransformSuperCategory;
 import org.datacleaner.util.ReflectionUtils;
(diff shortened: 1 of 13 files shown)


ClaudiaPHI:
LGTM!

engine/core/src/main/java/org/datacleaner/descriptors/AnnotationBasedTransformerComponentDescriptor.java
```diff
@@ -62,6 +63,9 @@ public boolean isDistributable() {
         if (annotation != null) {
             return annotation.value();
         }
+        if (ReflectionUtils.is(getComponentClass(), HasDistributionAdvice.class)) {
```
LosD: This code makes no sense. ReflectionUtils.is(xxx) then it will return true, otherwise it will return true?

Issue #855, opened 2015-10-27
Remote service gets quotes around enum values


khouzvicka:
## Description

When access a remote service that returns an enum, then in the results, the enum values will appear in quotes always. So it says "GREEN" and "ORANGE" instead of just GREEN or ORANGE. 
 This breaks the remote abstraction (i.e replacing a local component by a remote component will lead to unexpected results if filters are present that work on the enums value).

## Expected Behavior

Enum values are shown without quotes.

## Observed Behavior

They have quotes.

Jira: DCDQ-240

PR #885, opened 2015-11-04
Feature/855 bug remote service quotes around enum values

khouzvicka:
Fix for bug #855 - remote service quotes around enum values

### DIFF
--- a/engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
+++ b/engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
@@ -150,6 +150,9 @@ public class RemoteTransformer extends BatchRowCollectingTransformer {
     }
 
     private boolean isOutputColumnEnumeration(JsonSchema schema) {
+        if(schema == null){
+            return false;
+        }
         boolean isArray = schema.isArraySchema();
         JsonSchema baseSchema;
         if (isArray) {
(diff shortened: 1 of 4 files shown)


engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -120,9 +125,13 @@ public OutputColumns getOutputColumns() {
                         try {
                             outCols.setColumnType(i, Class.forName(colSpec.type));
                         } catch (ClassNotFoundException e) {
-                            // TODO: what to do with data types - classes that are not on our classpath?
-                            // Provide it as pure JsonNode?
-                            outCols.setColumnType(i, JsonNode.class);
+                            Class type;
+                            if (isOutputColumnEnumeration(colSpec.schema)) {
+                                type = EnumerationValue.class;
```
kaspersorensen: Shouldn't we rather return the enum values as strings? After all the client side cannot use any kind of type-based mapping of columns when the EnumerationValue class is used (because it is non-concrete). At least he can do string based mapping of columns instead.

khouzvicka: You are right. The String will be better. I will do changes, and I will do new pull request.

khouzvicka: I changed the type.

khouzvicka:
I've updated the PR, can you please review again?
:-)

kaspersorensen:
LGTM

PR #888, opened 2015-11-05
Feature/spark runner improvements

kaspersorensen:
Found various issues in the spark module which is fixed here.

 * ExampleLaunch better spec'ed with javadoc and SPARK_HOME setting
 * Dependency cleanup to make jar-with-dependencies sufficient for running most jobs.
 * Added yarn-site.xml file builder
 * Improved logging.


ClaudiaPHI:
LGTM!

PR #889, opened 2015-11-05
Spark execution properties

kaspersorensen:
Added the ability to specify execution properties and system properties for the job and configuration file loading, when running on Spark.

Since the command line length when running Spark jobs is important to keep low (we have seen issues because of long filenames already) I decided to include this using a properties file instead of specifying all properties on the command line directly.


ClaudiaPHI:
LGTM!

kaspersorensen:
Please wait with merging - I'd like to do a bit of monkey testing first (and maybe unittesting as well)

kaspersorensen:
Totally works in a monkey test too :-)

PR #890, opened 2015-11-05
Feature/887 restricted functionality exception

### LINKED ISSUE #887: Exception(s) for commercial extensions when a restricted function is attempted
We would like to make the licensing model of commercial extensions for DataCleaner easier to manage and to give the user meaningful warnings and blocking statements when he is trying to do something he cannot do. This could be for instance overspending credits or using functionality that is otherwise restricted.

Notes:

 * Validation time - a special exception type for invalid configurations etc.
 * Exception payload should include optionally link to purchase site and of course a message to the user
 * This somehow relates a bit to the payment aspects of #170

kaspersorensen:
Fixes #887 

An important aspect of this API addition is how components communicate "call to actions" from the component level to the end user. So it's important that we review the API here thoroughly - does it cover the stuff that we need, does it restrict us in any (bad) way? 

I would be particularly happy if @stefanjanssen could take a look at this since I know he wants to use the API. Stefan, would this be a reasonable API for you? You don't need to think too much about the UI side of it since that is rudimentary.

### DIFF
--- a/api/src/main/java/org/datacleaner/api/RestrictedFunctionalityCallToAction.java
+++ b/api/src/main/java/org/datacleaner/api/RestrictedFunctionalityCallToAction.java
@@ -21,8 +21,6 @@ package org.datacleaner.api;
 
 import java.io.Serializable;
 
-import com.google.common.base.Strings;
-
 /**
  * Represents a "Call to Action" for users when restricted functionality is
  * blocked, typically with the use of {@link RestrictedFunctionalityException}
@@ -62,10 +60,10 @@ public class RestrictedFunctionalityCallToAction implements Serializable {
      *            tooltip, a smaller-font text or so
      */
     public RestrictedFunctionalityCallToAction(String name, String href, String description) {
-        if (Strings.isNullOrEmpty(name)) {
+        if (name == null || name.isEmpty()) {
             throw new IllegalArgumentException("Action name cannot be null");
         }
-        if (Strings.isNullOrEmpty(href)) {
+        if (href == null || href.isEmpty()) {
             throw new IllegalArgumentException("Action href cannot be null");
         }
         _name = name;
--- a/api/src/main/java/org/datacleaner/api/RestrictedFunctionalityCallToAction.java
+++ b/api/src/main/java/org/datacleaner/api/RestrictedFunctionalityCallToAction.java
@@ -1,0 +1,104 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
+ * copy, or redistribute it subject to the terms and conditions of the GNU
+ * Lesser General Public License, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.api;
+
+import java.io.Serializable;
+
+import com.google.common.base.Strings;
+
+/**
+ * Represents a "Call to Action" for users when restricted functionality is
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/datacleaner/api/RestrictedFunctionalityException.java
```diff
@@ -0,0 +1,79 @@
...
+     * Constructs a {@link RestrictedFunctionalityException}
+     * 
+     * @param message
+     *            a message to the user about what and why he is being
+     *            restricted in functionality.
+     * @param callToActions
+     *            an array of call to actions for the user to pick from.
+     */
+    public RestrictedFunctionalityException(String message, RestrictedFunctionalityCallToAction... callToActions) {
+        super(message);
```
stefanjanssen: There is also a constructor that explicitly disables stracktraces, and given we are dealing with licensing code, I would suggest using that one.

kaspersorensen: Good idea - fixed

LosD:
LGTM

api/src/main/java/org/datacleaner/api/RestrictedFunctionalityCallToAction.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Represents a "Call to Action" for users when restricted functionality is
+ * blocked, typically with the use of {@link RestrictedFunctionalityException}
+ * or {@link RestrictedFunctionalityMessage}.
+ */
+public class RestrictedFunctionalityCallToAction implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private final String _name;
```
stefanjanssen: Right now  this is a string, but I think we could change it to an enum. The reasoning being that if we catch this in the UI, we probably want to show an advert specific towards either a buy credits page or a upgrade to enterprise page. Right now we would have to do with a generic popup and a URL. If the goal is to motivate users into actions, I think we need something more specific.

The other thing is that the URL right now has to be coded in the client, so if this url ever changes, we have to go through every transformer/analyzer and change it to the new one.

The description I would keep, so that the component can specify the details a bit more, eg. ""name cleansing requires x credits", "personal edition of duplicate detection is limited to x records".

stefanjanssen:
Well other than my remarks, looks good to me. Obviously user experience is still lacking, but I expect there will be improvements around that later.

api/src/main/java/org/datacleaner/api/RestrictedFunctionalityCallToAction.java
kaspersorensen: I was thinking something along the same line when I created it also ... Hmm maybe a solution could be to have a subclass for our most typical restrictions. That subclass could have the standard texts in there as well. Because I do see one good thing about the open-ended string which is that it will allow other vendors to make use of this construction too. This could be our sister companies within Neopost or even third parties.

stefanjanssen: Subclasses could also work too, instead of enums. Under normal situations, the only ones I can think of we would use right now are "upgrade to edition X" and "buy credits". Maybe in the future we would have something for user to buy premium versions of certain transformers, eg app store like. The most extreme scenario would probably be a third party reinventing the entire credit system (although they would probably just create a fork in that case).

Maybe turn it in an interface then and have implementing classes for the two scenarios we need right now?

kaspersorensen: Unfortunately throwable is a class so we cannot make interfaces.

Regarding third parties - I could imagine a less extreme case where they just build an extension and put it on the extension swap. The extension might have certain paid-for features or it's own license structure etc. That's quite common in many popular communities, and we should hope to be that popular too some day ;-)

stefanjanssen: I meant making RestrictedFunctionalityCallToAction an interface, that should be possible?

You could also go for an enum and then extend later extend it with the app store kind of option (so credits/dc edition/extension store). Also maybe for an app store option you need the url to be overrideable, so that it could point to a specific extension on the app store. That should cover pretty much all cases in the foreseeable future. 

I don't think we should make things more complicated than they need to be just for some extreme theoretical scenario.

kaspersorensen: I would agree if we knew more about it at this point, but I think we don't. For instance I could imagine that the Duplicate detection extension could link not just to the "upgrade editions" page but it could also be that there's a "read more" call to action which links to http://datacleaner.org/focus/dedup or whatever.

PR #891, opened 2015-11-06
Feature/metering api

jakubneubauer:


api/src/main/java/org/datacleaner/api/UsageMeteringMessage.java
```diff
@@ -53,12 +71,22 @@ public String getType() {
...
-        return "UsageMeteringMessage[" + type + ", " + additionalInfo + "]";
+        return "UsageMeteringMessage[" + type + " " + details+ "]";
     }
+
+    /** Escapes value for a CSV line */
+    private String escapeValue(String field) {
+        if(field.indexOf(DETAILS_SEPARATOR_CHAR) == -1 && field.indexOf(DETAILS_QUOTE_CHAR) == -1) {
+            return field;
+        }
+        return DETAILS_QUOTE_CHAR + field.replace(DETAILS_QUOTE_STR, DETAILS_TWO_QUOTES_STR) + DETAILS_QUOTE_CHAR;
```
kaspersorensen: String.replace only replaces one time. We should probably use replaceAll to avoid multiple things to be escaped.

api/src/main/java/org/datacleaner/api/UsageMeteringMessage.java
```diff
@@ -20,26 +20,44 @@
...
-public class UsageMeteringMessage implements Serializable {
+public class UsageMeteringMessage implements Serializable, ComponentMessage {
+
+    private static final long serialVersionUID = 42L;
+
+    public static final char DETAILS_SEPARATOR_CHAR = ',';
+    public static final char DETAILS_QUOTE_CHAR = '"';
+
+    private static final String DETAILS_QUOTE_STR = String.valueOf(DETAILS_QUOTE_CHAR);
+    private static final String DETAILS_TWO_QUOTES_STR = DETAILS_QUOTE_STR + DETAILS_QUOTE_STR;
```
kaspersorensen: Normally our default is to use backslash as the escape char. So I would prefer if we add an DETAILS_ESCAPE_CHAR constant and use that here.

api/src/main/java/org/datacleaner/api/UsageMeteringMessage.java
jakubneubauer: No, String.replace "Replaces each substring..." (citing javadoc)

api/src/main/java/org/datacleaner/api/UsageMeteringMessage.java
jakubneubauer: But it will not be CSV then. CSV escapes quotes by doubling it - see https://tools.ietf.org/html/rfc4180#section-2 point 7.

api/src/main/java/org/datacleaner/api/UsageMeteringMessage.java
kaspersorensen: You're right. Ouch, this means I've made some pretty stupid while-loops in my past ... Whatever, good to know!

api/src/main/java/org/datacleaner/api/UsageMeteringMessage.java
kaspersorensen: They put it well - "there is no formal specification in existence, which allows for a wide variety of interpretations of CSV files." - which basically means it's just a matter of following _some_ standard. Could we for the moment settle on the MetaModel standard? Maybe you can refer to MetaModel's ```CsvConfiguration.DEFAULT_ESCAPE_CHAR``` (which we might change in the future, although it would be a pretty "surprisey" thing to do for people who rely on that constant).

LosD: CSV predates IETFs attempts at standardizing it by decades. :)

jakubneubauer: Ok, let's do it consistently with metamodel, although it is surely a less frequent understanding of CSV.
And is there a need to encapsulate the value in double quotes, then? If the quotes in the value are escaped by backslash?

kaspersorensen: I would agan then save us from a discussion about what is best by just pointing to what we normally do. I would use ```CsvConfiguration.DEFAULT_QUOTE_CHAR``` and  ```CsvConfiguration.DEFAULT_SEPARATOR_CHAR``` :-)

jakubneubauer:
Hopefully ready to merge

kaspersorensen:
LGTM!

PR #900, opened 2015-11-09
Feature/spark json parser

### LINKED ISSUE #898: Json Files on hdfs

ClaudiaPHI:
Fixes #898 

-Works with json file in one line and with normal json files.


engine/env/spark/src/main/java/org/datacleaner/spark/functions/ExtractJsonInputRows.java
```diff
@@ -0,0 +1,59 @@
...
+        _jsonDatastore = jsonDatastore;
+    }
+
+    public List<Object[]> getInputRows() {
+        final List<Object[]> list = new ArrayList<>();
+        final JsonDataContext jsonContext = new JsonDataContext(_jsonDatastore.getResource());
+        final Table table = jsonContext.getDefaultSchema().getTable(0);
+        final Column[] columns = table.getColumns();
+        final Query query = new Query().select(columns).from(table);
+        final DataSet executeQuery = jsonContext.executeQuery(query);
```
kaspersorensen: This will make MetaModel read the file on HDFS from end to end. That basically means we don't get any scaling behaviour because the read is not distributed. Worse, it also means that every node will do the exact same thing, creating duplicated results.

Let's talk about this on Skype, but in short you have to use the SparkContext to read the text file line-by-line and pass each line further into the processing steps. If you don't do that you also build up the complete file in memory which isn't feasible either.

kaspersorensen:
The tests here contains multi-line JSON files which was explicitly not selected as part of the story. I see now that here it seems possible, but there's a big reason why it was explicitly not included - that it's not possible to allow parallel reading of the file if the lines aren't independent data items.

engine/env/spark/src/main/java/org/datacleaner/spark/functions/JsonParserFunction.java
```diff
@@ -0,0 +1,73 @@
...
+            final String value = node.asText();
+            list[i] = value;
+            i++;
+        }
+        return list;
+    }
+
+    public Column[] getColumns() {
+        if (_columns == null) {
+            final JsonDataContext jsonContext = new JsonDataContext(_jsonDatastore.getResource());
```
kaspersorensen: I think it would be a little clearer to just open a connection to the datastore (and thus let it instantiate the DataContext).

engine/env/spark/src/main/java/org/datacleaner/spark/functions/JsonParserFunction.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public Object[] call(String line) throws Exception {
+        final ObjectMapper mapper = new ObjectMapper();
+        final JsonNode readTree = mapper.readTree(line);
+        final Object[] inputRows = getInputRows(readTree);
+        return inputRows;
+    }
+
+    private Object[] getInputRows(JsonNode readTree) {
```
kaspersorensen: Method name a bit strange - get values?

engine/env/spark/src/main/java/org/datacleaner/spark/functions/JsonParserFunction.java
```diff
@@ -0,0 +1,73 @@
...
+        final Column[] columns = getColumns();
+        final Object[] list = new Object[columns.length];
+        final Iterator<JsonNode> mainIterator = readTree.iterator();
+        int i = 0;
+        while (mainIterator.hasNext()) {
+            final JsonNode node = mainIterator.next();
```
kaspersorensen: There's a potential issue here - the values might not come always in the same order. Imagine for instance:

```
{"name":"Claudia", "company":"HI"}
{"company":"HI", "name":"Kasper"}
```

So I guess we need to resolve values based on an iteration of column (names) and insert them in the values array according to the sequence of columns.

engine/env/spark/src/test/resources/conf_hdfs.xml
```diff
@@ -1,11 +1,12 @@
...
-<configuration xmlns="http://eobjects.org/analyzerbeans/configuration/1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<configuration xmlns="http://eobjects.org/analyzerbeans/configuration/1.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
 	<datastore-catalog>
-		<csv-datastore name="person_names">
-			<filename>hdfs://bigdatavm:9000/datacleaner/test/person_names.txt</filename>
-			<multiline-values>false</multiline-values>
-		</csv-datastore>
+		<json-datastore name="person_data">
```
kaspersorensen: Why replace the old one that was working well? Would be nicer to create a new one so that we can still keep using the old also.

ClaudiaPHI: I was thinking you might want to check out the branch and ran it.

ClaudiaPHI: It's pretty annoying to change settings all the time

kaspersorensen: But we're not restricted to having only one datastore?! Just create another one and make a job that refers to that?

ClaudiaPHI:
I reversed to vanilla job. Ready for merge ?

kaspersorensen:
I meant to keep both! Append a datastore instead of replacing. 

But yes I think it's fine either way - you're very welcome to add the JSON datastore.

ClaudiaPHI:
I have added the json datastore to conf_hdfs. Push merge ?

kaspersorensen:
yup! :)

ClaudiaPHI:
,

PR #903, opened 2015-11-11
Spark CSV write speed fixed by buffering and not needlessly checking isExists()

### LINKED ISSUE #902: Very poor write-performance of "Create CSV file" when executing in spark module
Although it "scales out" the performance of individual executor nodes is certainly wrong. We observed 6-8 records written per second. Strangely, it seems to write a lot faster when running via DC desktop (also to HDFS resource).

kaspersorensen:
Fixes #902

A few optimizations

 * Added buffering to the output stream (this did the trick)
 * Avoiding needless call to HdfsResource.isExists() - this might save some time because it happens again and again.
 * Avoiding re-reading the files on HDFS again and again by saving them as Strings in SparkJobContext instead. This means only the driver program needs to read them.

### DIFF
--- a/components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
+++ b/components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
@@ -142,7 +142,7 @@ public class CreateCsvFileAnalyzer extends AbstractOutputWriterAnalyzer implemen
 
     @Validate
     public void validate() {
-        if (file.isExists() && !overwriteFileIfExists) {
+        if (!overwriteFileIfExists && file.isExists()) {
             throw new IllegalStateException(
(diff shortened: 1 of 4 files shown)


components/writers/src/main/java/org/datacleaner/extension/output/CreateCsvFileAnalyzer.java
```diff
@@ -142,7 +142,7 @@ public String getSuggestedLabel() {
 
     @Validate
     public void validate() {
-        if (file.isExists() && !overwriteFileIfExists) {
+        if (!overwriteFileIfExists && file.isExists()) {
```
LosD: A lot cheaper, I guess?

LosD:
LGTM! Even saved quite a few lines of code

LosD:
(Waiting for Travis)

ClaudiaPHI:
LGTM!

PR #907, opened 2015-11-16
based64 component images for html output

### LINKED ISSUE #838: HTML output should generate component icons instead of relying on copies
The HTML rendering output currently relies on png files in a static directory for the icons of components.

Rather we should use the same approach as we do in the Component Library documentation where we create base64 encoded images that are embedded into the HTML report. That will make us much less dependent on external resources.

ClaudiaPHI:
- [x] A part of the documentation component builder code has been moved into html-rendering 
 project so that the code(for component library) can be reused. Note:desktop-ui cannot be imported in html-rendering due to cyclic dependencies. 
- [x] The component images have been imported as base64 into html.  
Fixes #838


ClaudiaPHI:
screen shot 2015-11-16 at 2 58 47 pm

screen shot 2015-11-16 at 2 58 47 pm

components/html-rendering/src/main/java/org/datacleaner/documentation/ComponentDocumentationWrapper.java
```diff
@@ -148,6 +149,9 @@ public String getIconSrc() throws IOException {
         return iconHtmlRepresentation;
     }
 
+    public int getIconSrcLarge(){
```
LosD: This should probably be ```getIconSizeLarge()```

LosD:
Except for a single naming issue, code LGTM.

I'm not totally sure about the moving stuff around. It's a little weird with documentation code in the result renderer code, but OTOH there is certainly relation there. Maybe some renaming when it's also used for non-documentation purposes? Let's see what @kaspersorensen says.

kaspersorensen:
Code looks alright, but I'm not sure the freemarker changes are valid freemarker syntax. Have you verified that the icons in the component documentation are still rendered correctly (remember to delete cached versions in §userhome/.datacleaner directory)?

ClaudiaPHI:
Yes. freemarker changes works. I have to change the method name and then I will double check it again.

ClaudiaPHI:
Ready to merge.

kaspersorensen:
Now I found the css/png files that I mentioned (offline) before:

In this repo: desktop/ui/src/test/resources/styling
In commercial repo: monitor-base-webapp/src/main/webapp/monitor-report-styling

ClaudiaPHI:
Do you want me to remove that unnecessary styling from these files  ? just to be sure ?

kaspersorensen:
Yep - thanks. Will merge.

PR #912, opened 2015-11-16
Issue #904: Added chart thresholds for Pattern Finder and Value Dist

### LINKED ISSUE #904: Prevent rendering chart with extreme no. of slices
I am working right now on profiling a column which contains tens of thousands of distinct values. I didn't realize that and tried to use "Value distribution" on it. Now what I see is just a "loading" spinner in the UI where the value distribution report would normally be. And in DC's log file it says:

```
INFO  08:33:34.008 [SwingWorker-pool-2-thread-10] ValueDistributionResultSwingRendererGroupDelegate - Rendering with 31746 slices
```

After waiting for a looong time I finally got a completely useless chart because every pixel would represent multiple slices anyway. I suggest that when we're above some (rather high) threshold, maybe 2000 or so, then we shouldn't show a chart at all since it will never ever make sense to look at anyway.

kaspersorensen:
Fixes #904 

Instead of a chart + table, the results now only contain a table.
One minor (unintended) drawback of these result displays is that they don't include the summary statistics which the chart _also_ showed - The number of distinct values/patterns.


ClaudiaPHI:
LGTM!

PR #913, opened 2015-11-17
Wraps drawing calls in ready handler.

### LINKED ISSUE #840: HTML rendered Value Distribution result bar charts are missing
I just did a "What's in my data" analysis on orderdb CUSTOMERS table. And then exported the result as HTML.

For some reason the value distribution result did not have any charts.

LosD:
Also updates Flotto 0.8.3

There seem to be an minor issue with the sizing of the graphs, they all need to scroll a few millimeters. However, I don't think it's related to this (it's the same issue with Flot 0.7), so I left it alone for now.

Fixes #840


LosD:
... Aaaaaand first comment is from myself. Looking at the diff, the indentation seems a bit off.

LosD:
Ugh, and stupid toString() tests are failing.

kaspersorensen:
LGTM. Happy to merge?

LosD:
You're happy, Travis is happy, I'm happy :)

However, I'm thinking we might want to make sure I didn't destroy anything for IE or FF (AFAIK it worked on them, so that would be a step backwards). I'll make a final test tomorrow (maybe we could get @ClaudiaPHI and her nice VBox instances to see how the older IEs react).

LosD:
Newest Firefox is happy, and IE 11 is happy.

PR #916, opened 2015-11-17
Adds right-click menu to directories so they can be selected

### LINKED ISSUE #915: Enable selecting an HDFS directory with part-files instead of only files
We need a way to select directories full of part files, and not only files in the desktop HDFS browser.

LosD:
This is an attempt to make it possible to select directories, by making a context menu. I'm not sure it's the best way, as it has some discoverability issues, but it's a simple way.

Unfortunately, I cannot test it, as I have some issues creating my vagrant VMs (for getting access to an HDFS), so there's currently a good chance this doesn't work at all.

Fixes #915


LosD:
Ouch. The chooser seems to work fine, but the ones using the chooser certainly doesn't expect a directory.

LosD:
No, that's actually fine. As long as it is a directory named as the expected filetype, (i.e. ```test.csv```).

The error isn't very nice (a popup with a stack trace), but it's the same for "normal" files.

kaspersorensen:
LGTM!

PR #917, opened 2015-11-18
Feature/905 gui element component list refresh

jhorcicka:
New item in "More" menu: 
menuitemdc

A loading bar to notify a user that something is happening, when the refresh is taking some time. Please, correct me here if I am going an undesired way (as for UX, functionality or whatever). Perhaps, there already is some other widget/tool for this purpose (which I did not find): 
loadingbardc


kaspersorensen:
I think this function is a bit too particular to put in the "More" menu. How about having it as a right-click option for the "Library" node in the tree itself?

jhorcicka:
Alright. ;-)

jhorcicka:
I removed the "Refresh component list" menu item in "More" menu and the Spinner as well. 

Now there is the popup menu available over "Library".

PR #920, opened 2015-11-18
Feature/914 save job result on hdfs

### LINKED ISSUE #914: Save result file of DataCleaner-spark jobs on HDFS
Save result file of DataCleaner-spark jobs on HDFS, maybe in the folder: /datacleaner/results.

ClaudiaPHI:
Fixes #914 

- The user has the option to define a path for the result in properties file. The property name is: datacleaner.result.hdfs.path  
- The filename template: jobname-timestamp.analysis.result.dat
- By default the result will be written to hdfs://hostname:port/datacleaner/results/

### DIFF
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
@@ -26,6 +26,8 @@ import java.util.Map;
 
 import org.apache.commons.lang.SerializationException;
 import org.apache.commons.lang.SerializationUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.log4j.Logger;
 import org.apache.metamodel.util.FileHelper;
 import org.apache.metamodel.util.HdfsResource;
 import org.apache.spark.SparkConf;
@@ -36,6 +38,10 @@ import org.datacleaner.job.runner.AnalysisResultFuture;
 import org.datacleaner.result.SimpleAnalysisResult;
 
 public class Main {
+
+    static Logger logger = Logger.getLogger(Main.class);
+
+    private static String DEFAULT_RESULT_PATH = "/datacleaner/results/";
 
     public static void main(String[] args) {
         if (args.length < 2) {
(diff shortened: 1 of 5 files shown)


engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -39,20 +56,53 @@ public static void main(String[] args) {
...
         try {
-            sparkAnalysisRunner.run();
+            final AnalysisResultFuture result = sparkAnalysisRunner.run();
+            if (result.isDone()) {
+                final Map<ComponentJob, AnalyzerResult> resultMap = result.getResultMap();
+                final SimpleAnalysisResult simpleAnalysisResult = new SimpleAnalysisResult(resultMap,
+                        result.getCreationDate());
+                final String resultJobFilePath = getResultJobFilePath(sparkContext, sparkJobContext);
+                if (resultJobFilePath != null) {
+                    final HdfsResource hdfsResource = new HdfsResource(resultJobFilePath);
```
LosD: Is there a guarantee that the resultJobFilePath is an HDFS path?

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -39,20 +56,53 @@ public static void main(String[] args) {
...
+                    final OutputStream out = hdfsResource.write();
+                    try {
+                        SerializationUtils.serialize(simpleAnalysisResult, out);
+                    } catch (SerializationException e) {
+                        throw e;
+                    } finally {
+                        FileHelper.safeClose(out);
```
LosD: It may be better to use try-with-resource here.

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
LosD: Oh, I guess HDFSResource would still handle it if it weren't, given that it's always parsable by the FS. Ignore :)

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -39,20 +56,53 @@ public static void main(String[] args) {
...
+                    }
+                }
+            }
         } finally {
             sparkContext.stop();
         }
     }
+
+    private static String getResultJobFilePath(final JavaSparkContext sparkContext, final SparkJobContext sparkJobContext) {
+        String resultPath = sparkJobContext.getResultPath();
```
LosD: Would it be feasible (or desirable) to validate the path is on the current cluster?
What happens if the path is relative? Does HDFS / HDFSResource take care of it?

ClaudiaPHI: If the path is relative then it will throw an exception while trying to write the job.

LosD: Hmmm... Should we consider prepending the server host unto relative paths (or, I guess they are actually absolute, just without a host)?

ClaudiaPHI: Everything path set in job files, configurations is in absolute path. The user has the option of writing a path in the job properties file. Shouldn't we keep the same pattern with an absolute path?  Why allow relative paths ?

LosD: For absolute, but host-less paths (i.e. ```/path/to/file```), it would be good because the path should more or less always be relative to the current cluster's namenode. A path with a full host of another namenode would usually be an error (unless the cluster is using federation).

For "true" relative paths (i.e. ```path/to/file```), it could be a good way to ensure it is still within the default datacleaner folder.

LosD: Anyway, maybe it isn't worth it, but I believe it would be a pretty simple change.

ClaudiaPHI: ok. I will change with a absolute path, relative to the host and test it.

LosD: Cool. If it proves more troublesome than I thought, I guess we can skip it.

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -39,20 +56,61 @@ public static void main(String[] args) {
...
+
+    private static String getResultJobFilePath(final JavaSparkContext sparkContext,
+            final SparkJobContext sparkJobContext) {
+        String resultPath = sparkJobContext.getResultPath();
+        final Configuration hadoopConfiguration = sparkContext.hadoopConfiguration();
+        final String fileSystemPrefix = hadoopConfiguration.get("fs.defaultFS");
+        if (resultPath == null) {
+            resultPath = fileSystemPrefix + DEFAULT_RESULT_PATH;
+        } else {
+            resultPath = fileSystemPrefix + resultPath;
```
LosD: Hmmm... Wont this break with URIs containg the host now (these may be needed by customers using [federated namenodes](https://hadoop.apache.org/docs/r2.4.1/hadoop-project-dist/hadoop-hdfs/Federation.html))? 

(It is pretty simple to test if it does, using the URI class' isAbsolute() method)

LosD:
LGTM

PR #921, opened 2015-11-18
Always show link target component dialog

### LINKED ISSUE #908: Clicking "Link [stream] to..." should show configuration dialog of target component
When creating a link from a output data stream producer to a consumer of the stream, the auto-mapping of input should also be aided by showing the dialog of the target component.

For instance: When linking "duplicate records" from "Duplicate detection" to a "Merge duplicates" component. Nothing happens in this case.

LosD:
A user should always see the component dialog when trying to link dialogs, even if there are no columns to map.

Fixes #908


ClaudiaPHI:
LGTM this one liner :-D

LosD:
Cool, I guess we can merge when Travis is happy :)

PR #926, opened 2015-11-19
Try to update scope for filters

### LINKED ISSUE #909: Component scope not changed when creating link from a scoped filter
When "Link to" is used to wire a transformer and another component, then we automatically ensure that the target component gets the same scope as the originating transformer.

The same should be the case for filters - that if I have e.g. an Equals filter within a child-job-builder, and I use "Link to" to wire it to a target component, then the target component's scope should be changed to match the filter's.

LosD:
Filter requirements are now treated like other components, so a component will have its scope changed if a filter from another scope is linked to it.

This also adds a general confirmation for any scope changes that will destroy user configuration, as well as remove a bit of duplication.

Fixes #909

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLinkPainter.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLinkPainter.java
@@ -106,7 +106,7 @@ public class JobGraphLinkPainter {
 
     /**
      * Called when the drawing of a new link/edge is started
-     * 
+     *
      * @param startVertex
      */
     public void startLink(VertexContext startVertex) {
@@ -140,7 +140,7 @@ public class JobGraphLinkPainter {
     /**
      * If startVertex is non-null this method will attempt to end the
      * link-painting at the given endVertex
-     * 
+     *
      * @return true if a link drawing was ended or false if it wasn't started
      */
     public boolean endLink(Object endVertex, MouseEvent mouseEvent) {
@@ -224,8 +224,8 @@ public class JobGraphLinkPainter {
 
         if (toVertex instanceof ComponentBuilder) {
             final ComponentBuilder componentBuilder = (ComponentBuilder) toVertex;
-            if (sourceColumns != null && !sourceColumns.isEmpty()) {
 
+            if (sourceColumns != null && !sourceColumns.isEmpty()) {
                 if (componentBuilder.getDescriptor().isMultiStreamComponent()) {
                     if (!fromVertex.getAnalysisJobBuilder().isRootJobBuilder()) {
                         // we don't yet support MultiStreamComponents on output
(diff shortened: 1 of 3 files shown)


ClaudiaPHI:
LGTM!

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLinkPainter.java
```diff
@@ -289,6 +295,36 @@ public void actionPerformed(ActionEvent e) {
...
+        if (sourceAnalysisJobBuilder != componentBuilder.getAnalysisJobBuilder()) {
+            if (componentBuilder.getInput().length > 0 || componentBuilder.getComponentRequirement() != null) {
+                final String scopeText;
+                if(sourceAnalysisJobBuilder.isRootJobBuilder()){
+                    scopeText = " into the default scope";
+                } else {
+                    scopeText = " into the scope " + sourceAnalysisJobBuilder.getDatastore().getName();
+                }
+                final int response = JOptionPane.showConfirmDialog(_graphContext.getVisualizationViewer(),
+                        "This will move " + componentBuilder.getDescriptor().getDisplayName() + scopeText
```
kaspersorensen: Instead of .getDisplayName() let's use LabelUtils.getLabel(componentBuilder). That will respect also if the user has done any renaming etc.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLinkPainter.java
```diff
@@ -289,6 +295,36 @@ public void actionPerformed(ActionEvent e) {
...
+     */
+    private boolean scopeUpdatePermitted(final AnalysisJobBuilder sourceAnalysisJobBuilder,
+            final ComponentBuilder componentBuilder) {
+        if (sourceAnalysisJobBuilder != componentBuilder.getAnalysisJobBuilder()) {
+            if (componentBuilder.getInput().length > 0 || componentBuilder.getComponentRequirement() != null) {
+                final String scopeText;
+                if(sourceAnalysisJobBuilder.isRootJobBuilder()){
+                    scopeText = " into the default scope";
+                } else {
+                    scopeText = " into the scope " + sourceAnalysisJobBuilder.getDatastore().getName();
```
kaspersorensen: Talking about labels - maybe we should add a LabelUtils.getScopeLabel(AnalysisJobBuilder) which takes care of the scopeText creation here.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLinkPainter.java
LosD: :+1:

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphLinkPainter.java
LosD: Hmmm... I'm not sure the text is general enough (notice the difference between "into the default scope" and "into the scope xxxxx"). I first used DEFAULT_SCOPE_TEXT from ScopeMenuBuilder, but it seemed a little silly to have half the text from there, while assembling the rest manually (and it's using capital D, which isn't really warranted here).

kaspersorensen: Maybe it could produce either "default scope" or "scope XYZ"?
But if not, no biggie.

LosD: Hmmm... Let's do it. It's simple, so there's almost no loss if it doesn't get reused.

kaspersorensen:
I'm gonna ask Travis for a new build ...

LosD:
And now Travis is happy. We might want to look into those, though. We see these a bit too often for comfort.

kaspersorensen:
Agreed. Plan for next sprint?
Merging this.

LosD:
Yeah

PR #934, opened 2015-11-19
made the library to included in the schema tree optional.

### LINKED ISSUE #929: Datastore dictionary dialog has a schema tree - including the component library
Eh, this schema tree here should only have the datastore showing - NOT the library and all it's children!

image

ClaudiaPHI:
Fixes #929. 

- made the library to included in the schema tree optional.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
@@ -88,8 +88,8 @@ import org.slf4j.LoggerFactory;
 
 import com.google.inject.Injector;
 
-public class SchemaTree extends JXTree
-        implements TreeWillExpandListener, TreeCellRenderer, ComponentDescriptorsUpdatedListener {
+public class SchemaTree extends JXTree implements TreeWillExpandListener, TreeCellRenderer,
+        ComponentDescriptorsUpdatedListener {
 
     private static final long serialVersionUID = 7763827443642264329L;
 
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
```diff
@@ -670,4 +674,9 @@ public void filter(String searchTerm) {
             }
         }
     }
+
+    public void setIncludeLibraryNode(boolean includeLibraryNode) {
+        _includeLibraryNode = includeLibraryNode;
+    }
```
LosD: Hmmm... I'm not entirely sure that this is the correct/best way when we're using Guice for injection. I think you're supposed to configure the injector, but honestly I've managed to avoid needing to know until now. :D

@kaspersorensen maybe we should get a crash course in Guice, and especially in the way we use it?

desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
```diff
@@ -108,13 +108,14 @@
...
     private final InjectorBuilder _injectorBuilder;
+    private boolean _includeLibraryNode = true;
 
     private DatastoreConnection _datastoreConnection;
     private String _searchTerm = "";
 
     @Inject
     protected SchemaTree(final Datastore datastore, @Nullable AnalysisJobBuilder analysisJobBuilder,
-            WindowContext windowContext, InjectorBuilder injectorBuilder) {
+            WindowContext windowContext, InjectorBuilder injectorBuilder) throws Exception {
```
kaspersorensen: I cannot see why the constructor should suddenly start throwing Exception? Nothing seems to have changed.

desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
kaspersorensen: The way we _could_ do it in Guice was to inject a SchemaTreeConfiguration object or so. But I think that's more work and doesn't make anything much clearer.

Another approach would be to break SchemaTree into two classes: A basic one, and one with the Library in it. And then only inject the basic one when needed.

But all in all, I am fine with this "keep it simple, stupid" solution. It's not particularly pretty, but it's also not unlike many other swing components which are generally mutable.

desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
ClaudiaPHI: We have a datastore example with salesforce that doesn't have the password set. Normally when we try to use this datastore in the main window it will throw an exception about it's missing password. My eclipse was too slow and I did not throw the exception in the ui. I ran the jar and it seems to work fine without the constructor throwing the exception. Therefore I removed the unnecessary code

LosD: Tsk, tsk, tsk, tsk... Eclipse again :stuck_out_tongue_winking_eye:

LosD:
Merging! :)

PR #938, opened 2015-11-20
Bug/910 cancel jobs disregarded

### LINKED ISSUE #910: "Cancel job" is disregarded on output data streams
It seems that clicking "Cancel job" has no effect if the parent job has already finished, but output data streams are still running.

For instance: If I used Duplicate detection then it emits all the output data stream rows after it has completed itself. So the child jobs will inherently run _after_ the duplicate detection job. If I click "Cancel job" after the source data stream has finished, nothing happens - or at least it does not stop emitting (and processing) records in/to the output data stream.

ClaudiaPHI:
Fixes #910 

A job could never be cancelled because _resultFuture in AnalysisRunnerSwingWorker class would always be null. The reason was that the job is started in ResultWindow using the method execute from super class SwingWorker, and not doInBackground() from AnalysisRunnerSwingWorker class. I hope it makes sense what I am trying to say. 

Fixed by fetching the result first in cancelIfRunning().

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/util/AnalysisRunnerSwingWorker.java
+++ b/desktop/ui/src/main/java/org/datacleaner/util/AnalysisRunnerSwingWorker.java
@@ -19,6 +19,8 @@
  */
 package org.datacleaner.util;
 
+import java.util.concurrent.ExecutionException;
+
 import javax.swing.SwingWorker;
 
 import org.datacleaner.configuration.DataCleanerConfiguration;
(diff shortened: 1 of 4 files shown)


kaspersorensen:
Hmm I am quite surprised if this is true. Did you verify that this is the cause? Normally we see that the AnalysisRunner.run(..) method returns quite quickly a result future. Also with output data stream jobs. So the object would be available, and I also think it calls ```.cancel()``` on it. But I simply just don't think that it has any effect on the row processing that is going on in the child job.

Of course I don't have much else to say than "I think", but this solution seems unlikely to me at least.

kaspersorensen:
Closing this one as we've found a better approach :)

PR #940, opened 2015-11-20
Bug/899 upload conf.xml

### LINKED ISSUE #899: Upload conf.xml file returns HTTP 406
It seems that the controller that receives an uploaded conf.xml file has stopped working.

This is the form that should be made functional again:
image

It seems that a change in Spring versions may have caused this because nothing really changed on the ```ConfigurationFileController``` for ages.

kaspersorensen:
This does not yet fix #899 - but improves upon the situation ;-)

Update: Now it fixes #899 :)


kaspersorensen:
Aaaaand _now_ it should work.

Fixes #899

PR #946, opened 2015-11-21
Issue #924: Added grouper component

### LINKED ISSUE #924: A transformation that can group records by some key
More of an abstract idea than a concrete function description (for now) ...

Now that we have OutputDataStreams we should also be able to have a transformer that does the equivalent of a GROUP BY. This is quite often going to be useful, especially when using something like the tokenizer (to rows) which produces more than one record for each input record. Using a group by (with some concatenation options) you could re-assemble a string after processing tokens of it.

Note on that thought: Sorting of tokens inside the group might also very well be relevant!

kaspersorensen:
Fixes #924 

Based on my initial extension on repo https://github.com/kaspersorensen/extension_grouper
Works quite well and tested in the field already ;-) Supports various ways of sorting and aggregating values to be grouped.


components/grouper/.gitignore
```diff
@@ -0,0 +1,4 @@
+/.project
+/.settings
+/.classpath
+/target
```
LosD: Can't we stop adding these in every single project folder, and use a proper global .gitignore instead?

```
.project
.settings/
.classpath/
target/
/.idea
*.iml
´´´
Please correct if wrong; I believe all but .project are folders. I also took the liberty to add IDEA project folder and module files (the .idea folder is only created at top level, that's why it's prefixed with /).

File: components/grouper/src/main/java/org/datacleaner/components/group/GrouperTransformer.java [hunk 2]
```diff
@@ -0,0 +1,286 @@
...
+            _name = name;
+        }
+
+        @Override
+        public String getName() {
+            return _name;
+        }
+    }
+
+    @Configured(order = 1, value = PROPERTY_GROUP_KEY)
```
LosD: How will the UI respond to this being first? If I'm not wrong, it would try to link this instead of the aggregated values, which seems wrong.

Or do I remember this wrongly, and the *@*MappedProperty types make it impossible to link using the UI anyway? Then I guess this is the best way to get a bit of the functionality back.

ClaudiaPHI:
LGTM!

components/grouper/src/main/java/org/datacleaner/components/group/GrouperTransformer.java
```diff
@@ -0,0 +1,286 @@
...
+        final List<Object> values = row.getValues(aggregatedValues);
+        if (valueSortation == SortationType.RECORD_ORDER) {
+            // add row ID for the sortation
+            values.add(row.getId());
+        }
+
+        ConcurrentLinkedDeque<List<Object>> newCollectionOfValues = new ConcurrentLinkedDeque<>();
+        ConcurrentLinkedDeque<List<Object>> previousCollectionOfValues = _values
+                .putIfAbsent(key, newCollectionOfValues);
+        if (previousCollectionOfValues == null) {
```
LosD: This seems expensive, memory-wise. Wouldn't it be better to do the aggregation here, instead of storing all values and doing them later?

LosD: Ugh. Sorting destroys that idea.

LosD:
Code looks good to me, but pretty limited in scope, maybe making it a bit useless for more general use.
Worries:
- What about aggregation functions for non-string values? 
- How will it survive on large amounts of data? The CREATE_LIST and sorting features seems to make it very memory intensive.

For a general grouper, I'd prefer the simple case working perfectly (i.e. aggregating instantly), only having the more special cases be expensive.

components/grouper/.gitignore
kaspersorensen: You're right, we should do that...
But .classpath is also a file.

LosD: Thanks, updated comment

components/grouper/src/main/java/org/datacleaner/components/group/GrouperTransformer.java
kaspersorensen: Here's how it looks in UI:

image

components/grouper/src/main/java/org/datacleaner/components/group/GrouperTransformer.java
kaspersorensen: That would be a nice optimization for some cases (depends on combination of sortation and aggregation type). I guess we can cover some of those cases already, but we could also consider it a non-functional improvement/optimization.

kaspersorensen:
I agree that it's not great as is, but it does work. So under the saying "first make it work then make it better", I suggest to take it as it is and just create improvement stories on the rest.

kaspersorensen:
Raised #949 to address your concerns. And now merging ;-)

PR #948, opened 2015-11-22
Preview output data stream

### LINKED ISSUE #931: Add preview option for components in child jobs - when not cautioned against
This story is __Impeded__ by #170

When there are no cautions (see #170) involved in running a "preview" of a job then we should provide a preview function for components placed within child jobs (output data stream consumers). Today we chose not to do that because it is hard to predict if a particular record cut-off is going to satisfy the publisher of the output data stream records, but... That was too conservative and restrictive - when working with output data streams the need for a preview is particularly great because you generally have more complex jobs then.

kaspersorensen:
Implements a preview function for transformers in output data streams.

Despite the claimed prerequisite described in #931 - that we should take caution into account with this - I decided to go ahead for now. Upon further thought I guess the caution-taking part can be also added later if it is deemed important. But for now I see that preview of output data streams is becoming more and more natural, especially if we add also the Grouper component (another open PR) then this should "just work".

Fixes #931 

screen shot 2015-11-22 at 23 00 44

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/actions/PreviewTransformedDataActionListener.java
+++ b/desktop/api/src/main/java/org/datacleaner/actions/PreviewTransformedDataActionListener.java
@@ -22,6 +22,7 @@ package org.datacleaner.actions;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.List;
+import java.util.UUID;
 import java.util.concurrent.Callable;
 
 import javax.swing.table.DefaultTableModel;
@@ -52,6 +53,8 @@ import org.slf4j.LoggerFactory;
  * {@link DataSetWindow}.
  */
 public final class PreviewTransformedDataActionListener implements ActionListener, Callable<TableModel> {
+
+    private static final String METADATA_PROPERTY_MARKER = "org.datacleaner.preview.targetcomponent";
 
     private static final Logger logger = LoggerFactory.getLogger(PreviewTransformedDataActionListener.class);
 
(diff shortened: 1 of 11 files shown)


kaspersorensen:
One note - I'd like to first manually test this on one more case: When the Union component generates the output data stream. I can see that as a corner case because it has potentially multiple source tables. But actually I think it works ... testing ...

kaspersorensen:
Ugh. Didn't work. Something with the added max-rows filter ... will check

desktop/api/src/main/java/org/datacleaner/actions/PreviewTransformedDataActionListener.java
```diff
@@ -196,8 +289,8 @@ public TableModel call() throws Exception {
     }
 
     private AnalysisJobBuilder copy(final AnalysisJobBuilder original) {
-        final AnalysisJob analysisJob = original.withoutListeners().toAnalysisJob(false);
-        final AnalysisJobBuilder ajb = new AnalysisJobBuilder(original.getConfiguration(), analysisJob);
-        return ajb;
+        final AnalysisJob analysisJob = original.getRootJobBuilder().withoutListeners().toAnalysisJob(false);
+        final AnalysisJobBuilder copy = new AnalysisJobBuilder(original.getConfiguration(), analysisJob);
```
LosD: Not new code, but pretty funky going from builder to job to builder to make a copy :)

LosD:
Code looks good to me. Wonderful to see some of the ```if```s in the preview "stream" go away.

kaspersorensen:
Do you wanna look again @losd or is it fine?

desktop/api/src/main/java/org/datacleaner/actions/PreviewTransformedDataActionListener.java
```diff
@@ -136,12 +168,52 @@ public TableModel call() throws Exception {
...
+            for (FilterComponentBuilder<?, ?> filter : filters) {
+                if (filter.getComponentInstance() instanceof MaxRowsFilter) {
+                    rootJobBuilder.removeFilter(filter);
+                }
+            }
+
+            final SourceColumnFinder sourceColumnFinder = new SourceColumnFinder();
+            sourceColumnFinder.addSources(rootJobBuilder);
+            final List<Table> sourceTables = rootJobBuilder.getSourceTables();
+            final int maxRows = Double.valueOf(Math.ceil(((double) _previewRows) / sourceTables.size())).intValue();
```
LosD: It's not really a problem currently, but I don't believe that there is anything inherently in the multi stream component base class that makes 1 row per source table row, so we may end up with less rows that actually wanted for a new component, unless the developer is aware of this limitation.

Maybe we should at least point it out in the JavaDoc for the ```MultiStreamComponent``` class.

LosD:
A single worry, and a single indentation issue, otherwise LGTM

PR #951, opened 2015-11-23
Added ignore rule for common JUNG error

kaspersorensen:
I've run into this issue a number of times and don't see any way we can "fix" it since it must be burried deep in the JUNG framework.

So instead here's a suggestion to make our "uncaught exception handler" simply ignore it.


LosD:
It's nasty, but it's better than random-ish errors in the UI.

PR #952, opened 2015-11-23
Fixes various inconsistencies around model and graph not auto-updating.

### LINKED ISSUE #953: "Build key/value map" looses all it's input columns when linked to "Compose JSON document"
I have no clue what is going on here, but to describe a (reproducable) scenario:

1) Add source table to canvas
2) Add "Compose & write JSON document" to canvas
3) Add "Build key/value map" to canvas
4) Open configuration dialog for "Build key/value map".
5) Click "Select all" on the input columns.
6) Close the configuration dialog for "Build key/value map".
7) Right-click the "Build key/value map" component, select "Link to" and link it to the "Compose & write JSON document" component.

At this point both components seem unconfigured.

Expected: Both components would be configured.

kaspersorensen:
While monkey-testing the output data streams and the preview function on
transformations following output data streams, a couple of issues popped
up in nasty ways. This commit fixes these.

This is a bit of a bucket of different things. I will add my own remarks in the diff so that it can become clear why the changes was made.

And now also ...

Fixes #953


components/grouper/src/main/java/org/datacleaner/components/group/GrouperTransformer.java
```diff
@@ -132,7 +132,8 @@ public void init() {
         outputDataStreamBuilder.withColumn("row_count", ColumnType.INTEGER);
         for (int i = 0; i < aggregatedValues.length; i++) {
             final InputColumn<?> inputColumn = aggregatedValues[i];
-            final AggregationType aggregationType = aggregationTypes[i];
+            final AggregationType aggregationType = (aggregationTypes.length <= i ? AggregationType.CREATE_LIST
+                    : aggregationTypes[i]);
```
kaspersorensen: During updates of the graph this method will be called multiple times - also before the property widget has been able to provide an 'aggregationTypes' value which matches the size of the 'aggregatedValues' value.

desktop/api/src/main/java/org/datacleaner/windows/DataSetWindow.java
```diff
@@ -158,12 +170,12 @@ protected void done() {
...
+                } catch (Throwable e) {
+                    e = ErrorUtils.unwrapForPresentation(e);
+                    logger.error("Unexpected error occurred while building DataSetWindow contents", e);
                     DataSetWindow.this.dispose();
-                    if (e instanceof RuntimeException) {
-                        throw (RuntimeException) e;
-                    }
-                    throw new IllegalStateException(e);
+                    WidgetUtils.showErrorMessage("Unexpected error",
+                            "An unexpected error occurred while building data set. See logs for details.");
```
kaspersorensen: This window is used to present data, but querying it or building the preview job is a pretty complex task that may quite likely go wrong. For this reason I felt it was better to hide the underlying exception from the user's view - The fact that it didn't work it most likely enough.

desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleInputColumnsPropertyWidget.java
```diff
@@ -362,10 +364,20 @@ public void onRemove(TransformerComponentBuilder<?> transformerJobBuilder) {
     }
 
     @Override
+    protected void onPanelAdd() {
+        super.onPanelAdd();
+        final AnalysisJobBuilder analysisJobBuilder = getAnalysisJobBuilder();
+        analysisJobBuilder.addSourceColumnChangeListener(this);
+        analysisJobBuilder.addTransformerChangeListener(this);
```
kaspersorensen: This was moved from constructor to here - to make sure the add and remove timing is completely orthogonal.

engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
```diff
@@ -310,11 +310,23 @@ public void importJob(AnalysisJob job) {
     }
 
     public AnalysisJobBuilder removeSourceColumn(MetaModelInputColumn inputColumn) {
-        final boolean removed = _sourceColumns.remove(inputColumn);
-        if (removed) {
+        final int index = _sourceColumns.indexOf(inputColumn);
+        if (index != -1) {
+            final MetaModelInputColumn removedColumn = _sourceColumns.remove(index);
```
kaspersorensen: Using index we get the exact same input column, not just one that is equal. This turned out to be important or else some property values would not update when calling ComponentBuilder.removeInputColumn(...)

engine/core/src/main/java/org/datacleaner/job/builder/AnalyzerComponentBuilder.java
```diff
@@ -235,6 +236,21 @@ public AnalyzerJob toAnalyzerJob(boolean validate) throws IllegalStateException
...
 
     @Override
+    public AnalyzerComponentBuilder<A> removeInputColumn(InputColumn<?> inputColumn,
+            ConfiguredPropertyDescriptor propertyDescriptor) {
+        assert propertyDescriptor.isInputColumn();
+        if (inputColumn == null) {
+            throw new IllegalArgumentException("InputColumn cannot be null");
+        }
+        if (isMultipleJobsDeterminedBy(propertyDescriptor)) {
+            _inputColumns.remove(inputColumn);
```
kaspersorensen: This uses the same pattern as the addXxxx method. For some reason it was missing here and it could easily cause runtime exceptions when removing source columns.

LosD:
Code looks good, but there's a lot of weird re-wrapping of code to 80 characters instead of 120. It seems as if the computer used has not been using our standard formatting.

LosD:
Hmmm... I'm not sure it is 80 characters, some of the rewrapping is actually making lines longer, by wrapping early.

desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleInputColumnsPropertyWidget.java
```diff
@@ -178,8 +177,8 @@ protected void onChange(DocumentEvent event) {
                 final JButton expressionColumnButton = WidgetFactory
                         .createSmallButton(IconUtils.MODEL_COLUMN_EXPRESSION);
                 expressionColumnButton.setToolTipText("Create expression/value based column");
-                expressionColumnButton.addActionListener(AddExpressionBasedColumnActionListener
-                        .forMultipleColumns(this));
+                expressionColumnButton
+                        .addActionListener(AddExpressionBasedColumnActionListener.forMultipleColumns(this));
```
LosD: Example of weird rewrap

desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleInputColumnsPropertyWidget.java
```diff
@@ -228,8 +228,8 @@ private void updateComponents() {
 
     private void updateComponents(final InputColumn<?>[] value) {
         // fetch available input columns
-        final List<InputColumn<?>> availableColumns = getAnalysisJobBuilder().getAvailableInputColumns(
-                getComponentBuilder(), _dataType);
+        final List<InputColumn<?>> availableColumns = getAnalysisJobBuilder()
+                .getAvailableInputColumns(getComponentBuilder(), _dataType);
```
LosD: This isn't really weird, but it seems pretty aggressive. The old line was 96, and I believe we use 120 as the max width.

PR #955, opened 2015-11-24
Bug/910 error aware cancellation

### LINKED ISSUE #910: "Cancel job" is disregarded on output data streams

kaspersorensen:
Fixes #910 

The approach here has a few important aspects to understand what is going on:

 * The way we do cancellation is by updating the ErrorAware implementation (which is the analysis listener).
 * Now the OutputDataStreamRowCollector has access to the ErrorAware and will throw exceptions if the job is errornuous or cancelled. This exception will be used to break any long-running for loop in components that are publishing output data stream records.
 * In AnalysisRunnerJobDelegate we've had to go away from "assisting execution" to sleeping. This was to avoid that the return of AnalysisResultFuture from AnalysisRunner (in the swing worker) was blocked by a long-running task that was assisted by the AnalysisRunnerJobDelegate.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/util/AnalysisRunnerSwingWorker.java
+++ b/desktop/ui/src/main/java/org/datacleaner/util/AnalysisRunnerSwingWorker.java
@@ -19,6 +19,8 @@
  */
 package org.datacleaner.util;
 
+import java.util.concurrent.TimeUnit;
+
 import javax.swing.SwingWorker;
 
 import org.datacleaner.configuration.DataCleanerConfiguration;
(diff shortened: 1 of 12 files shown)


ClaudiaPHI:
LGTM!

PR #956, opened 2015-11-25
Remove order by from datahubconnector

drexler42:
DataHubConnector no longer enforces an ORDER by clause on all queries sent to the DataHub virtual data stores, because that responsibility is now handled inside the Virtual Datastores. This simplifies the code of the DataHubConnector.


kaspersorensen:
LGTM!

PR #961, opened 2015-11-25
Feature/959 hdfs filename instead folder

### LINKED ISSUE #959: Ability to specify result HDFS filename instead of directory name

ClaudiaPHI:
Fixes #959 

- Gave the user the possibility to name it's job 
- Handled various path naming scenarios 
- Added Tests


engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -59,6 +62,8 @@
     private static String DATA_CLEANER_RESULT_PATH_PROPERTY = "datacleaner.result.hdfs.path";
     public static final String ACCUMULATOR_CONFIGURATION_READS = "DataCleanerConfiguration reads";
     public static final String ACCUMULATOR_JOB_READS = "AnalysisJob reads";
+    private static String DEFAULT_RESULT_PATH = "/datacleaner/results/";
+    private static String RESULT_FILE_EXTENSION = ".analysis.result.dat";
```
kaspersorensen: Make sure to also add "final" to static properties. I see this also needed a few lines above (although not part of this PR).

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -220,11 +232,49 @@ public String getResultPath() {
...
+        final String jobName = _analysisJobXmlPath.substring(lastIndexOfSlash + 1, lastIndexOfFileExtension);
+        return jobName;
+    }
+
+    /**
+     * Gets the hdfs path of job's result
+     * 
+     * @return
+     */
+    public String getResultJobFilePath() {
```
kaspersorensen: the concept "result job file" does not make sense to me. Either it's a "job file" or a "result file".

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -212,7 +219,12 @@ private ComponentJob getComponentByKey(final AnalysisJob job, final String queri
         return null;
     }
 
-    public String getResultPath() {
+    /**
+     * Gets the path defined in the job properties file
+     * 
+     * @return
+     */
+    public String getResultPathUsercustomized() {
```
kaspersorensen: I think I liked the previous method name better here ... what is the user customized part? Is it important for the method?

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -212,7 +219,12 @@ private ComponentJob getComponentByKey(final AnalysisJob job, final String queri
...
         if (_customProperties != null) {
             if (_customProperties.containsKey(DATA_CLEANER_RESULT_PATH_PROPERTY)) {
```
kaspersorensen: This if-block doesn't archieve anything since the next line would also return null when the key is not represented in the map.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
kaspersorensen: And looking at what it does, compared to the method ```getResultJobFilePath``` below, then I don't think this method should be public since it serves only an internal purpose.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -220,11 +232,49 @@ public String getResultPath() {
...
+        String resultPath = getResultPathUsercustomized();
+        final Configuration hadoopConfiguration = _sparkContext.hadoopConfiguration();
+        final String fileSystemPrefix = hadoopConfiguration.get("fs.defaultFS");
+        if (resultPath == null || resultPath.isEmpty()) {
+            resultPath = fileSystemPrefix + DEFAULT_RESULT_PATH;
+        } else {
+            final URI uri = URI.create(resultPath);
+            if (!uri.isAbsolute()) {
+                if (!resultPath.startsWith("/")) {
+                    resultPath = "/" + resultPath;
```
kaspersorensen: doesn't the path normally start with "hdfs://.."? In that case we are now constructing a path like "/hdfs://.." I think

engine/env/spark/src/test/java/org/datacleaner/spark/JobResultPathTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.junit.Test;
+
+/**
+ * This class tests the path of the result specified in the job properties file
+ * or otherwise. The default setting for 'fs.defaultFS' is 'file:///'
+ */
+public class JobResultPathTest extends TestCase {
+
+    @Test
+    public void testResultJobNameCorrectSpecified() throws Exception {
```
kaspersorensen: I have the same confusion regarding naming here. It interchangably uses term "result job" and "job result".

engine/env/spark/src/test/java/org/datacleaner/spark/JobResultPathTest.java
```diff
@@ -0,0 +1,152 @@
...
+import junit.framework.TestCase;
+
+import org.apache.spark.SparkConf;
+import org.apache.spark.api.java.JavaSparkContext;
+import org.datacleaner.job.AnalysisJob;
+import org.junit.Test;
+
+/**
+ * This class tests the path of the result specified in the job properties file
+ * or otherwise. The default setting for 'fs.defaultFS' is 'file:///'
```
kaspersorensen: How is the default setting "file:///" set? For me this is news, but also a bit of a surprise and strange thing in the test to rely on?

engine/env/spark/src/test/java/org/datacleaner/spark/JobResultPathTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Test
+    public void testResultJobNameJustFolderSpecified() throws Exception {
+        final SparkConf sparkConf = new SparkConf().setMaster("local").setAppName("DCTest - " + getName());
+        final JavaSparkContext sparkContext = new JavaSparkContext(sparkConf);
+        try {
+
+            final SparkJobContext sparkJobContext = new SparkJobContext(sparkContext,
+                    "src/test/resources/conf_local.xml", "src/test/resources/vanilla-job.analysis.xml",
+                    "src/test/resources/jobProperties/job2.properties");
```
kaspersorensen: could we name all these properties file with some semantics instead of numbers? Right now it is pretty tough to pair a properties file with a test case in the reader's mind.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -68,6 +73,7 @@
     private final String _analysisJobXml;
     private final String _analysisJobXmlPath;
     private final Map<String, String> _customProperties;
+    private final JavaSparkContext _sparkContext;
```
kaspersorensen: You cannot keep the spark context in this class, since it needs to be serializable.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
ClaudiaPHI: It's just a part of a solution. 
If the hostname and port specified in core-site xml is 'hdfs://bigdatavm:9000' and the user introduced the path: target/results/myresult.analysis.dat. Then the path will be: 
hdfs://bigdatavm:9000target/results/myresult.analysis.dat which is not valid. 

What I did not handled is the case where the hostname and port allready has a "/". 
The path will be 'hdfs://bigdatavm:9000//target/results/myresult.analysis.dat' which is not valid either.

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -40,9 +41,9 @@
 
 public class Main {
 
-    static Logger logger = Logger.getLogger(Main.class);
-
-    private static String DEFAULT_RESULT_PATH = "/datacleaner/results/";
+    private static final Logger logger = Logger.getLogger(Main.class);
+    private static final String DEFAULT_RESULT_PATH = "/datacleaner/results";
+    private static final String RESULT_FILE_EXTENSION = ".analysis.result.dat";
```
kaspersorensen: Would be slightly better to refer to ```FileFilters.ANALYSIS_RESULT_SER.getExtension()```.

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -96,25 +97,50 @@ public static void main(String[] args) {
...
     }
 
-    private static String getResultJobFilePath(final JavaSparkContext sparkContext,
-            final SparkJobContext sparkJobContext)  {
+    /**
+     * Gets the hdfs path of job's result
+     * 
+     * @return
+     */
+    protected static String getResultFilePath(final JavaSparkContext sparkContext, final SparkJobContext sparkJobContext) {
```
kaspersorensen: Can we move these methods to a Util class instead of Main? Preferably the Main class is only for spark-submit and reading the command line arguments into some sensible object.

engine/env/spark/src/main/java/org/datacleaner/spark/utils/ResultFilePathUtils.java
```diff
@@ -0,0 +1,86 @@
...
+        try {
+            URI systemPrefix = new URI(fileSystemPrefix);
+            if (!resultPath.startsWith("/")) {
+                resultPath = "/" + resultPath;
+            }
+            uri = new URI("hdfs", null, systemPrefix.getHost(), systemPrefix.getPort(), systemPrefix.getPath()
+                    + resultPath, null, null);
+            return uri.toString();
+        } catch (URISyntaxException e) {
+            logger.error("Error while trying to create url for saving the job", e);
```
kaspersorensen: Is there any sense in just logging the error here, and continuing with a null value, effectively swallowing the exception? Normally failing fast is better because then the user knows he needs to change something.

engine/env/spark/src/test/java/org/datacleaner/spark/SparkAnalysisRunnerTest.java
```diff
@@ -67,10 +67,9 @@ public void testVanillaScenario() throws Exception {
                     "src/test/resources/conf_local.xml", "src/test/resources/vanilla-job.analysis.xml");
             final AnalysisJob job = sparkJobContext.getAnalysisJob();
             assertNotNull(job);
-            assertEquals("vanilla-job",sparkJobContext.getAnalysisJobName());
```
kaspersorensen: why is this removed?

engine/env/spark/src/main/java/org/datacleaner/spark/utils/ResultFilePathUtils.java
```diff
@@ -0,0 +1,86 @@
...
+    
+    /**
+     * Gets the hdfs path of job's result
+     * 
+     * @return
+     */
+    public static String getResultFilePath(final JavaSparkContext sparkContext, final SparkJobContext sparkJobContext) {
+        String resultPath = sparkJobContext.getResultPath();
+        final Configuration hadoopConfiguration = sparkContext.hadoopConfiguration();
+        final String fileSystemPrefix = hadoopConfiguration.get("fs.defaultFS");
```
kaspersorensen: Can you add a comment here about what this value will normally be? In general please document this method with comments because it's pretty involved.

engine/env/spark/src/test/java/org/datacleaner/spark/SparkAnalysisRunnerTest.java
ClaudiaPHI: because the test was to crowded. I added, I deleted it. :-P

engine/env/spark/src/main/java/org/datacleaner/spark/utils/ResultFilePathUtils.java
ClaudiaPHI: It does return null. code line 82.

kaspersorensen: yes but my point is whether that is good behaviour. Returning null will lead the application down another path than failing. Rather I guess it will make the job run but save the result in some generated-path. If the user asks to have the result saved in X, then we should save it in X or fail, not save it in Y.

ClaudiaPHI: This method will be called if the user has set a relative path such as "myresults/" and there is a an error and it will return null then hdfs("Main" line 76 ) will try to write to an invalid path, the application will fail.

kaspersorensen:
LGTM

PR #968, opened 2015-11-27
remove dictionary matches improvement

### LINKED ISSUE #944: Add "removed part" output column to "Remove dictionary matches"
In the component "Remove dictionary matches" we remove a part (or even multiple parts) of a string based on dictionary matches.

For the user it is sometimes very useful to know _what_ was removed. So we should add an output column for that.

Technical note: Since it may sometimes remove multiple parts of the string, should this output type really be a List rather than a String? I think so, although I also don't feel that it's very intuitive for beginners. Maybe we should have an optional property for selecting how to handle that.

ClaudiaPHI:
The remove dictionary transformer returns the "removed" matches as String or as a List. The default is String. 
Fixes #944


ClaudiaPHI:
image

image


image

image

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -59,9 +65,14 @@
     @Configured(value = PROPERTY_COLUMN)
     InputColumn<String> _column;
 
+    @Inject
+    @Configured(required = false)
+    @Description("Get removed strings as List")
+    boolean _removedMatchesAsList = false;
```
kaspersorensen: Could we use an enum for this instead? Would be slightly more intuitive (and future-proof) to select between multiple strategies. And I would like the default to be list.

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -98,27 +114,61 @@ public void close() {
...
-            return value;
+    public Object[] transform(final String value) {
+        if (!_removedMatchesAsList) {
+            return getRemovedMatchesAsString(value);
+        } else {
+            return getRemovedMatchesAsList(value);
         }
+    }
+
+    public String[] getRemovedMatchesAsString(final String value) {
```
kaspersorensen: I don't think the name of this method is correct, since it does the complete transformation not just the latter part.

ClaudiaPHI:
image

ClaudiaPHI:
Merge ?

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -50,8 +56,25 @@
 @Categorized(superCategory = ImproveSuperCategory.class, value = ReferenceDataCategory.class)
 public class RemoveDictionaryMatchesTransformer implements Transformer {
 
+    public static enum RemovedMatches implements HasName {
```
kaspersorensen: Consider if we can find a better name for the enum. Consider if it can describe that it's a choice/option/setting/type.

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -79,14 +108,21 @@ public RemoveDictionaryMatchesTransformer(InputColumn<String> column, Dictionary
     @Override
     public OutputColumns getOutputColumns() {
         final String name = _column.getName() + " (" + _dictionary.getName() + " removed)";
-        return new OutputColumns(String.class, new String[] { name });
+        switch (_removedMatches) {
+        case LIST:
+            return new OutputColumns(new String[] { name, OUTPUT_COLUMN_REMOVED_MATCHES }, new Class[] { String.class,
+                    List.class });
+        default:
```
kaspersorensen: While the meaning is the same (right now), using default instead of a concrete value isn't very easy on the reader of the code. Consider (in general, not just here) that you have two settings, but tomorrow you might have three. So arrange the code as if they are adjacent options instead of an "if else" kind of structure.

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -98,27 +134,62 @@ public void close() {
...
+                    removedString.append(' ');
+                }
+                removedString.append(token);
             }
         }
+        return new String[] { survivorString.toString(), removedString.toString() };
+    }
 
-        return sb.toString();
+    private Object[] transform2(final String value) {
```
kaspersorensen: Very undescriptive method name

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -98,27 +134,62 @@ public void close() {
...
 
-    public String transform(final String value) {
-        if (Strings.isNullOrEmpty(value)) {
-            return value;
+    public Object[] transform(final String value) {
+        switch (_removedMatches) {
+        case LIST:
+            return transform2(value);
+        default:
+            return transformDefault(value);
```
kaspersorensen: The logic for the first value (value without dictionary matches) is now duplicated code. Wouldn't it be better to make it a shared codebase and just handle a match-list differently. That way we don't branch out at the top level of the transform-method but just have two (maybe more) different ways of dealing with the "Removed matches".

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
ClaudiaPHI: could not find a better name.

ClaudiaPHI:
Fixed per suggestions

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -50,8 +56,25 @@
...
+
+        @Override
+        public String getName() {
+            if (this == STRING) {
+                return "String";
+            } else {
+                return "List";
+            }
+
+        }
```
LosD: Wouldn't it be better to give the names as part of the enum constructor?

```Java
STRING("String), LIST("List);

private final String _name;
private RemovedMatchesType(String name){ _name = name; }

@Override
public String getName(){ return _name;}
```

The less logic, the better. We get the added bonus of not having to change the if to if-else or a switch if other types are added.

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -98,27 +135,64 @@ public void close() {
...
+        if (removedObjects instanceof StringBuilder) {
+            StringBuilder removedString = (StringBuilder) removedObjects;
+            if (removedString.length() != 0) {
+                removedString.append(' ');
+            }
+            removedString.append(token);
+        } else if (removedObjects instanceof ArrayList) {
+            ((ArrayList<String>) removedObjects).add(token);
+        }
+    }
```
LosD: This is... Honestly, pretty nasty. I'd consider using overloading, if using the Object class for removedObjectMatches is absolutely needed (which I would try very, very hard to avoid).

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -98,27 +135,64 @@ public void close() {
...
+                addRemovedMatches(removedStringMatches, token);
             }
         }
 
-        return sb.toString();
+        switch (_removedMatchesType) {
+        case STRING:
+            return new String[] { survivorString.toString(), removedStringMatches.toString() };
+        case LIST:
+            return new Object[] { survivorString.toString(), removedStringMatches };
```
LosD: I may be wrong here, but I don't think this really has any effect. Just always returning Object[] should be okay.

kaspersorensen:
I suggest to squash this before merging ... But other than that, please review.

LosD:
LGTM!

PR #969, opened 2015-11-27
Feature/966 component variant flag

jhorcicka:
In XML representation of a component (transformer) there is a new metadata property called "source" with potential values "local" or the remote server name (which is specified in conf.xml).


kaspersorensen:
Right now this metadata property is set via the graph UI, but I would think it much more natural to set it in the core of the job building (AnalysisJobBuilder or maybe AbstractComponentBuilder). Otherwise it become UI-driven which it shouldn't.

jhorcicka:
We think coordinates metadata can be set in desktop (JobGraph) as it was before. ComponentDescriptor was extended with updateMetadata(...) method which is called in AbstractComponentBuilder. Particular implementations of ComponentDescriptor fill their metadata. So, RemoteTransformerDescriptorImpl fills "source" now.

In desktop there already is the JobGraphMetadata class which fills the coordinates. If this needs to be refactored it might be a bigger issue to be solved on its own, I guess.

api/src/main/java/org/datacleaner/descriptors/ComponentDescriptor.java
```diff
@@ -44,6 +45,13 @@
 public interface ComponentDescriptor<B> extends Comparable<ComponentDescriptor<?>>, Serializable {
 
     /**
+     * It updates component's metadata.
+     * 
+     * @param metadata
+     */
+    public void updateMetadata(Map<String, String> metadata);
```
kaspersorensen: I think the location of this method is a bit difficult... First of all it leaves me wondering "when does this happen". But after answering that (it happens when the component is created) then I think that the order of control is a bit reversed here. A descriptor is usually unaware of the component instances, and I think it's best to keep it that way so that we don't introduce coupling everywhere. The other problem I see is that we don't want the descriptor to just update _anything_ in the map. We want to just provide one specific property, so passing the map around might be dangerous.

My suggestion would be to write and read this metadata as part of the job building API. I would suggest to simply do a check when the component is created: Find the appropriate descriptor provider, and if it's remote then set the metadata property which deals with the origin of the descriptor. That way we also don't pass around a map that could contain _anything_  - we only set one specific metadata property which could/should also be defined as a constant somewhere so that we can do an easy and safe lookup.

jhorcicka:
There is a clash between "RemoteDescriptorProvider should provide server name and priority" and "remove RemoteTransformerDescriptor.setRemoteDescriptorProvider". 

At some places it is required to get remote server data from a transformer. It can reach it via RemoteDescriptorProvider property (setter and getter defined in the interface). If the access to RemoteDescriptorProvider was removed from RemoteTransformerDescriptor, the server data (name, priority, ...) could be provided within a creation of a transformer. In this case getters&setters for name, priority, ... are needed.

jhorcicka:
I added additional getters (host, credentials) to RemoteDescriptorProvider so that I could remove duplicated fields (that are reachable via descriptor provider's server data) in RemoteTransformerDescriptorImpl. 

I also removed setter for RemoteDescriptorProvider in RemoteTransformerDescriptor (main reason for the last update). 

After these changes I could also simplified RemoteTransformerDescriptorImpl's constructor.

kaspersorensen:
I only meant to remove the setters from the interface - NOT the getters. That way it can still be retrieved, but mutating the object is not as easy (preferably impossible, but I can live with it if it's not).

engine/core/src/main/java/org/datacleaner/descriptors/RemoteTransformerDescriptor.java
```diff
@@ -0,0 +1,29 @@
...
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.descriptors;
+
+/**
+ * A {@link TransformerDescriptor} for remote transformers.
+ */
+public interface RemoteTransformerDescriptor extends TransformerDescriptor {
+    public void addPropertyDescriptor(ConfiguredPropertyDescriptor propertyDescriptor);
```
kaspersorensen: Again I don't think mutating behaviour should be part of the interface. At least consider: Would anyone (except the provider which is effectively the factory of this object) have any use of setting/adding a property?

jhorcicka:
What removed getters do you mean? From RemoteTransformerDescriptor (interface) I removed only a setter (setRemoteDescriptorProvider). 

As for getters, I removed unnecessary getters from RemoteTransformerDescriptorImpl (class).

kaspersorensen:
I was referring to your statement "In this case getters&setters for name, priority, ... are needed." :)

jhorcicka:
Ahh, OK. I guess that's fine now because the reference RemoteTransformerDescriptor -> RemoteDescriptorProvider is still available.

jhorcicka:
Please, review again.

engine/xml-config/src/test/resources/JaxbJobWriterTest-file4.xml
```diff
@@ -16,6 +16,9 @@
     <transformation>
         <transformer>
             <descriptor ref="Email standardizer"/>
+            <metadata-properties>
+                <property name="source">local</property>
```
kaspersorensen: Oh I just realize that this will impact _ALL_ jobs ever made .... Could we have "local" as a default and not write it if we don't need to? Sorry for not seeing this before now.

jhorcicka: OK. Property "source" is not filled when the value is "local", so only remote components have this metadata property and its value is the server name.

kaspersorensen:
Looks good to me then! Merge at will ;-)

Issue #967, opened 2015-11-27
Hadoop server dialog looses the port number on "blur"


kaspersorensen:
If I tab through to the port number in this dialog:

![image](https://cloud.githubusercontent.com/assets/291450/11440805/f4c22060-9505-11e5-8982-9343f53acec5.png)

and then tab once again, then the port number is removed and I never seem to be able to set it again.

PR #973, opened 2015-11-30
Issue #967: Fixed HDFS port text field

kaspersorensen:
The issue was caused by the NumberFormat applying grouping (thousand
separator) which then made the FormattedTextField try to submit the
value as e.g. "8,020" instead of "8020". This then caused the
NumberDocument to reject the value and return empty string instead.

Also did a small change to update the default port from 9000 to 8020.


LosD:
LGTM

PR #980, opened 2015-12-02
Bug/963 remote servers problem notifications

rposkocil:


desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -293,6 +243,11 @@ public void onRemove(final InputColumn<?> sourceColumn) {
     private static final int DEFAULT_WINDOW_WIDTH = 1000;
     private static final int DEFAULT_WINDOW_HEIGHT = 710;
 
+    private static final int SERVER_CHECK_INTERVAL = 2000; // [ms]
+    private DescriptorProviderState _descriptorErrorState;
+    private Thread serverCheckingThread;
```
kaspersorensen: It looks like the thread is completely controlled by the ServerChecker class. So I would suggest not to save the thread as a field in this class - this class is already very big and should not increase in complexity. Avoiding the extra field helps in that.

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -293,6 +243,11 @@ public void onRemove(final InputColumn<?> sourceColumn) {
     private static final int DEFAULT_WINDOW_WIDTH = 1000;
     private static final int DEFAULT_WINDOW_HEIGHT = 710;
 
+    private static final int SERVER_CHECK_INTERVAL = 2000; // [ms]
```
kaspersorensen: This is a very aggressive check. I would suggest something in an order of magnitude bigger. Every 2 minutes would be more sensible.

kaspersorensen:
I don't like that this facility is tied to the UI of datacleaner instead of being a core/backend thing. I would initially think it should be part of the remote descriptor provider itself.

The timeout period is pretty insane here too. I would expect a very low-frequency check (in minutes, certainly not seconds) and maybe even a reactive check so that it only checks whenever I try to get new descriptors, run a job or so.

Next to that, it is actually scoped to the _window_ which means that if I have multiple windows open, then it will be doing multiplying the checks for each window.

khouzvicka:
I mean the thread must be in the window class. The checker thread uses backend method DescriptorPrivider.getStatus to get information about servers. And this information is used to notify all windows. Each window has a special workflow with different transformers. This thread analyzes this workflow and notifies the window.
Notification is different for workflows with remote transformers and for workflows without them, when the remote server is offline.
Special thread is useful because in code there might be a long call to remote server.

kaspersorensen:
Could we instead have the thread in the descriptor provider and then have a listener/observer pattern implemented, so that it's not bound to just the UI?

khouzvicka:
I've updated the PR, can you please review again?
Your suggestions are implemented.
New functionalities are tested and bugs are fixed.

api/src/main/java/org/datacleaner/configuration/DataCleanerEnvironment.java
```diff
@@ -62,4 +62,9 @@
      */
     public InjectionManagerFactory getInjectionManagerFactory();
 
+    /**
+     * @see DescriptorProviderStateNotifier
+     * @return instance of notifier
+     */
+    public DescriptorProviderStateNotifier getDescriptorProviderStateNotifier();
```
kaspersorensen: Feels like this would be better off just being part of the DescriptorProvider. Adding the method at a higher level (the configuration - which is a configuration of many things) makes the specific thing impact the more general thing, which isn't a good separation of concerns.

kaspersorensen:
This looks better, but I don't think the relationship between the different objects/interface here is quite clear and correct. Specifically:

 * DataCleanerConfiguration - I would not want a change to this very central interface just because we want to add a listener/monitoring mechanism on a totally different interface.
 * DescriptorProvider and the new notifier class - I feel that there is really some abstraction leak going on here right now. Why would anyone want to register listeners on a separate notifier class? Normally when you register a listener, you register it on the object that you want to observe. So I would rather just suggest to add to methods a la ```addListener(...)``` and ```removeListener(...)``` on the DescriptorProvider interface and let everything else be considered implementation details and NOT expose those details as part of the interface. The same principle is with regards to the new method for getting the descriptor provider state, which is now a ```Map<DescriptorProvider, Status>```. This seems overly complex because it leaks the internal fact that the descriptor provider _could_ be a composite object. But it should simply delegate the ```.addListener``` call to its children and otherwise behave like a normal citizen.

kaspersorensen:
Hi Karel ( @khouzvicka ) please get in touch regarding this PR. I think it might conflict with some other work I've been working on, so I think it's better if we have a sync on skype or so.

kaspersorensen:
Karel I think you're working on a new approach to solve this issue? So maybe we should close this PR? Btw. we just merged "5.0" into master branch, so you can branch off of that from now on.

khouzvicka:
So, you can close this PR and I will do a new PR - branch 963 to master

PR #981, opened 2015-12-02
Add transformer for subtracting text.

### LINKED ISSUE #945: A transformer for removing/subtracting string A from string B
In parsing and standardization scenarios there is often a situation where one or more strings should be removed from another string.

To give a full scenario: Imagine we have a unstructured text "Metal plate 20x30x2 with 2 inch punch holes". A parser would maybe identify "20x30x2" as the dimensions of the plate, and have that available in a separate transformed output column. Now the task would be to remove the dimensions from the unstructured text so that it is reduced to just "Metal plate with 2 inch punch holes".

On the configuration side:

 * Should have a String input for the text to subtract from.
 * Should have a String, Number and/or List input for the stuff to subtract.
  * If String or Number is provided, simply subtract once.
  * If a list is provided, subtract every element.

LosD:
This will add a new transformer, capable of subtracting substrings from a base text.

Fixes #945


ClaudiaPHI:
LGTM!

LosD:
Thanks. I'd like a short review from @kaspersorensen, especially regarding the naming of the transformer, as well as if the functionality is as he wanted.

I extended the functionality it in two way: You can choose as many columns as you want for subtracting, and there is the possibility to only remove matches for on full words (i.e., If your base text is "Goodbye", and the matching word is "bye", it will be removed with Only whole words disabled, and not removed if it is disabled).

I'm thinking that it may be a good idea to add a case sensitivity setting as well?

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextSubtractTransformer.java
```diff
@@ -0,0 +1,83 @@
...
+    @Description("Column containing the text to subtract from")
+    InputColumn<String> baseColumn;
+
+    @Configured("Substring columns")
+    @Description("Columns containing the substrings to remove from the base text")
+    InputColumn<?>[] substringColumns;
+
+    @Configured(value="Whole words only", required = false)
+    @Description("If set, only whole words (surrounded by whitespace or punctuation) will be removed")
+    boolean wholeWordsOnly = false;
```
kaspersorensen: For me we could maybe explain this a bit easier by telling what you are preventing. If it was called "prevent partial word replacement" or something like that, then I think I would better understand it.

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextSubtractTransformer.java
```diff
@@ -0,0 +1,83 @@
...
+import org.datacleaner.api.Configured;
+import org.datacleaner.api.Description;
+import org.datacleaner.api.InputColumn;
+import org.datacleaner.api.InputRow;
+import org.datacleaner.api.OutputColumns;
+import org.datacleaner.api.Transformer;
+import org.datacleaner.components.categories.TextCategory;
+import org.datacleaner.components.convert.ConvertToStringTransformer;
+
+@Named("Text subtractor")
```
kaspersorensen: Suggestion: "Remove substring"

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextSubtractTransformer.java
```diff
@@ -0,0 +1,83 @@
...
+@Description("Subtracts one or more values from a base text")
```
kaspersorensen: Can we give this a bit more body to explain it more?

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextSubtractTransformer.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public OutputColumns getOutputColumns() {
+        return new OutputColumns(String.class, baseColumn.getName() + " (subtracted)");
+    }
+
+    @Override
+    public String[] transform(final InputRow inputRow) {
+        String subtractedString = inputRow.getValue(baseColumn);
+        for (final InputColumn<?> inputColumn : substringColumns) {
+            if (inputColumn.getDataType() == List.class) {
```
kaspersorensen: It's not a real biggie, but somehow I would prefer to first get the value (as Object) and then do instance-of checks on it. That would allow untyped columns or even list subtypes (not nice, but can happen) to be properly handled.

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/RemoveSubstringTransformer.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public OutputColumns getOutputColumns() {
+        return new OutputColumns(String.class, baseColumn.getName() + " (subtracted)");
+    }
+
+    @Override
+    public String[] transform(final InputRow inputRow) {
+        String subtractedString = inputRow.getValue(baseColumn);
+        for (final InputColumn<?> inputColumn : substringColumns) {
+            if (inputColumn.getDataType() == List.class) {
```
kaspersorensen: See previous remark on same line :)

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/RemoveSubstringTransformer.java
```diff
@@ -0,0 +1,84 @@
...
+    InputColumn<?>[] substringColumns;
+
+    @Configured(value = "Match whole words only", required = false)
+    @Description("If set, only whole words (surrounded by whitespace or punctuation) will be removed.\n"
+            + " This prevents removing partial words.")
+    boolean wholeWordsOnly = false;
+
+    @Override
+    public OutputColumns getOutputColumns() {
+        return new OutputColumns(String.class, baseColumn.getName() + " (subtracted)");
```
kaspersorensen: how about also updating this: "[column name] (substring removed)"

kaspersorensen:
LGTM, merge if Travis is successful

kaspersorensen:
Oh a final thing - I'll create an icon :-)

Issue #982, opened 2015-12-03
Bad behavior during refreshing of component list in case of more remote servers


khouzvicka:
# Branch 5.0

## Description

In case of more remote servers and action "Refresh" on component list DataCleaner works crazy.

## Expected Behavior

"Refresh" of component list should work in all cases - one or more servers, doesnt matter.

## Observed Behavior

In this case I'm getting 4 different wrong errors:

Unexpected error!

Error dialog with a true comparison like "25 >= 23"
 Stacktrace:
 ERROR 14:31:31 DCUncaughtExceptionHandler - Thread Thread-1057 threw uncaught exception
 java.lang.ArrayIndexOutOfBoundsException: 25 >= 23
 at java.util.Vector.removeElementAt(Vector.java:558)
 at javax.swing.tree.VariableHeightLayoutCache$TreeStateNode.collapse(VariableHeightLayoutCache.java:1601)
 at javax.swing.tree.VariableHeightLayoutCache.treeStructureChanged(VariableHeightLayoutCache.java:658)
 at javax.swing.plaf.basic.BasicTreeUI$Handler.treeStructureChanged(BasicTreeUI.java:3917)
 at javax.swing.tree.DefaultTreeModel.fireTreeStructureChanged(DefaultTreeModel.java:580)
 at javax.swing.tree.DefaultTreeModel.reload(DefaultTreeModel.java:297)
 at org.datacleaner.widgets.tree.SchemaTree.componentDescriptorsUpdated(SchemaTree.java:656)
 at org.datacleaner.descriptors.AbstractDescriptorProvider$1.run(AbstractDescriptorProvider.java:268)
 at java.lang.Thread.run(Thread.java:745)

## Replication Steps

1. Run two monitors
 2. Configure these servers in config file of DataCleaner
 3. Open or create a job
 4. Try to fewtimes refresh the component list

PR #983, opened 2015-12-03
#982 - Resolution of problem with refresh of component list

khouzvicka:
Fix of issue #982

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
@@ -647,7 +647,7 @@ public class SchemaTree extends JXTree implements TreeWillExpandListener, TreeCe
     }
 
     @Override
-    public void componentDescriptorsUpdated() {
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
```diff
@@ -647,7 +647,7 @@ public Datastore getDatastore() {
     }
 
     @Override
-    public void componentDescriptorsUpdated() {
+    public synchronized void componentDescriptorsUpdated() {
```
kaspersorensen: Actually the issue here is not really one of concurrency, but one of worker thread responsibility. You need to ensure that the update of the tree happens in the Swing UI thread instead of whatever thread is now invoking this.

To do that, use the following:
```
WidgetUtils.invokeSwingAction(new Runnable() {
  // any UI modification code goes into the runnable here
});
```

kaspersorensen:
Closing this PR since it cannot be merged and the 5.0 branch is no longer the working space for this code (master is now based on 5.0 development)

PR #984, opened 2015-12-04
Use save handler in DC monitor and DC spark

kaspersorensen:
I wanted to use the AnalysisResultSaveHandler not only in Desktop DC but also in the Spark module and in DC monitor. This has the advantage of allowing a partial persistance of a DC result file in case some of the result elements fails to serialize.


LosD:
A little too much code in catch blocks, and a little too much duplication between them for my tastes, but we discussed it, and it's hard to do anything proper about, so LGTM.

PR #989, opened 2015-12-04
Issue #988: Made spark module capable of skipping result reduction

### LINKED ISSUE #988: Flag for not saving result when running Spark jobs
Sometimes you might want to create a job that should run on HDFS, but you're not interested in the result. Typically this could be because the job execution itself has side-effects that you want to keep (e.g. saving records somewhere) but the result (and the result reduction phase which can be substantial) is not needed.

kaspersorensen:
Fixes #988


ClaudiaPHI:
LGTM !

PR #990, opened 2015-12-04
catch the NoClassDefException exception

ClaudiaPHI:
PR for  NoClassDefException exception while scanning the packages. 

(The tests pass now)


kaspersorensen:
LGTM, although I would have preferred a multi-catch statement for brevity - but no biggie ;-)

PR #991, opened 2015-12-05
Feature/987 compare filter

### LINKED ISSUE #987: A "Compare" filter based on value, operator and operand
Quite often we have a validation requirement which involves comparing some values - either two values in the dataset or a value in the dataset and a configured value.

Examples:

 * Shipped date must be greater than order date
 * Total cost must be equal or greater than unit cost
 * Employee age must be greater than 15

For this we should create a filter which is similar to the Equals filter, but tuned towards dynamic choice of operator, and only a single operand choice (whereas Equals can have multiple "IN" values).

In more technical terms:

 * A ```@Configured InputColumn```  for the value to compare
 * A ```@Configured OperatorType``` for the operator (enum)
 * A ```@Configured(required=false) InputColumn``` for the operand (mutually exclusive to the next property)
 * A ```@Configured(required=false) String``` for the operand (mutually exclusive to the previous property)

kaspersorensen:
Fixes #987 

Here are some screenshots:

screen shot 2015-12-05 at 23 51 06
screen shot 2015-12-05 at 23 51 27


LosD:
LGTM, though the use of final in method signatures isn't entirely consistent.

Fix if you feel like it, otherwise just merge :)

PR #992, opened 2015-12-07
New Spark lifecycle listener

LosD:
This implements a new Spark lifecycle listener interface in support of configuring nodes from other clients than org.datacleaner.spark.Main


engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobLifeCycleListener.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.datacleaner.spark;
+
+import java.io.Serializable;
+
+public interface SparkJobLifeCycleListener extends Serializable{
+    void onNodeStart();
+    void onNodeEnd();
+    void onJobStart();
+    void onJobEnd();
```
LosD: I should probably add some JavaDoc to these, and to the Listener.

LosD:
Hmmmm... I seem to have developed the same itchy reformat trigger finger that I whined about recently, so the SparkJobContext.java file is a bit more changed than it should be. However, the change is pretty small (mostly moving a static function up) compared to the "real" changes, so I'll leave it in unless someone objects.

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -97,7 +97,7 @@ public static void main(String[] args) throws Exception {
                 }
             }
         } catch (Exception e) {
-            logger.error("Exception " + e.getStackTrace());
+            logger.error("Exception ", e);
```
kaspersorensen: Pretty sparse log message

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -54,11 +56,11 @@
  * {@link Serializable} properties.
  */
 public class SparkJobContext implements Serializable {
-
-    private static String DATA_CLEANER_RESULT_PATH_PROPERTY = "datacleaner.result.hdfs.path";
     public static final String ACCUMULATOR_CONFIGURATION_READS = "DataCleanerConfiguration reads";
     public static final String ACCUMULATOR_JOB_READS = "AnalysisJob reads";
+    private static final String DATA_CLEANER_RESULT_PATH_PROPERTY = "datacleaner.result.hdfs.path";
```
kaspersorensen: Muhaha you have a merge conflict now (this was renamed in #988)

kaspersorensen: No good deed goes unpunished ...

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobLifeCycleListener.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Also note that the serialization is one-way. You cannot retrieve anything that
+ * happened on the nodes through this listener.
+ */
+public interface SparkJobLifeCycleListener extends Serializable{
+    /**
+     * Triggered when a node starts processing a task. This will be executed
+     * on the nodes themselves.
+     */
+    void onNodeStart();
```
kaspersorensen: Actually it's not so much a "node" that is starting and ending with this event. It's really that the row processing on a node is starting and ending. Maybe in the future we will need more events for other stuff that is going on at the nodes, we would need to give it a more distinct name?

kaspersorensen:
LGTM, except for the blatant issues mentioned ;-D

LosD:
Ready for review again

kaspersorensen:
Great, will merge.

PR #993, opened 2015-12-07
fixed the problem with simple descriptor provider from listiq project

ClaudiaPHI:


kaspersorensen:
What a beautiful pull request - I would have written it the exact same way ;-)

ClaudiaPHI:
It was probably your clone or a doppelganger :-D

PR #1000, opened 2015-12-10
Feature/996 api for execute button options

### LINKED ISSUE #996: Add (minimal) API for adding execute options
In some distributions of DataCleaner we want to have specialized execution options such as "Execute on Hadoop" or "Execute on DataCleaner monitor". For this the normal drop-down menu next to the "Execute" button should be made pluggable.

We have done something similar already in the ResultWindow where we can plug in new "Save" options. I'm not sure whether this was the best facility we ended up making. Now that we are going to do it again, maybe first review if we can make it somehow smarter than before, since this is a pattern likely to be repeated again and again for more pluggable UI elements.

kaspersorensen:
Fixes #996


ClaudiaPHI:
LGTM!

PR #1001, opened 2015-12-10
Issue #923: Coalesce transformer preview fix

### LINKED ISSUE #923: Preview of "Fuse/Coalesce fields" returns null values
I've made a job like the one below:

image

But strange thing is that the fuse component works well when I run the job. It's just that the preview data more or less always yields __null__ instead of a coalesced value.

kaspersorensen:
Fixes #923 and related columnname serialization issue in coalesceunit


ClaudiaPHI:
LGTM!

PR #1003, opened 2015-12-10
Issue #986: Decreased the amount of sample records to store in Spark

### LINKED ISSUE #986: Large data amounts shuffled in partial results between spark nodes
We have seen that way too much data gets shuffled in the partial results of different spark nodes. We believe this is because the storage providers of each individual node is set to contain the default amount of sample records (500 sets of max 500 records). Probably we should change that storage setting so that we have less sample records per set on individual nodes.

kaspersorensen:
Fixes #986 
The new setting used here is to keep up to 500 (normal) sample sets of maximum 20 records each (decreased from 500)


ClaudiaPHI:
LGTM!

PR #1005, opened 2015-12-10
Making sure to close the Client object made for ElasticSearch connection

kaspersorensen:
This should ensure that we keep the client objects created for ElasticSearch under control - that we don't let them linger but close them once the datastore connection is closed.

__PLEASE AWAIT CONFIRMATION BEFORE MERGING__


kaspersorensen:
Confirmed. Merging.

PR #1007, opened 2015-12-14
Bring 5.0 branch into master

kaspersorensen:
This PR is here as a convenient way for us to align the master branch and the 5.0 branch. We wish to bring our working branches together and bring focus to the DataCleaner 5.0 delivery, but first it is suggested to do a general review of the 5.0 branch and bring risks, remarks and other stuff to light in this PR so that we get a controlled merging of branches.


api/src/main/java/org/datacleaner/configuration/RemoteServerConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+package org.datacleaner.configuration;
+
+import java.util.List;
+
+/**
+ * Object for data about all DataCleaner Monitor servers.
+ */
+public interface RemoteServerConfiguration {
+    public void setShowAllServers(boolean showAll);
+    public boolean showAllServers();
```
kaspersorensen: It's unclear to me (no javadoc) what the "show" is (where is it shown) and why a display would not show all of the servers.

api/src/main/java/org/datacleaner/descriptors/DescriptorProviderState.java
```diff
@@ -0,0 +1,78 @@
...
+    public Level getLevel() {
+        return level;
+    }
+
+    public String getMessage() {
+        return message;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
kaspersorensen: I suggest to simplify equals and hashCode by using the ```Objects.equals(...)``` and ```Objects.hash(...)``` methods.

desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
```diff
@@ -210,6 +211,10 @@ private IconUtils() {
         // prevent instantiation
     }
 
+    public static ImageIcon getIcon(String relativePath) {
```
kaspersorensen: I can't find any places in the code where this method is used - and it seems to be overlapping in functionality with ```ImageManager.get().getImageIcon(...)```

desktop/api/src/main/java/org/datacleaner/widgets/DCHtmlBox.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        super.setText(this.HTML_START_TAG + getTableHtml(text) + this.HTML_END_TAG);
+    }
+
+    private String getTableHtml(String content) {
+        return "<table border=\"0\" cellspacing=\"0\" cellpadding=\"0\"><tr><td>" + content + "</td></tr></table>";
+    }
+
+    private static class LinkRunner extends SwingWorker<Void, Void> {
```
kaspersorensen: Opening a link using a SwingWorker seems a bit exaggerated. SwingWorker is normally used for long-running stuff and we have opening of links in DC done in a blocking fashion without problems.

desktop/api/src/main/java/org/datacleaner/widgets/DCHtmlBox.java
```diff
@@ -0,0 +1,101 @@
...
+
+        private LinkRunner(URL url) {
+            if (url != null) {
+                this.url = url;
+            }
+        }
+
+        @Override
+        protected Void doInBackground() throws Exception {
+            Desktop.getDesktop().browse(url.toURI());
```
kaspersorensen: Normally we use the OpenBrowserAction class for this kind of stuff. Not sure it makes any difference, but maybe we'd want to only use one approach.

api/src/main/java/org/datacleaner/configuration/DataCleanerEnvironment.java
```diff
@@ -31,6 +31,13 @@
 public interface DataCleanerEnvironment {
 
     /**
+     * Gets the {@link RemoteServerConfiguration}.
+     *
+     * @return the box for credentials defined for this environment.
+     */
+    public RemoteServerConfiguration getRemoteServerConfiguration();
```
kaspersorensen: Can the server configuration be used for anything else than the descriptor provider? If not, then maybe it should not be exposed/contained within the ```DataCleanerEnvironment``` interface (and corresponding class) but only be an implementation detail of the ```RemoteDescriptorProviderImpl```.

As far as I can see, the server configuration isn't used for anything else?

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphTransformers.java
```diff
@@ -235,7 +236,16 @@ public Icon transform(Object obj) {
                 } else {
                     configured = componentBuilder.isConfigured(false);
                 }
-                return IconUtils.getDescriptorIcon(descriptor, configured, IconUtils.ICON_SIZE_LARGE);
+
+                Icon descriptorIcon = IconUtils.getDescriptorIcon(descriptor, configured, IconUtils.ICON_SIZE_LARGE);
+
+                if (descriptor instanceof RemoteTransformerDescriptor) {
```
kaspersorensen: Hmm maybe this should be a part of the above ```IconUtils.getDescriptorIcon``` method call? Then we would benefit from the overlay also in other places where the icon might be rendered. And it could also be correleated with the "configured" flag (which may already be false - yielding a 'double' overlay).

desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
```diff
@@ -103,6 +103,7 @@
     public static final String MENU_OPEN = "images/menu/open.png";
     public static final String MENU_NEW = "images/menu/new.png";
     public static final String MENU_EXECUTE = "images/menu/execute.png";
+    public static final String MENU_REFRESH = "images/menu/refresh.png";
```
kaspersorensen: There is already an identical ```ACTION_REFRESH``` icon. Do we need an extra for some reason?

api/src/main/java/org/datacleaner/configuration/RemoteServerData.java
```diff
@@ -0,0 +1,37 @@
...
+public interface RemoteServerData {
+    public RemoteServerData setServerPriority(Integer serverPriority);
+    public Integer getServerPriority();
+    public RemoteServerData setServerName(String serverName);
+    public String getServerName();
+    public RemoteServerData setHost(String host);
+    public String getHost();
+    public RemoteServerData setUsername(String username);
+    public String getUsername();
+    public RemoteServerData setPassword(String password);
```
kaspersorensen: I would prefer this interface to only expose getter methods so that an immutable implementation is also possible.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerDataImpl.java
```diff
@@ -0,0 +1,86 @@
...
+        return this;
+    }
+
+    @Override
+    public String getServerName() {
+        return serverName;
+    }
+
+    @Override
+    public RemoteServerDataImpl setHost(String host) {
```
kaspersorensen: I would prefer this to be immutable. Consider the scenario where it changes host. I cannot figure out what the side-effects are. But probably that already-loaded components will use the old host and next time a refresh is made, completely different components could be loaded. All configuration that is being actively used should be immutable IMO. Then if the user changes it, we create a new descriptor provider and load new descriptors. That's the only way to ensure we don't run into integrity issues between client and server.

engine/core/src/main/java/org/datacleaner/lifecycle/AssignConfiguredPropertiesHelper.java
```diff
@@ -28,7 +28,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class AssignConfiguredPropertiesHelper {
+public final class AssignConfiguredPropertiesHelper {
```
kaspersorensen: It doesn't matter much, but since I couldn't find any other reference to ```AssignConfiguredPropertiesHelper``` then I guess there is no need to make it public.

engine/core/src/main/java/org/datacleaner/util/batch/BatchRowCollectingTransformer.java
```diff
@@ -0,0 +1,87 @@
...
+
+import org.datacleaner.api.Close;
+import org.datacleaner.api.Initialize;
+import org.datacleaner.api.InputRow;
+import org.datacleaner.api.OutputRowCollector;
+import org.datacleaner.api.Provided;
+import org.datacleaner.api.Transformer;
+
+/**
+ * Added by jakub on 20.10.15
```
kaspersorensen: A bit of javadoc here would be nice. Took me a while to figure out the difference between this class and ```BatchTransformer```. In the end I guess it is that this one supports one-to-n transformations using an ```OutputRowCollector```.

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -61,25 +67,33 @@
...
-    public void init() {
-        logger.debug("Initializing '{}' @{}", componentDisplayName, this.hashCode());
-        client = new ComponentRESTClient(baseUrl, username, password);
+    @Initialize
+    public void initClient() throws RestrictedFunctionalityException {
+        try {
+            logger.debug("Initializing '{}' @{}", componentDisplayName, this.hashCode());
+            client = new ComponentRESTClient(baseUrl, username, password);
+        } catch (Exception e) {
+            throw new RestrictedFunctionalityException(
```
kaspersorensen: The ```RestrictedFunctionalityException``` is intended for situations where the user needs to buy or upgrade or something like that. In here it is thrown for every exception that may occur. That seems like a wrong mapping of exception types.

desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
jhorcicka: Well, I know there is ACTION_REFRESH constant pointing to images/actions/refresh.png. First of all I did not want to use ACTION_* for menu item creation in code since there are other MENU_* constants. I guess I can create another constant with same path like MENU_REFRESH = images/actions/refresh.png which is also a bit strange for me. That is why I did the copy of the file itself and put it into different directory.

api/src/main/java/org/datacleaner/configuration/RemoteServerData.java
jhorcicka: There is a list of these instances in OptionsDialog. The first instance can be updated via GUI dialog dynamically which (according to me) requires at least setUsername() and setPassword() to be available.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerDataImpl.java
jhorcicka: Setters for host, name a priority can be simply removed. Setters for credentials are, IMO, needed so that they can be updated via OptionsDialog.

kaspersorensen:
Regarding the discussion on immutability, OptionsDialog and RemoteServerData (which somehow got lost inbetween the commits - grrrr github): For me this is not a blocker, so we can leave it as is, but I would like to explain it because immutability is close to my heart and it's anyways something I think we will run into later:

My suggestion was to make the objects truly immutable. So no setters, and all fields final. The reason for this is that these objects are shared between many subprocesses and changing it is actually pretty dangerous - it might happen in the middle of a subprocess running in the background, yielding errors and cornercase scenarios at unpredictable places. And with a batch processing tool like DC this is very likely because you may have jobs running at the same time someone is changing the configuration.

Immutability is a common answer to this problem: Make sure that the data cannot mutate. If you need different data, create a copy with the changes incorporated in it. So from that point of view, if the user opens the OptionsDialog, then when saving the dialog I would create a completely new RemoteDescriptorProvider with completely new RemoteServerData objects. That way any lingering processes will not be impacted, but new job executions etc. will use the new descriptors.

kaspersorensen:
With #1015 fixed this PR is now ready for ___final review___ before we merge into master.

jhorcicka:
About the immutability... let's create a separate issue for this and continue in January.

kaspersorensen:
Actually I couldn't help myself ... Please review #1023 if you have a chance

PR #1014, opened 2015-12-16
HTTP request transformer

kaspersorensen:
This adds a transformer for invoking a HTTP request per record. In my opinion this is quite useful, although not a really planned issue, just something I did in my spare time and while listening to @wvholland blabber about how cool it would be ;-) So please review, give remarks etc.

One remark: I think we need a good component category for this. For now I used the "Reference data" category, which is a bit weird. But it puts it in the same spot as "Table lookup" which is kind of good because they both represent a kind of external call. But ideally I think we should maybe have a "Lookup and search" (or something - ideas welcome) category for these two instead.


components/http/src/main/java/org/datacleaner/components/http/HttpRequestTransformer.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Override
+    public Object[] transform(InputRow inputRow) {
+        final Charset usedCharset = Charset.forName(charset);
+
+        final String requestBody = applyVariablesToString(this.requestBody, inputRow);
+        final String url = applyVariablesToString(this.url, inputRow);
+
+        final HttpUriRequest request = method.createRequest(url);
+        if (requestBody != null && request instanceof HttpEntityEnclosingRequest) {
```
wvholland: I prefer additional "(" and ")" to prevent errors in order of execution and enhances readibility.
So :
if ((requestBody != null)) && (..))

PR #1019, opened 2015-12-20
Use SparkContext.hadoopConfiguration()

### LINKED ISSUE #1016: Ensure that all HdfsResource's in Spark module uses the Hadoop Configuration provided by spark
Everywhere that we currently use ```HdfsResource``` in the spark module, we need to instead use a new resource type. The new resource type needs to be fed with the ```Configuration``` coming from ```sparkContext.hadoopConfiguration()``` instead of creating it's own configuration. Experience have shown us that creating that configuration object is not trivial and may be externally facilitated by the Hadoop distributions that bundle spark themselves.

kaspersorensen:
This PR goes some way to fix #1016, but it is just the first step.

Fetching Hadoop 'Configuration' from SparkContext and use it for all resources in spark runner.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/guice/DCModuleImpl.java
+++ b/desktop/ui/src/main/java/org/datacleaner/guice/DCModuleImpl.java
@@ -19,9 +19,7 @@
  */
 package org.datacleaner.guice;
 
-import java.io.File;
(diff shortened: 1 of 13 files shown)


LosD:
LGTM

Issue #1022, opened 2015-12-22
Remote components interfaces -- immutability


jhorcicka:
This is an issue about refactoring mutable interfaces in remote components part. The following is Kasper's citation: 

Regarding the discussion on immutability, OptionsDialog and RemoteServerData (which somehow got lost inbetween the commits - grrrr github): For me this is not a blocker, so we can leave it as is, but I would like to explain it because immutability is close to my heart and it's anyways something I think we will run into later:

My suggestion was to make the objects truly immutable. So no setters, and all fields final. The reason for this is that these objects are shared between many subprocesses and changing it is actually pretty dangerous - it might happen in the middle of a subprocess running in the background, yielding errors and cornercase scenarios at unpredictable places. And with a batch processing tool like DC this is very likely because you may have jobs running at the same time someone is changing the configuration.

Immutability is a common answer to this problem: Make sure that the data cannot mutate. If you need different data, create a copy with the changes incorporated in it. So from that point of view, if the user opens the OptionsDialog, then when saving the dialog I would create a completely new RemoteDescriptorProvider with completely new RemoteServerData objects. That way any lingering processes will not be impacted, but new job executions etc. will use the new descriptors.

kaspersorensen:
No longer applies to CE

PR #1023, opened 2015-12-22
Options dialog split .. and more

kaspersorensen:
This PR started with me wanting to make the ```OptionsDialog``` a bit more componentized... But then it took off a bit. I ran into the concern about ```RemoteServerData``` being mutable and then into the fact that we have many different ways of formatting XML when saving conf.xml and the like. Oh well, in the end it's a big mash of things, sorry for that. But generally good changes I think.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/options/RemoteComponentsOptionsPanel.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/options/RemoteComponentsOptionsPanel.java
@@ -1,0 +1,179 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
+ * copy, or redistribute it subject to the terms and conditions of the GNU
+ * Lesser General Public License, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.widgets.options;
+
+import java.awt.Component;
+import java.awt.GridBagConstraints;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+import javax.swing.JButton;
+import javax.swing.JComponent;
+import javax.swing.JLabel;
+import javax.swing.JPasswordField;
+import javax.swing.JTextField;
+import javax.swing.event.DocumentEvent;
+
+import org.apache.metamodel.util.Resource;
+import org.datacleaner.configuration.DataCleanerConfiguration;
+import org.datacleaner.configuration.DataCleanerConfigurationImpl;
+import org.datacleaner.configuration.DataCleanerConfigurationUpdater;
+import org.datacleaner.configuration.RemoteServerConfiguration;
+import org.datacleaner.configuration.RemoteServerData;
+import org.datacleaner.configuration.RemoteServerDataImpl;
+import org.datacleaner.panels.DCPanel;
+import org.datacleaner.repository.RepositoryFile;
+import org.datacleaner.util.DCDocumentListener;
+import org.datacleaner.util.SecurityUtils;
+import org.datacleaner.util.WidgetFactory;
+import org.datacleaner.util.WidgetUtils;
+import org.datacleaner.widgets.DCHtmlBox;
+import org.datacleaner.windows.OptionsDialog;
+
+import com.google.common.base.Strings;
+
+/**
+ * The "Remote components" panel found in the {@link OptionsDialog}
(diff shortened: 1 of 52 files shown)


kaspersorensen:
Fixes #1022

desktop/ui/src/main/java/org/datacleaner/widgets/options/RemoteComponentsOptionsPanel.java
```diff
@@ -0,0 +1,197 @@
...
+        if (remoteServerConfiguration.getServerList().isEmpty()) {
+            remoteServerConfiguration.getServerList().add(newServerData);
+        } else {
+            remoteServerConfiguration.getServerList().remove(0);
+            remoteServerConfiguration.getServerList().add(0, newServerData);
+        }
+
+        final DataCleanerConfigurationUpdater configurationUpdater = new DataCleanerConfigurationUpdater(
+                getDataCleanerConfigurationFileResource());
+        configurationUpdater.update("descriptor-providers:remote-components:server:username", username);
```
kaspersorensen: To be honest I don't understand how this will work and I couldn't see it working locally either. Maybe someone like @jhorcicka can help me out with this particular aspect?

kaspersorensen: Ah, figured it out and fixed my code in 7469836

desktop/ui/src/main/java/org/datacleaner/widgets/options/RemoteComponentsOptionsPanel.java
```diff
@@ -0,0 +1,197 @@
...
+        passwordTextField.getDocument().addDocumentListener(documentListener);
+        setSize(passwordTextField);
+
+        setTitledBorder("Credentials");
+        setupFields();
+        addAllFields();
+
+    }
+
+    // TODO: Check necesity
```
LosD: Is this for later, or did you forget to check it?

desktop/ui/src/main/java/org/datacleaner/widgets/options/RemoteComponentsOptionsPanel.java
LosD: Not really for this change, but why in the world isn't this using XPath instead of a homegrown format?

desktop/ui/src/main/java/org/datacleaner/widgets/options/RemoteComponentsOptionsPanel.java
kaspersorensen: I needed to fix it! :-) Doing so now.

desktop/ui/src/main/java/org/datacleaner/widgets/options/RemoteComponentsOptionsPanel.java
kaspersorensen: Yea that would be quite a lot prettier ... In general I don't like that "Updater" class because it is just a glorified DOM editor. It has not knowledge about the XSD of conf.xml. So for now I consider it just throw-away-ware until something better comes along.

jhorcicka: Because I was not aware of XPath. :-(

LosD: Ah. Yeah, let's take a look at it later, it should be a reasonably simple change.

kaspersorensen: We can improve this later. I think it should eventually be merged somehow with the ```DatastoreXmlExternalizer``` - and be something which encapsulates the details of the conf.xml file format. We also eventually need something to update dictionaries and synonym catalogs in conf.xml. Plenty of work ahead of us in that area ;-)

engine/xml-config/src/test/java/org/datacleaner/configuration/DatastoreXmlExternalizerTest.java
```diff
@@ -108,11 +101,12 @@ public void testIsExternalizableAndExternalize() throws Exception {
         final String str = transform(elem);
 
         final char sep = File.separatorChar;
-        assertEquals("<csv-datastore name=\"foo\">" + "<filename>src" + sep + "test" + sep + "resources" + sep
-                + "example-dates.csv</filename><quote-char>\"</quote-char>"
-                + "<separator-char>,</separator-char><escape-char>\\</escape-char><encoding>UTF-8</encoding>"
-                + "<fail-on-inconsistencies>true</fail-on-inconsistencies><multiline-values>true</multiline-values>"
-                + "<header-line-number>1</header-line-number></csv-datastore>", str);
+        assertEquals("<csv-datastore name=\"foo\">\n" + "  <filename>src" + sep + "test" + sep + "resources" + sep
```
LosD: Speaking of XPath. A Hamcrest XPath matcher would a much more understandable test than this... String nightmare :)

LosD:
Code LGTM. Has it been (manually) tested?

kaspersorensen:
Yes I have tested it manually.

engine/xml-config/src/test/java/org/datacleaner/configuration/DatastoreXmlExternalizerTest.java
kaspersorensen: Agree!!! I knew this would be fround upon at the review ;-)

kaspersorensen:
Merging then :)

LosD:
:+1:

PR #1025, opened 2015-12-22
Feature/1018 load into memory xml element

### LINKED ISSUE #1018: DatastoreDictionary: Add element to conf.xml for setting the loadIntoMemory flag
Today we default to ```true``` for the ```DatastoreDictionary._loadIntoMemory``` property. I think the default is good, but we don't a way to configure it via conf.xml.

kaspersorensen:
Fixes #1018


kaspersorensen:
Wrong target branch, closing

Issue #1027, opened 2015-12-22
Unwanted http requests on job window closing with "discard changes". 


jhorcicka:
#### Description 
A job with a remote component in it. When trying to close the job window with "discard changes" option there are additional/unwanted HTTP requests to get output columns of the remote component. When the remote server is down, this can freeze the GUI for some time. 

#### Suggestion1
Presence of a remote component should be checked and when the remote server is down empty list of output columns should be returned. The event should be logged. 

#### Suggestion2
Similar to the first suggestion but instead of returning the empty list an exception should be raised and showed in GUI error dialog.

PR #1028, opened 2015-12-22
Bug/1027 unwanted requests on job window closing

jhorcicka:
Fix for a bug #1027. 

A job with a remote component in it. When trying to close the job window with "discard changes" option there are additional/unwanted HTTP requests to get output columns of the remote component. When the remote server is down, this can freeze the GUI for some time. 

### Question
How to deal with the problem? 

I logged the event and returned empty list of output columns. 

Alternative would be to throw an IllegalStateException. I tried it and it causes another problem. After closing job window there is a dialog with option "discard changes". When this option is clicked, the exception dialog window appears but even after closing this exception dialog the job window is not closed. In this situation I did not find the way how to get rid of the "broken" job window. There already is a "throw new IllegalStateException" in the TransformerComponentBuilder.getOutputColumns method. This causes the same problem (but probably the condition has never occurred or the bug was ignored).

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
+++ b/engine/core/src/main/java/org/datacleaner/job/builder/AnalysisJobBuilder.java
@@ -1192,11 +1192,21 @@ public final class AnalysisJobBuilder implements Closeable {
     }
 
     public void removeAllSourceColumns() {
-        final List<MetaModelInputColumn> sourceColumns = new ArrayList<>(_sourceColumns);
-        for (MetaModelInputColumn inputColumn : sourceColumns) {
-            removeSourceColumn(inputColumn);
+        final Collection<ComponentBuilder> componentBuilders = getComponentBuilders();
+
+        for (ComponentBuilder componentBuilder : componentBuilders) {
+            componentBuilder.clearInputColumns();
         }
-        assert _sourceColumns.isEmpty();
+
+        final List<SourceColumnChangeListener> listeners = new ArrayList<>(_sourceColumnListeners);
+
+        for (SourceColumnChangeListener listener : listeners) {
+            for (MetaModelInputColumn inputColumn : _sourceColumns) {
+                listener.onRemove(inputColumn);
+            }
+        }
+
+        _sourceColumns.clear();
     }
 
     public void removeAllAnalyzers() {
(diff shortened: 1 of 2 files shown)


kaspersorensen:
Actually when I read what the problem is, I am thinking "but it shouldn't try to get/refresh output columns at that time". I realize that probably is does so right now. Maybe we can prevent that by at least checking that the transformer's configuration hasn't changed. Would be best to try and keep web server traffic low, and especially in a case like this where you're saving your job - you would always want to be able to do that, regardless of what the server status is.

jhorcicka:
Here is the stack trace. 
stacktrace

jhorcicka:
AnalysisJobBuilder.removeAllSourceColumns changes configuration for each source column. That's why TransformerComponentBuilder.getOutputColumns is called multiple times (for each source column that is removed). In the case of a remove component there is always the same http request for output columns, I guess.

kaspersorensen:
Could we make a change such that removing all columns is done in one go, and then automatically the component is no longer having a valid configuration (at which point I think we will anyways not request output columns because the configuration is incomplete)?

LosD:
Isn't the larger problem that getOutputColumns has a _huge_ side effect in remote components? Shouldn't they have a proper resolution stage/phase/time/whatever, instead of changing the contract of a standard Component method?

jhorcicka:
I can move the server availability check from TransformerComponentBuilder into RemoteTransformer.getOutputColumns. That way it is "remote components" responsibility.

LosD:
Hmmm... I'm not sure if that is really an improvement. Honestly, I haven't checked out the whole system deep enough to really have a proper opinion, it just seems a bit extreme that a simple getter starts contacting servers (and plenty of parts of the system has no idea what kind of component it is working with).

If it is unavoidable (or hugely complicates or worsens things), that is of course how it is, and this is the lesser evil. :)

jhorcicka:
Just a remark from investigation. Even with the server availability check REST client can be initialized and it can send requests. This is caused by the timeout interval and cache. 

So, when I want to close the job window "too early" after the server is down, the delay still takes place. In cache there is still "sever is up value". Client tries to send the request but the response never comes so it waits for timeout to expire.

LosD:
As a workaround, is the threat available, and could it be interrupted (```Thread.interrupt()```) without ill effect (if the waiting code does not reloop after InterruptExceptions, it should at least be possible to interrupt it)?

jhorcicka:
Which thread do you mean?

jhorcicka:
I tried to remove all source columns "at once" in AnalysisJobBuilder.reset/removeAllSourceColumns. It fixes the problem with requests on job window closing. 

Tests went fine. However, I don't know what all might have been affected by this change. Please take a look at it and let me know what's wrong.

LosD:
I meant the thread that is doing the request, that one should be interruptible. But that may not be necessary with your latest change. It scares the hell out of me, though... But let's look at it after Christmas. :)

LosD:
Okay, we took a good hard look, and it seems to be pretty safe :)

PR #1029, opened 2015-12-22
Assemble spark module for mapr compatiblity

kaspersorensen:
I've been spending most of my day making sure we are compatible with Spark running on MapR (using the MapR 5.0.0 sandbox).

This PR unlocks that capability. It ensures that the new ```HdfsHelper``` is in control at all the important stages where resources/paths are loaded. It also makes HDFS the default scheme to use in the Spark module when parsing resource paths without any scheme defined.

### DIFF
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
@@ -19,6 +19,7 @@
  */
 package org.datacleaner.spark;
 
+import java.net.URI;
 import java.util.Arrays;
 import java.util.Map;
 
@@ -49,12 +50,12 @@ public class Main {
         final SparkConf conf = new SparkConf().setAppName("DataCleaner-spark");
         final JavaSparkContext sparkContext = new JavaSparkContext(conf);
 
-        final String confXmlPath = args[0];
-        final String analysisJobXmlPath = args[1];
+        final URI confXmlPath = URI.create(args[0]);
+        final URI analysisJobXmlPath = URI.create(args[1]);
 
-        final String propertiesPath;
+        final URI propertiesPath;
         if (args.length > 2) {
-            propertiesPath = args[2];
+            propertiesPath = URI.create(args[2]);
         } else {
             propertiesPath = null;
         }
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
@@ -23,7 +23,7 @@ import java.util.Arrays;
 import java.util.Map;
 
 import org.apache.commons.lang.SerializationException;
-import org.apache.metamodel.util.HdfsResource;
+import org.apache.metamodel.util.Resource;
 import org.apache.spark.SparkConf;
 import org.apache.spark.api.java.JavaSparkContext;
 import org.datacleaner.api.AnalyzerResult;
@@ -59,49 +59,48 @@ public class Main {
             propertiesPath = null;
         }
 
-        final SparkJobContext sparkJobContext = new SparkJobContext(confXmlPath, analysisJobXmlPath,
-                propertiesPath, sparkContext.hadoopConfiguration());
-
-        final String resultJobFilePath;
-        if (sparkJobContext.isResultEnabled()) {
-            // get the path of the result file here so that it can fail fast(not
-            // after the job has run).
-            resultJobFilePath = ResultFilePathUtils.getResultFilePath(sparkContext, sparkJobContext);
-            logger.info("DataCleaner result will be written to {}", resultJobFilePath);
-        } else {
-            resultJobFilePath = null;
-        }
+        final SparkJobContext sparkJobContext = new SparkJobContext(confXmlPath, analysisJobXmlPath, propertiesPath,
+                sparkContext);
 
         final SparkAnalysisRunner sparkAnalysisRunner = new SparkAnalysisRunner(sparkContext, sparkJobContext);
         try {
             final AnalysisResultFuture result = sparkAnalysisRunner.run();
-            if (resultJobFilePath != null) {
-                final HdfsResource hdfsResource = new HdfsResource(resultJobFilePath);
-                final AnalysisResultSaveHandler analysisResultSaveHandler = new AnalysisResultSaveHandler(result,
-                        hdfsResource);
-                try {
-                    analysisResultSaveHandler.saveOrThrow();
-                } catch (SerializationException e) {
-                    // attempt to save what we can - and then rethrow
-                    final AnalysisResult safeAnalysisResult = analysisResultSaveHandler.createSafeAnalysisResult();
-                    if (safeAnalysisResult == null) {
-                        logger.error("Serialization of result failed without any safe result elements to persist");
-                    } else {
-                        final Map<ComponentJob, AnalyzerResult> unsafeResultElements = analysisResultSaveHandler
-                                .getUnsafeResultElements();
-                        logger.error("Serialization of result failed with the following unsafe elements: {}",
-                                unsafeResultElements);
-                        logger.warn("Partial AnalysisResult will be persisted to filename '{}'", resultJobFilePath);
 
-                        analysisResultSaveHandler.saveWithoutUnsafeResultElements();
-                    }
+            result.await();
 
-                    // rethrow the exception regardless
-                    throw e;
-                }
+            if (sparkJobContext.isResultEnabled()) {
+                final Resource resultResource = ResultFilePathUtils.getResultResource(sparkContext, sparkJobContext);
+                logger.info("DataCleaner result will be written to: {}", resultResource);
+                saveResult(result, resultResource);
+            } else {
+                logger.info("DataCleaner result will not be written - disabled");
             }
         } finally {
             sparkContext.stop();
+        }
+    }
+
+    private static void saveResult(AnalysisResultFuture result, Resource resultResource) {
+        final AnalysisResultSaveHandler analysisResultSaveHandler = new AnalysisResultSaveHandler(result, resultResource);
+        try {
+            analysisResultSaveHandler.saveOrThrow();
+        } catch (SerializationException e) {
+            // attempt to save what we can - and then rethrow
+            final AnalysisResult safeAnalysisResult = analysisResultSaveHandler.createSafeAnalysisResult();
+            if (safeAnalysisResult == null) {
+                logger.error("Serialization of result failed without any safe result elements to persist");
+            } else {
+                final Map<ComponentJob, AnalyzerResult> unsafeResultElements = analysisResultSaveHandler
+                        .getUnsafeResultElements();
+                logger.error("Serialization of result failed with the following unsafe elements: {}",
+                        unsafeResultElements);
+                logger.warn("Partial AnalysisResult will be persisted to filename '{}'", resultResource.getQualifiedPath());
+
+                analysisResultSaveHandler.saveWithoutUnsafeResultElements();
+            }
+
+            // rethrow the exception regardless
+            throw e;
         }
     }
 }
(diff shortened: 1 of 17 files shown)


engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -59,49 +59,48 @@ public static void main(String[] args) throws Exception {
...
-                    }
+            result.await();
 
-                    // rethrow the exception regardless
-                    throw e;
-                }
+            if (sparkJobContext.isResultEnabled()) {
+                final Resource resultResource = ResultFilePathUtils.getResultResource(sparkContext, sparkJobContext);
+                logger.info("DataCleaner result will be written to: {}", resultResource);
+                saveResult(result, resultResource);
```
LosD: Haven't we lost fail-fast abilities by moving this here?

engine/env/spark/src/main/java/org/datacleaner/spark/functions/RowProcessingFunction.java
```diff
@@ -69,8 +71,8 @@
  * This class implements two interfaces because it has two (quite similar)
  * styles of usages in the {@link SparkAnalysisRunner}.
  */
-public final class RowProcessingFunction implements
-        Function2<Integer, Iterator<InputRow>, Iterator<Tuple2<String, NamedAnalyzerResult>>>,
+public final class RowProcessingFunction
+        implements Function2<Integer, Iterator<InputRow>, Iterator<Tuple2<String, NamedAnalyzerResult>>>,
         PairFlatMapFunction<Iterator<InputRow>, String, NamedAnalyzerResult> {
```
LosD: This file is full of strange reformatting.

engine/env/spark/src/main/java/org/datacleaner/spark/utils/ResultFilePathUtils.java
```diff
@@ -19,86 +19,54 @@
...
+        final HdfsHelper hdfsHelper = new HdfsHelper(sparkContext);
+
+        final String resultPath;
+        if (Strings.isNullOrEmpty(sparkJobContext.getResultPath())) {
+            resultPath = DEFAULT_RESULT_PATH + '/' + generateResultFilename(sparkJobContext);
         } else {
-            final URI uri = URI.create(resultPath);
-            if (!uri.isAbsolute()) {
-                resultPath = createPath(fileSystemPrefix, resultPath);
+            if (hdfsHelper.isDirectory(sparkJobContext.getResultPath())) {
```
LosD: Ew. Strings everywhere.

This seems like a huge step backwards to me. We should go to URIs where ever we can (and that isn't in a tight loop, URIs are a little heavier), not the other way.

LosD:
Formatting seems to be off everywhere, and I really, really hate stupid Strings that represents URIs unless it is absolutely necessary (and it doesn't seem to be in this case. Unix paths are fully legal relative URIs).

engine/env/spark/src/main/java/org/datacleaner/spark/utils/ResultFilePathUtils.java
kaspersorensen: But there's nothing here that would benefit from being an URI as far as I can see? The logic here is about checking whether something is a directory or not - which requires actually inquiring the underlying file/resource system. And the other part is about generating a filename with a timestamp etc. - which would be exactly the same when building a URI.

kaspersorensen: Note that there _used to be_ something related to URIs, namely the hdfs scheme or whatever. But that part is no longer there, intentionally, to delegate as much as possible to the Resource resolution mechanisms underneath.

engine/env/spark/src/main/java/org/datacleaner/spark/functions/RowProcessingFunction.java
kaspersorensen: It's pretty strange - I have imported the ```Formatter-Eclipse.xml``` file. It does it like this on my Mac, but I think (given experience) that it will be OK on my Windows machine.

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
kaspersorensen: You're right. The Resource resolution should be moved up.

engine/env/spark/src/main/java/org/datacleaner/spark/utils/ResultFilePathUtils.java
LosD: That seems to me like wanting to use dictionary instead of a proper class representing the data, just because there currently isn't a direct benefit. It's a URI, I really think we should use URI objects for those, or maybe even the (URI backed) ```org.apache.hadoop.fs.Path``` (I believe all  filesystems on Hadoop must use these?), to get even more precise in what it is we are passing around.

For one thing we get proper URI checking right away instead of randomly failing when it proves that it was a bad URI when it needs to be used later. Of course that doesn't fix all wrong entries, but we should catch as much as we can, as early as we can.

kaspersorensen: I agree with checking it early on, so that will be done.
But the trouble is really that we don't _always_ want to use a hadoop Path. For one thing, during testing we would not want to use it like that because we use the local file system (or in other words: We will resolve a FileResource instead of a HdfsResource/HadoopResource). The problem we have here is not about whether it is a URI or a String in my opinion. We don't use URI for anything - there are no longer any uses of "get scheme" or similar behaviour. The only thing that is left is to figure out if it is a file or a directory, and to check if we need to append the file extension (.analysis.result.dat)

kaspersorensen: Early validation of the result file path has now been added as of commit  9b731f6 - this is not done in the Main class, but rather in SparkJobContext since this is also where the paths for the job and configuration file is done. And it is done at construction time, which would make it fail fast always.

LosD: I still don't agree at all, but I don't seem to be able to convince you :)

LosD: (For the record, I even think we should be using URI's everywhere in non-Hadoop cases. Windows paths should be converted to file:// the moment we see them, non-file and non-hostname should use opaque URI's (i.e. jdbc:derby:mydb) and so on and so forth)

kaspersorensen: I think using file:// would lead most interpreters to think that we're using absolute file paths, although they're most often relative to DCs home directory. In addition, not specifying the scheme allows the application to have a default scheme, which is also what HDFS does when it allows taking a path like /path/to/file and then deciding, based on configuration, where to resolve that. I think we are doing exactly the same here.

LosD: A legal URI does not require a scheme (and Hadoop uses that fact... All Hadoop paths are URIs)

kaspersorensen: Looking at the javadoc there are 5+ constructors for Path and only one of them takes a URI. That's extremely close to the characteristics of MetaModel's Resource implementations.

LosD: Take a look at the class itself. Everything becomes an URI (the only non-static field)

kaspersorensen: Ok understood. But that's (probably a fine) implementation detail. Similarly we should probably use URI more in the Resource implementations. But it doesn't change the fact that _this_ code is not using URI, rather it is essentially building one by appending necessary details to it. I don't see the problem?

LosD: It's not really an implementation detail. java.io.file.Path/FileSystem does somewhat the same (though a little looser).

It's not a direct problem here. It's the general idea of passing strings around when we can use a  narrower class I really don't like. And I think it is something we already do too much, so going towards more of that instead of away from it makes my nose go all wrinkly ;)

kaspersorensen: Hmm after much thought, and while generally agreeing on your last statement (passing strings around is inferior to a proper type), here's my suggestion then: The problem needs to be solved not here, but then at the interface of SparkJobContext. This class returns the result path as a String, so the problem "starts" there. On the other hand, this remark was made in the Main class, which has a very string oriented interface (```main(String[] args)```) so in _this_ class I would never raise any alarm over handling strings because it's the intrinsic nature of a main class. But let's change it so that all the non-main classes (SparkJobContext in particular, and then also HdfsHelper I guess) deals with URI.
The other aspect to the discussion is about your general feeling about this in relation to loading of resources. And again I principally agree, but an important fact is missing: That the resources are loaded using a ```Converter``` implementation (ResourceConverter) and a Converter is _exactly_ for converting something to/from Strings (usually in relation to configured property values serialized in XML - but in this case also command line arguments). What we _could_ do is to add also a (non-interface) method to ResourceConverter which converts a URI to a Resource. That's totally OK with me, but it doesn't mean that the String-variant is not relevant - it really is, but just not for standard programmatic access like you see it, but for more polymorphic use (we have many Converters for all the different types of property values).

kaspersorensen:
As per discussion - HdfsHelper and SparkJobContext updated to work on URI instead of Strings. Can you please check it again when convenient, grand reviewer @LosD ;-)

kaspersorensen:
:christmas_tree: :santa: :snowman: :gift:

LosD:
I love it! :)

Still problems with the reformatting, but unless we need to merge before returning to work, I guess this can be fixed with the office computer. But it's weird... Could it be the Eclipse version that is different, and the exported formatting being incompatible (and Eclipse for some reason swallowing the error)? It's extremely weird that the build OS could change it so much.

engine/env/spark/src/main/java/org/datacleaner/spark/functions/RowProcessingFunction.java
kaspersorensen: I am using Eclipse Mars on my Mac, which is never than the eclipse version (Juno I think) that was used when creating the formatter file.
I guess it It might be new features of eclipse's formatter. Looking at it on Mac I see the line-wrapping feature (per element type, in this case the 'implements' clause) in the formatter which probably makes the difference:

screen shot 2015-12-26 at 22 16 56

kaspersorensen: Seems also the ```Prefer wrapping outer expressions (keep nested expressions on one line)``` makes the difference. I will adjust in the coming commit(s).

kaspersorensen:
Strange - I tried to export my eclipse formatter settings on Mac and make a branch where they're pushed to Formatter-Eclipse.xml. But the diff is not that interesting - I was expecting that maybe they'd introduced a bunch of new properties which we hadn't set in the old config.

Commit is here for reference: 826d96ffe08745a804a740a9e247caae1c8f9b02

LosD:
Hmmmm... [wrap_outer_expressions_when_nested](help.eclipse.org/juno/topic/org.eclipse.jdt.doc.isv/reference/api/org/eclipse/jdt/core/formatter/DefaultCodeFormatterConstants.html#FORMATTER_WRAP_OUTER_EXPRESSIONS_WHEN_NESTED) seems to me to be some of the things we see?

LosD:
Wait! This actually looks like it might be because it is returning to the "proper" formatting now:
For example:
image
I'm almost sure that the old symptom of your working from home was that it would wrap stuff that it shouldn't have, but here it is opposite. Your imports may have worked, now the resulting change just took us by surprise. :)

kaspersorensen:
Upgraded my Windows Eclipse to "Mars" instead of "Luna" and now the formatting is consistent. So let's recommend upgrading Eclipse rather than trying to make the old eclipse versions work consistently.

PR #1036, opened 2016-01-05
Issue #949: Fixed by applying MetaModel "streaming" AggregateBuilders

### LINKED ISSUE #949: Improve "Grouper" component to be less memory intensive
The new grouper component (see #924 and #946) is functionally working like we want it to, but currently takes up a lot of memory because it keeps all values pertaining to all groups in a big collection.

A lot of aggregations don't need all the values however. So a clear improvement would be that we instead make something like an AggregationBuilder (consider the MetaModel interface of same name) to build the specific aggregation in the most appropriate way without wasting memory on stuff that is not needed.

kaspersorensen:
Fixes #949 
And also adds a bunch of new aggregations such as SUM, AVG, LAST and RANDOM. During an experiment today @LosD and I found that such transformations could be quite beneficial.


components/grouper/src/main/java/org/datacleaner/components/group/SortationType.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private SortationType(String name) {
+        _name = name;
+    }
+
+    @Override
+    public String getName() {
+        return _name;
+    }
+}
```
LosD: This would love a newline :)

LosD:
Nice!

A little bit much Object and unchecked casts for my tastes, but when there's no common collection interface for TreeSet and TreeMap, there's probably no way around that without doing something huge like building our own sorted map implementation (where all the unchecked casts would probably have to hide anyway), so LGTM!

kaspersorensen:
I had the same feeling wrt. the casting. But every alternative I tried ended up being more ugly ;-)

PR #1037, opened 2016-01-06
Support Cli Spark running and lifecycle listeners

### LINKED ISSUE #1013: Run Spark jobs from CLI
Currently a manual spark-submit is required to run a Spark job. As we are preparing for full running from the GUI, a nice step along the way is making the CLI able to launch the Spark jobs (i.e. using org.datacleaner.spark.ApplicationDriver).

Implement a minimal way of doing just that.

LosD:
This will add support for running Spark jobs from the command line. It uses ApplicationDriver in the background, so this should also cover at least _some_ of the groundwork for GUI running.

A service provider interface has been added to support integrating "Spark-DC" into other products, for example to facilitate license checks or logging, or other life cycle checks.

Fixes #1013 

(Sorry about the branch naming... Had a few issues)


desktop/ui/pom.xml
```diff
@@ -182,6 +187,75 @@
 			<groupId>org.freemarker</groupId>
 			<artifactId>freemarker</artifactId>
 		</dependency>
+
+		<!-- Engough Hadoop/Spark for uploading and submitting jobs -->
```
kaspersorensen: Not that it matters much, but there's a typo in "enough" :)

desktop/ui/pom.xml
```diff
@@ -182,6 +187,75 @@
...
+		<dependency>
+			<groupId>org.apache.spark</groupId>
+			<artifactId>spark-launcher_2.11</artifactId>
+			<version>${spark.version}</version>
```
kaspersorensen: I guess we should have this dependency in dependencyManagement so that you don't need to specify the version or the exclusions

desktop/ui/pom.xml
```diff
@@ -182,6 +187,75 @@
...
+				<exclusion>
+					<groupId>org.spark-project.spark</groupId>
+					<artifactId>unused</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.hadoop</groupId>
+			<artifactId>hadoop-hdfs</artifactId>
+			<version>${hadoop.version}</version>
```
kaspersorensen: Same with this dependency

desktop/ui/src/main/java/org/datacleaner/bootstrap/Bootstrap.java
```diff
@@ -163,6 +164,11 @@ private void runInternal() throws FileSystemException {
             LookAndFeelManager.get().init();
         }
 
+        if(arguments.getRunType() == CliRunType.SPARK){
```
kaspersorensen: Missing space before (

desktop/ui/src/main/java/org/datacleaner/cli/CliRunType.java
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.cli;
+
+public enum CliRunType {
+    DESKTOP, SPARK
```
kaspersorensen: Is this really a "CLI run type" or rather a "execution platform"? I'm thinking that maybe we should replace "DESKTOP" with "LOCAL".

(and maybe in the future there could also be an option like "MONITOR" if one wanted to instrument the DC monitor to run a job there or so.)

engine/env/spark/src/main/java/org/datacleaner/spark/SparkRunner.java
```diff
@@ -0,0 +1,101 @@
...
+import java.net.URI;
+import java.util.UUID;
+
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.spark.launcher.SparkLauncher;
+import org.datacleaner.spark.utils.HadoopUtils;
+
+/**
+ * Prepares for Spark launching from
```
kaspersorensen: ... aaaand the sentence ended ;-)

engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopUtils.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.InputStream;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.metamodel.MetaModelException;
+import org.apache.metamodel.util.FileHelper;
+
+import com.google.common.base.Strings;
+
+public class HadoopUtils {
```
kaspersorensen: Maybe we should call this something like HadoopClientUtils because it is really only valid for cases where there is a namenode (e.g. not MapR) or so I think ... At least my experiments so far failed to show that a client on MapR could create a configuration like this. But this stuff works with "vanilla" Hadoop and also Hortonworks - maybe also Cloudera.

engine/env/spark/src/main/java/org/datacleaner/spark/ApplicationDriver.java
```diff
@@ -159,34 +174,42 @@ public SparkLauncher createSparkLauncher(File hadoopConfDir, String configuratio
...
+        sparkLauncher.addAppArgs(toHadoopPath(configurationHdfsPath));
+        sparkLauncher.addAppArgs(toHadoopPath(jobHdfsPath));
+
+        if (!StringUtils.isNullOrEmpty(resultHdfsPath)) {
+            Properties properties = new Properties();
+            properties.setProperty("datacleaner.result.hdfs.path", resultHdfsPath);
+            File tempFile = File.createTempFile("job-", ".properties");
+            properties.store(new FileWriter(tempFile), "DataCleaner Spark runner properties");
+            final URI uri = copyFileToHdfs(tempFile,
+                    _fileSystem.getHomeDirectory().toUri().resolve("temp/" + tempFile).toString());
```
kaspersorensen: Nice touch with the home-dir! We should consider using that more ... But not a matter for this PR ;-)

engine/env/spark/src/main/java/org/datacleaner/spark/Main.java
```diff
@@ -63,6 +64,13 @@ public static void main(String[] args) throws Exception {
         final SparkJobContext sparkJobContext = new SparkJobContext(confXmlPath, analysisJobXmlPath, propertiesPath,
                 sparkContext);
 
+        final ServiceLoader<SparkJobLifeCycleListener> listenerLoaders =
+                ServiceLoader.load(SparkJobLifeCycleListener.class);
+
+        for(SparkJobLifeCycleListener listener : listenerLoaders){
```
kaspersorensen: space before "(" :)

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobLifeCycleListener.java
```diff
@@ -36,21 +36,21 @@
      * Triggered when a node starts processing a task. This will be executed
      * on the nodes themselves.
      */
-    void onPartitionProcessingStart();
+    void onPartitionProcessingStart(SparkJobContext sparkJobContext);
```
kaspersorensen: Could we add the partition index and the number of partitions as arguments here? Only if it's low-hanging fruit ...

engine/env/spark/src/main/java/org/datacleaner/spark/SparkRunner.java
```diff
@@ -0,0 +1,101 @@
...
+    private final ApplicationDriver _applicationDriver;
+    private final URI _configurationFile;
+    private final URI _jobFile;
+    private final URI _resultFile;
+    private final File _hadoopConfigurationDirectory;
+
+    public SparkRunner(String configurationFilePath, String jobFilePath, String resultFilePath) throws IOException {
+        try {
+            _hadoopConfigurationDirectory = HadoopUtils.getHadoopConfigurationDirectoryToUse();
+            _hadoopDefaultFS = FileSystem.newInstance(HadoopUtils.getHadoopConfiguration(_hadoopConfigurationDirectory));
```
kaspersorensen: Seems to me like the HadoopUtils class should have a single method for getting this FileSystem ... At least it seems like intricate code that, if we need to change it, would be difficult to ensure we get it right everywhere.

desktop/ui/src/main/java/org/datacleaner/bootstrap/Bootstrap.java
LosD: And before { (or rather, too many spaces everywhere else ;))

desktop/ui/src/main/java/org/datacleaner/cli/CliRunType.java
LosD: The Cli part is simply following the same pattern as other Cli argument enums.

Execution platform implies that it is always a complete platform change. I'm not sure all run types will be platform changes? We could also go for "execution mode" instead.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobLifeCycleListener.java
LosD: Yeah, I guess no problem... Though I don't where to get them. You can show me when you get here.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkRunner.java
LosD: It's there to give the reader a sense of mystery. :)

engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopUtils.java
LosD: But shouldn't we cover both cases? I mean, _something_ needs to decide what to do in the MapR case... Right now the rest of my code expects to be able to just call this. Should we make that determination somewhere else?

engine/env/spark/src/main/java/org/datacleaner/spark/SparkRunner.java
LosD: Makes sense

desktop/ui/src/main/java/org/datacleaner/cli/CliRunType.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.cli;
+
+public enum CliRunType {
```
LosD: See outdated comment from @kaspersorensen (only part of the comment has been fixed)

LosD:
Test failures seems unrelated. Some remote component thing.

LosD:
I'll retract that. The actual error is the expected list of arguments when CLI is doing the usage list. It was just hidden right after a lot of remote component exceptions.

desktop/ui/src/main/java/org/datacleaner/cli/CliRunType.java
kaspersorensen: This is OK with me - my primary concern was about the "DESKTOP" enum-constant where "LOCAL" is a better name.

desktop/ui/src/main/java/org/datacleaner/cli/CliRunner.java
```diff
@@ -70,7 +73,6 @@
  * Contains the execution logic to run a job from the command line.
  */
 public final class CliRunner implements Closeable {
-
```
kaspersorensen: OK, it really doesn't matter but I am starting to notice that you generally don't like this extra line after the class declaration, but I kind of like it. But it's just a damn visual preference, like space between methods or whatever. So we should just maybe agree on what is best.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobLifeCycleListener.java
kaspersorensen: It's coming in RowProcessingFunction, but I took a look and it's not trivial because that function has two was it can be invoked (it also implements two interfaces accordingly) and the partition number is only available in one of them. So let's not care about it for now.

engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopUtils.java
kaspersorensen: At the moment, the only thing we/I managed to get working on MapR is a completely manual ```spark-submit``` invocation from within the console on one of the nodes itself. I hope we find a way to also do something as a more external client, but if we don't then at least we should keep it in mind that for MapR users we have some limitations and certain convenience utilities that aren't working.

kaspersorensen:
LGTM, please merge unless you feel that what I have mentioned here in the remarks are cause for concern.

desktop/ui/src/main/java/org/datacleaner/cli/CliRunner.java
LosD: Hehe, I don't care that much, it just looks weird to me, like an empty line after other scope operators (which I guess isn't really comparable, but eyes still reacts the same). :)

PR #1039, opened 2016-01-08
Bug/963 remote servers problem notifications

### LINKED ISSUE #963: Use DC desktop notifications areas to inform users about remote server's problems
For now, when the remote component is not available, the only way how a user can get this information is within an error overlay in the component's icon. Other notification areas could be used to provide more specific hints and/or problem reports.

khouzvicka:
Fix of notification center

Fixes #963


api/src/main/java/org/datacleaner/configuration/DescriptorProviderStateListener.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.datacleaner.descriptors.DescriptorProvider;
+import org.datacleaner.descriptors.DescriptorProviderStatus;
+
+import java.util.Map;
+
+/**
+ * Listener for notification about providers. {@link DescriptorProviderStateNotifier}
+ */
+public interface DescriptorProviderStateListener {
```
kaspersorensen: What is this new interface doing? I don't see it used anywhere (BUT it is being imported below, creating an unused import!).

api/src/main/java/org/datacleaner/descriptors/DescriptorProvider.java
```diff
@@ -185,6 +180,11 @@
             Class<? extends RenderingFormat<?>> renderingFormat);
 
     /**
+     * Notify all listeners
+     */
+    public void notifyListeners();
```
kaspersorensen: Hmm... Something which has listeners should never have a method to notify them. The whole point of having a listener is that the listener is automatically invoked when it is relevant. It should never be the responsibility of someone coming from the outside and telling that listeners should be notified.

LosD:
It doesn't really make sense to me that a new approach is still using the old branch? The commit history will look pretty horrible after merging it.

I'd suggest you made a new branch and merged this one into it with ```git merge --squash bug/963-remote-servers-problem-notifications```.

api/src/main/java/org/datacleaner/descriptors/DescriptorProvider.java
```diff
@@ -198,4 +198,17 @@
...
+     *
+     * @return
+     */
+    public Map<DescriptorProvider, DescriptorProviderStatus> getActualStatusMap();
+
+    /**
+     * Check status od descriptor provider.
+     * Don't use that. If you want to be notified, please add the listener to this object.
+     */
+    public void checkStatus();
```
kaspersorensen: Why add a method where it says "don't use that"? I'm confused!

LosD:
Hmmm... There also seems to be a lot of leftovers from the old code. That's unfortunate.

api/src/main/java/org/datacleaner/descriptors/DescriptorProvider.java
```diff
@@ -198,4 +198,17 @@
      * @param listener
      */
     public void removeListener(DescriptorProviderListener listener);
+
+    /**
+     * Get information about providers.
+     *
+     * @return
+     */
+    public Map<DescriptorProvider, DescriptorProviderStatus> getActualStatusMap();
```
kaspersorensen: Hmm, to be honest I'm quite concerned about the changes here.

Most importantly on the return type: The fact that it returns a Map is really leaking the fact that this was made for a composite implementation.

Secondly: Why is this even needed? I thought this was exactly what the listeners are for - so the information can be given to the listeners instead of them having to pull it out themselves.

Thirdly: The naming - "Actual" - this seems to indicate to me that there is something "unactual" in play also, or?

desktop/ui/src/main/java/org/datacleaner/bootstrap/DCWindowContext.java
```diff
@@ -151,6 +153,10 @@ public void exit() {
 		}
 		if (_configuration != null) {
 			_configuration.getEnvironment().getTaskRunner().shutdown();
+			DescriptorProvider descriptorProvider = _configuration.getEnvironment().getDescriptorProvider();
+			if(descriptorProvider instanceof CompositeDescriptorProvider){
+				((CompositeDescriptorProvider) descriptorProvider).shutdown();
```
kaspersorensen: Probably we could make this unnecesary by using daemon threads instead of normal threads. I think that would be better for other environments too since we use descriptor providers also in DC monitor, DC for Apache Spark and so on. In other words: It shouldn't be something that we have to handle specifically just in the DC desktop application.

desktop/ui/src/main/java/org/datacleaner/guice/DCModuleImpl.java
```diff
@@ -32,14 +32,7 @@
...
 import org.datacleaner.bootstrap.WindowContext;
-import org.datacleaner.configuration.DataCleanerConfiguration;
-import org.datacleaner.configuration.DataCleanerConfigurationImpl;
-import org.datacleaner.configuration.DataCleanerEnvironment;
-import org.datacleaner.configuration.DataCleanerEnvironmentImpl;
-import org.datacleaner.configuration.DataCleanerHomeFolder;
-import org.datacleaner.configuration.DatastoreXmlExternalizer;
-import org.datacleaner.configuration.InjectionManager;
-import org.datacleaner.configuration.InjectionManagerFactory;
+import org.datacleaner.configuration.*;
```
kaspersorensen: Please import IntelliJ IDEA formatter from the root directory. This change does not comply :)

desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
```diff
@@ -645,9 +645,9 @@ public String getToolTipText(MouseEvent event) {
     public Datastore getDatastore() {
         return _datastore;
     }
-    
+
     @Override
-    public void onDescriptorsUpdated(DescriptorProvider descriptorProvider) {
+    public synchronized void onDescriptorsUpdated(DescriptorProvider descriptorProvider) {
```
kaspersorensen: I doubt this is related to the issue you're trying to solve actually.

But _if_ you want to make a fix also for this here, then ```synchronized``` is not the right way. The potential issue you're facing is that a non-AWT thread is invoking this method (since it's a listener method - invoked by the descriptor provider monitoring thread). And all changes to AWT (and thereby Swing) must happen in the AWT dispatcher thread. So instead use WidgetUtils.invokeSwingAction(...).

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraph.java
```diff
@@ -79,6 +71,15 @@
...
 import org.slf4j.LoggerFactory;
 
+import edu.uci.ics.jung.algorithms.layout.StaticLayout;
+import edu.uci.ics.jung.graph.DirectedGraph;
+import edu.uci.ics.jung.visualization.GraphZoomScrollPane;
+import edu.uci.ics.jung.visualization.RenderContext;
+import edu.uci.ics.jung.visualization.VisualizationServer.Paintable;
+import edu.uci.ics.jung.visualization.VisualizationViewer;
+import edu.uci.ics.jung.visualization.VisualizationViewer.GraphMouse;
+import edu.uci.ics.jung.visualization.control.PluggableGraphMouse;
```
kaspersorensen: Sorry for nitpicking, but there's no point in this change (moving imports up and down) and it is just polluting our change log.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphTransformers.java
```diff
@@ -235,7 +240,20 @@ public Icon transform(Object obj) {
                 } else {
                     configured = componentBuilder.isConfigured(false);
                 }
-                return IconUtils.getDescriptorIcon(descriptor, configured, IconUtils.ICON_SIZE_LARGE);
+
+                Icon descriptorIcon = IconUtils.getDescriptorIcon(descriptor, configured, IconUtils.ICON_SIZE_LARGE);
+
+                if (descriptor instanceof RemoteTransformerDescriptorImpl) {
+                    Map<DescriptorProvider, DescriptorProviderStatus> actualDescriptorProviderStatusMap  = componentBuilder.getAnalysisJobBuilder().getConfiguration().getEnvironment().getDescriptorProvider().getActualStatusMap();
```
kaspersorensen: OK so this answers what the method is used for - but unfortunately also confirms why it's not so good.

This method (```Icon transform(Object)```) is invoked for _every_ change in the graph and for _every_ component on it - even just a move of a node to a new position or whatever. So having to traverse through descriptors, maps and states is maybe a small operation, but you're going to do it again and again and again. This can make our UX become quite slow and annoying.

Which is why a listener approach is much better. When a change is detected (by some background process) - update something. In this case I would suggest to have a listener that is aware of which descriptor providers are in error state. That way you will essentially have something like ```Set<ComponentDescriptor> errorStateDescriptors``` which you can do a single check in: ```if (errorStateDescriptors.contains(descriptor) { addErrorOverlay(icon); }```.

desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphTransformers.java
```diff
@@ -235,7 +240,20 @@ public Icon transform(Object obj) {
                 } else {
                     configured = componentBuilder.isConfigured(false);
                 }
-                return IconUtils.getDescriptorIcon(descriptor, configured, IconUtils.ICON_SIZE_LARGE);
+
+                Icon descriptorIcon = IconUtils.getDescriptorIcon(descriptor, configured, IconUtils.ICON_SIZE_LARGE);
+
+                if (descriptor instanceof RemoteTransformerDescriptorImpl) {
```
kaspersorensen: Since we are talking about an API and an interface here, I wouldn't want to restrict this to only remote transformers. The descriptor state stuff is capable of working on _all_ components, so let's code against that API ourselves too.

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -490,19 +524,54 @@ protected void onWindowVisible() {
     }
 
     public void updateStatusLabel() {
+        DescriptorProviderStatus remoteErrorState = null;
+        boolean remoteExecuteEnable = true;
+
+        for (ComponentBuilder componentBuilder : _analysisJobBuilder.getComponentBuilders()) {
+            ComponentDescriptor<?> descriptor = componentBuilder.getDescriptor();
+            if (descriptor instanceof RemoteTransformerDescriptorImpl) {
```
kaspersorensen: Same remark as before regarding global applicability of the API / interface change.

engine/core/src/main/java/org/datacleaner/descriptors/CompositeDescriptorProvider.java
```diff
@@ -22,23 +22,32 @@
...
 
 /**
  * DescriptorProvider that provides a composite view of descriptors from a list
  * of delegate providers.
  */
 public class CompositeDescriptorProvider implements DescriptorProvider {
+    private static final Logger logger = LoggerFactory.getLogger(CompositeDescriptorProvider.class);
+    private static final long CLOSE_TIMEOUT = 5000;
+    private ServerChecker serverChecker;
+    private Thread serverCheckerThread;
```
kaspersorensen: These two should be final IMO

engine/core/src/main/java/org/datacleaner/descriptors/CompositeDescriptorProvider.java
```diff
@@ -48,6 +57,9 @@ public CompositeDescriptorProvider(DescriptorProvider delegate1, DescriptorProvi
 
     public CompositeDescriptorProvider(List<DescriptorProvider> delegates) {
         this.delegates = delegates;
+        serverChecker = new ServerChecker();
+        serverCheckerThread = new Thread(serverChecker);
```
kaspersorensen: Let's make the thread a daemon thread, and let's also give it a name so that it is easy to identify when investigating thread dumps.

engine/env/spark/src/test/java/org/datacleaner/spark/JobResultPathTest.java
```diff
@@ -32,6 +32,7 @@
 import org.apache.spark.api.java.JavaSparkContext;
 import org.datacleaner.job.AnalysisJob;
 import org.datacleaner.spark.utils.ResultFilePathUtils;
+import org.junit.Ignore;
```
kaspersorensen: A new unused import

kaspersorensen:
Sorry, needless to say this PR isn't really how I was hoping it :) I honestly think you need to start fresh and really be careful about making the right changes only. We are dealing with an API and a public interface here, so any changes must be thought about properly. This PR to me seems more to be a case of patching together a solution that might work, but won't make us happy in the long run - and we don't like to inherit such a solution.

api/src/main/java/org/datacleaner/configuration/DescriptorProviderStateListener.java
khouzvicka: Yes, of course.
I missed it.

desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
khouzvicka: Yes, I know. This problem will be fixed other way. Special branch is for this. I will remoce this change.

kaspersorensen:
Is work still being done on this branch? I am unsure if we should review it again or if something else is going on, so please keep us updated.

kaspersorensen:
Closing due to inactivity.

PR #1043, opened 2016-01-11
Issue #1030: Fixes saving of only 1 component when having only 1 builder

### LINKED ISSUE #1030: Multiple input columns in "Value distribution" component
#### Description
A workflow with some input, transformer (e. g. Concatenator) and a value distribution. When multiple input columns for value distribution are selected, no change in GUI is performed (no more value distribution components, no more links). After loading (previously saved) job there is a value distribution component for each column (with proper link). 

##### Before save
beforesave

##### After load
afterload

#### Expected behavior
1) Multiple value distribution components are created right after the selection of more input columns. 
[Or. 2) There is only one distribution component with multiple links in the workflow. ]

kaspersorensen:
Fixes #1030

Finally got around to fix this bug ... The individual components was created because the builder creates many analyzer instances. My approach to solving the bug was to then also instrument each produced analyzer with a "secret" (at least never published) metadata value which the JaxbJobWriter then can use to correlate instances of the same builder and write them under the same ```<analyzer>``` element.

If it passes review, I suggest to also apply this fix to the coming 4.5.4 release.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/builder/AnalyzerComponentBuilder.java
+++ b/engine/core/src/main/java/org/datacleaner/job/builder/AnalyzerComponentBuilder.java
@@ -55,8 +55,10 @@ import org.slf4j.LoggerFactory;
  * @param <A>
  *            the type of {@link Analyzer} being built.
  */
-public final class AnalyzerComponentBuilder<A extends Analyzer<?>>
-        extends AbstractComponentBuilder<AnalyzerDescriptor<A>, A, AnalyzerComponentBuilder<A>> {
+public final class AnalyzerComponentBuilder<A extends Analyzer<?>> extends
+        AbstractComponentBuilder<AnalyzerDescriptor<A>, A, AnalyzerComponentBuilder<A>> {
+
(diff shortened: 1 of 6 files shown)


LosD:
LGTM

arjansh:
I still have a question regarding the fix. In the `AnalyzerComponentBuilder#createPartitionedJob(Object, Map)` method you retrieve the map of `metadataProperties`, then put the `METADATA_PROPERTY_BUILDER_ID` property on it and pass the map to the new `ImmutableAnalyzerJob`. Why not invoke the `setMetadataProperty(String, String)` method from the `toAnalyzerJobs(boolean, AnalysisJobImmutabilizer)` method (outside the for loop where you iterate over the `entrySet` of `originatingTables`) instead to set the `METADATA_PROPERTY_BUILDER_ID` property?

kaspersorensen:
Valid remark, and maybe that would have been a bit clearer/easier actually.

I did it because I wanted to make the minimum footprint I could. So for the moment I thought "better only apply this property on the new copy of the metadata". This would save it from ever being introduced in cases where we don't have the "special case" that one builder produces multiple components. But actually it might have just made things more difficult.

PR #1044, opened 2016-01-12
Fixed list-index bug in Invoke child job transformer + integration test

kaspersorensen:
While I was building an example of using the ```Invoke child Analysis job``` transformation I discovered that it was actually having a pretty sincere bug. So this fixes it and also adds an integration test to improve the test coverage of this piece.


engine/core/src/main/java/org/datacleaner/components/composition/AbstractWrappedAnalysisJobTransformer.java
```diff
@@ -133,7 +133,7 @@ public OutputColumns getOutputColumns() {
         final String[] names = new String[size];
         final Class<?>[] types = new Class[size];
         for (int i = 0; i < size; i++) {
-            InputColumn<?> outputColumn = _outputColumns.get(0);
+            final InputColumn<?> outputColumn = _outputColumns.get(i);
```
kaspersorensen: For reference - this was the "sincere" bug - it wasn't applying the right index.

LosD:
LGTM!

PR #1046, opened 2016-01-12
Make CoalesceUnit immutable

### LINKED ISSUE #1045: CoalesceUnit can be updated by multiple transformers, leading to corruption on preview
When previewing, a copy is taken of the job. Unfortunately, the CoalesceUnits used by the fuse and  union components are mutable, and will be refreshed by the copied job, corrupting their configuration. This can both make the preview fail, and remove wiring.

Steps to reproduce:
1. Create a job that uses fuse/coalesce to merge columns in multiple units. Send results to e.q. a value distribution.
2. Run job. It should execute fine.
3. Open configuration of the fuse component and run a preview.
4. Run the job again.

Expected result:
No errors, all configured columns are fused.

Actual result:
Error in step 3. Maybe wiring disappears, making step 4 only give a partial result (i.e. a column missing).

LosD:
This avoid problems with copied job and transformers affecting each other.

Fixes #1045

### DIFF
--- a/components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceMultipleFieldsTransformer.java
+++ b/components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceMultipleFieldsTransformer.java
@@ -21,6 +21,7 @@ package org.datacleaner.components.fuse;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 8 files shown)


kaspersorensen:
Damn. I couldn't find anything to nag about. Merge when Travis passes ;-)

PR #1049, opened 2016-01-13
Analyzer no errordatastore

MennoB:
Basically added a nullcheck for the error datastore. This results is simply showing the number of failed rows to the user if no error datastore is configured.

I created this branch from the 4.5.4 release branch. If it should be based on the master instead, please let me know.


kaspersorensen:
I've restarted the Travis build for this PR since the error it gave was unrelated (very sporadic incident - but I'll adjust the threshold in the junit assertion that caused it).

kaspersorensen:
Otherwise looks good to me.
We will simply apply it to both master and the 4.5.4 branch IMO. I'll take care of it when Travis agrees.

MennoB:
Great! Thanks.

PR #1052, opened 2016-01-13
Fixes #1040

### LINKED ISSUE #1040: Scrollbar thumbs can get quite hard to notice in results
Just had a job with a huge amount of values in value distribution, and I thought we had a bug because I didn't see the scrollbar thumb, as it is tiny with that many rows. We should probably make it a liiiittle larger.

image

ClaudiaPHI:
Assigned a minimum height to the scrollbar in the analyser result panel. 

Fixes #1040 

image


kaspersorensen:
LGTM!

PR #1062, opened 2016-01-16
Rewrote RemoveUnwantedCharsTransformer from scala to Java

kaspersorensen:
Just a small rewrite from Scala to Java. This will help ensure that the "basic transformers" module is (except for a bit of test code) fully functional with just a java compiler / IDE


LosD:
LGTM!

PR #1065, opened 2016-01-18
Bug/1042 cannot preview transformation

### LINKED ISSUE #1042: Can't preview transformation when another/unrelated component is not configured
I noticed today that if I try to preview a transformation, then even a "dangling" (completely unconfigured - only placed on the canvas) component can make my preview fail.

image

I suggest to do two fixes:

 * First improve the error message since in this case we could tell which component's configuration is causing the issue.
 * Secondly make the preview function ignore completely irrelevant components to the extent feasible.

ClaudiaPHI:
Fixes #1042. 
- I improved the error message handling. 
- The components that were not configured (as in do not have input columns) are removed from the analysisJobBuilder.


desktop/api/src/main/java/org/datacleaner/windows/DataSetWindow.java
```diff
@@ -174,7 +174,7 @@ protected void done() {
                     logger.error("Unexpected error occurred while building DataSetWindow contents", e);
                     DataSetWindow.this.dispose();
                     WidgetUtils.showErrorMessage("Unexpected error",
-                            "An unexpected error occurred while building data set. See logs for details.");
+                            "An unexpected error occurred while building data set. See logs for details." + e.getMessage());
```
kaspersorensen: This error message is deliberately not showing the exception message because it can be _anything_. So reintroducing it is not good - it should be hidden from the user since most users don't want a very techical message thrown in their head anyway.

Besides that - it's also missing a space or else it would append it directly onto the dot ;-)

desktop/api/src/main/java/org/datacleaner/actions/PreviewTransformedDataActionListener.java
```diff
@@ -290,6 +290,14 @@ private void sanitizeIrrelevantComponents(AnalysisJobBuilder ajb, TransformerCom
                     // components
                     relevantAnalysisJobBuilder.removeComponent(componentBuilder);
                 }
+
+                if (!importantComponent) {
+                    // remove the components that are not configured.
+                    if (componentBuilder.getInput().length == 0) {
```
kaspersorensen: We could do this more generically by asking if componentBuilder is configured. Then it would not be restricted to just input columns, but to any missing configuration aspect.

ClaudiaPHI: Ok.

kaspersorensen:
LGTM!

kaspersorensen:
I understand the irony and all, but for the future please avoid the commit messages like this:
image
It just doesn't look professional in our commit log. I'll squash it locally and then merge it in to avoid it.

PR #1067, opened 2016-01-18
Issue #954: Fixed

### LINKED ISSUE #954: Support record counts greater than Integer.MAX_VALUE
Applying DC on big data is so far showing to be feasible, but if we go to extreme amounts of data we may run into int capacity issues. We use __int__ in various places such as InputRow.getId(). Currently we don't even know if this is a problem or not, but probably we will see some strange issues caused by this.

For that reason we should review our use of __int__ for all metrics and values that represent e.g. "record number" or so. To the extent possible, replace it with __long__.

kaspersorensen:
Fixes #954

I should add a few notes to this:

 * I first tried something quite ambitious - implementing ```long``` instead of ```int``` more or less everywhere. The important places I identified this was:
  * InputRow.getId() (which is what #954 tells about)
  * RowAnnotation.getRowCount() (not described in #954)
  * RowProcessingMetrics.getExpectedRow() (not described in #954)
 * But the above quickly became a nightmare to solve in one go. Especially the RowAnnotation change will have a huge impact. It literally is used in hundreds of places. And it's used a lot in result reducer code where it becomes hard to fix unless you really consider line-by-line or even class-by-class what makes sense to do.
 * So I reverted to "just" fixing #954. And here's the result.
 * Once it was fixed I saw a lot of breaking tests because of deserializing legacy InputRow objects. So I introduced a new class to consolidate the somewhat involved tricks to convert an old ```int``` variable into a new ```long``` value. This wasn't pretty IMO, but it's the only way (that I know of) to do it while keeping deserialization compatibility.

### DIFF
--- a/api/src/main/java/org/datacleaner/api/InputRow.java
+++ b/api/src/main/java/org/datacleaner/api/InputRow.java
@@ -49,7 +49,7 @@ public interface InputRow extends Serializable {
      * 
      * @return an identifier for this row
      */
-    public int getId();
+    public long getId();
 
     /**
      * @return the input columns represented in this row
--- a/components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
+++ b/components/basic-transformers/src/main/java/org/datacleaner/beans/numbers/GenerateIdTransformer.java
@@ -19,7 +19,7 @@
  */
 package org.datacleaner.beans.numbers;
 
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import javax.inject.Named;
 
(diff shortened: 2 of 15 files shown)


LosD:
This is one of the few places I really miss C and it's typedef'ing; typedef an id type to int, and simply change that typedef when times change.

LosD:
Wow, yeah that is quite a turd, but that's how it is, I guess. But if it goes into 5.x, is it necessary to keep deserialization compatibility?

Otherwise LGTM.

LosD:
By the way, expanding an API field from int to long, will make the API source-incompatible (narrowing without explicit cast), but I guess it should keep the bytecode compatibility. Is that okay?

kaspersorensen:
For a moment I was also really thinking "OK let's use the 5.0 gap to ditch deserialization compatibility". But it's really one of the hardest things to do for us. It would involve asking all customers to throw away old results (in DC monitor that means "kill your history") and also for ourselves - all example results must be re-made. That sounds quite more nasty to me.

LosD:
I don't care so much about ourselves... That's a one-time thing, and a few releases down the road would have paid itself off many times.

But the DC Monitor history nuking... That's bad.

kaspersorensen:
Regarding source/binary compatibility: I don't think it's a problem. A recompile is fine when we're talking about a 5.0 release. That's only a (quite minor) annoyance for developers, not for end users.

PR #1080, opened 2016-01-19
Fixes #1074: Create crucial repository folders automatically.

### LINKED ISSUE #1074: Autocreate jobs, timeslines and results folder in monitor
The DC monitor fails in various ways if any of the folders ```jobs```, ```timelines``` and ```results``` doesn't exit. While not a bug per se, let's autocreate them instead.

kaspersorensen:
Fixes #1074 by using ```getOrCreateFolder(...)``` for crucial folders instead of ```getFolder(...)```

### DIFF
--- a/monitor/services/src/test/java/org/datacleaner/monitor/configuration/ComponentStoreImplTest.java
+++ b/monitor/services/src/test/java/org/datacleaner/monitor/configuration/ComponentStoreImplTest.java
@@ -47,7 +47,7 @@ public class ComponentStoreImplTest {
 
     private RepositoryFolder getRepositoryFolderMock() {
         RepositoryFolder repositoryFolder = createNiceMock(RepositoryFolder.class);
-        expect(repositoryFolder.getFolder(ComponentStoreImpl.FOLDER_NAME)).andReturn(getComponentsFolderMock()).anyTimes();
+        expect(repositoryFolder.getOrCreateFolder(ComponentStoreImpl.FOLDER_NAME)).andReturn(getComponentsFolderMock()).anyTimes();
         replay(repositoryFolder);
 
         return repositoryFolder;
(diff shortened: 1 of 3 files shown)


ClaudiaPHI:
LGTM!

monitor/services/src/test/java/org/datacleaner/monitor/configuration/ComponentStoreImplTest.java
```diff
@@ -47,7 +47,7 @@ private Repository getRepositoryMock() {
 
     private RepositoryFolder getRepositoryFolderMock() {
         RepositoryFolder repositoryFolder = createNiceMock(RepositoryFolder.class);
-        expect(repositoryFolder.getFolder(ComponentStoreImpl.FOLDER_NAME)).andReturn(getComponentsFolderMock()).anyTimes();
+        expect(repositoryFolder.getOrCreateFolder(ComponentStoreImpl.FOLDER_NAME)).andReturn(getComponentsFolderMock()).anyTimes();
```
LosD: No biggie, but isn't it supposed to be called only once? Why the anyTimes?

kaspersorensen: That's not a change of this PR, but my 5c on the topic would be: Probably it doesn't matter but in the eyes of the unittest it also doesn't matter, it's just a specification of behaviour for the mock rather than a requirement that it happens 0, 1 or 1000 times.

LosD: Hmmmm, I'm not sure I agree here: If it calls it more than once, something weird is going on. Catching when code does weird stuff (even if it seems harmless at first) is a great may to discover bugs. Using anyTimes() right away is actively hiding bugs. Sure, if it is discovered that it does it more than once for a good reason, by all means add anyTimes() (or preferably times(x) if you have a good idea what x is, and x wont be easily affected by irrelevant code changes), but it's a terrible default.

(And expected execution count is certainly not just specifying the mock's behaviour, or it would only have to specified when you have differing outputs for every call. That's also why most mocking frameworks has a verification step after execution, to make sure methods has been called the minimum amount of times).

But yeah, it's old code, so not really related. It just seemed curious to me.

kaspersorensen: I think that really depends on the test case. Sometimes you want to ensure exact amount of invocations, and sometimes it may be too far away from the control of the thing you're testing. In the latter case you shouldn't require a specific number of invocations because the role of the mock is just to "behave like an implementation would".

PR #1082, opened 2016-01-19
Feature/1075 root url security

### LINKED ISSUE #1070: Implement "pretty URLs" in DC monitor
In a number of projects we've now made use of OCP rewrite - let's also do it in DC monitor so that we don't have to use ".jsf" extension in the URL for the pages.

### LINKED ISSUE #1075: Make root of monitor context unsecured
Because the root is secured, there are some pretty weird redirect issues when opening the monitor. It may help to set its security to none in security context.

kaspersorensen:
Fixes #1075
Fixes #1070

I found that these two issues was so related that I decided to fix them in the same branch.


kaspersorensen:
Updated and tattletale test now passes :)

kaspersorensen:
Please review this. After a couple of commits I'm finally happy with it now :-)
Have tested it on many configurations - Jetty, Tomcat, on the "/" context path and on the normal "/DataCleaner-monitor" context path.

ClaudiaPHI:
:+1:

arjansh:
LGTM!

PR #1066, opened 2016-01-18
Reinstate coalesce update in union refresh

LosD:
Revert commit 1eeed14a8c1f19b8d2ffe43fe1da093f828ddd98 which had unintended sideeffects, and seems to be a wrong fix to a wider problem (see #1056).

Fixes #1064

PR #1084, opened 2016-01-20
Make CoalesceUnit path-based for physical columns

### LINKED ISSUE #1056: Union produces null values when input columns have name clashes
With a Union configuration like this (using orderdb):

image

Then the country and state values (which clash in column name) are only materialized for one of the two unioned tables:

image

I suppose this somehow has to do with the identification of columns by name.

LosD:
This is the first attempt at making CoalesceUnit use paths instead of names. This avoids clashes between two tables with the same name.

Potential issues:
- Transformed columns may still very much clash
- Dots in names will probably cause problems
- Test test test... And then test some more. We have had many unexpected problems, and while this is a pretty minor change code-wise, the change could have big implications.

 Fixes #1056

### DIFF
--- a/components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceUnit.java
+++ b/components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceUnit.java
@@ -62,7 +62,12 @@ public class CoalesceUnit {
     private String[] getInputColumnNames(InputColumn<?>[] inputColumns) {
         final String[] result = new String[inputColumns.length];
         for (int i = 0; i < inputColumns.length; i++) {
-            result[i] = inputColumns[i].getName();
+            final InputColumn<?> inputColumn = inputColumns[i];
+            if(inputColumn.isPhysicalColumn()){
+                result[i] = inputColumn.getPhysicalColumn().getQualifiedLabel();
+            } else {
+                result[i] = inputColumn.getName();
+            }
         }
         return result;
     }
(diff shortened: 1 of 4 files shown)


kaspersorensen:
Happy to read that you did a lot of test :-) But I guess then the million dollar quesiton is: Did you do those tests while also having applied #1066? Or does this replace that one? I'm unsure if there is an impact of combining the two?

components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceUnit.java
```diff
@@ -62,7 +62,12 @@ public CoalesceUnit(InputColumn<?>... inputColumns) {
     private String[] getInputColumnNames(InputColumn<?>[] inputColumns) {
         final String[] result = new String[inputColumns.length];
         for (int i = 0; i < inputColumns.length; i++) {
-            result[i] = inputColumns[i].getName();
+            final InputColumn<?> inputColumn = inputColumns[i];
+            if(inputColumn.isPhysicalColumn()){
```
kaspersorensen: i spot 2 missing spaces ;-)

kaspersorensen:
I could only find very minor stuff :-)
LGTM!

LosD:
I did quite a bit of test, but my note was that we _must_ do a lot of testing before merging... Including checking that #1066 does not screw it up. It shouldn't though. This branch fixes the same issue that the commit #1066 reverts also tried to fix (an update makes columns "jump table"), but in a little more proper way, and without breaking job loading.

But again. Let's test the living daylight out of this before merging.

components/fuse/src/main/java/org/datacleaner/components/fuse/CoalesceUnit.java
LosD: AW, MAN! I think I'll have to make a plugin that automatically inserts them :)

kaspersorensen:
Tested, checked - looks good!

PR #1087, opened 2016-01-22
Improved exception message when SaxExecutionIdentifierReader fails

kaspersorensen:
I made a small improvement here to ensure that if parsing fails, then the exception includes the file name so that investigation is easier. We recently had this happen for a customer and figuring out what went wrong wasn't very easy ...


ClaudiaPHI:
LGTM!

PR #1094, opened 2016-01-22
Feature/1069 update storage provider defaults

### LINKED ISSUE #1069: Update defaults settings for InMemoryStorageProvider
Right now the default setttings storage provider is to use max. 500 sample sets with max 500 sample records each.

We would like to change the default to allow for more sets, potentially with fewer sample records, to allow at least previewing more of the "long tail" cases in a typical DQ analysis.

See ```InMemoryStorageProvider```

New defaults should be to have default max 1000 sample sets with max 150. records each.

kaspersorensen:
Fixes #1069

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/storage/InMemoryStorageProvider.java
+++ b/engine/core/src/main/java/org/datacleaner/storage/InMemoryStorageProvider.java
@@ -39,11 +39,11 @@ public final class InMemoryStorageProvider implements StorageProvider {
     private final int _maxSampleRecords;
 
     public InMemoryStorageProvider() {
-        this(500, 500);
+        this(1000, 150);
     }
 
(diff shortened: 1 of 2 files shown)


ClaudiaPHI:
LGTM!

LosD:
Not new, but a bit too magic-numbery for my tastes

kaspersorensen:
You mean the math stuff for the one-argument constructor? Or all of it?

LosD:
Mostly the repeating numbers in the two contructors, but all magic numbers should preferably be declared constants.

kaspersorensen:
Added constants now. Obviously that's not possible in the .xsd, but now they're at least available in the java code.

LosD:
LGTM

PR #1096, opened 2016-01-22
Issue #1095: Added reducer for CategorizationResult

### LINKED ISSUE #1095: Table lookup is not distributable (due to it's CategorizationResult not having a reducer)
Spark jobs that perform table lookup cannot run because CategorizationResult does not have a reducer.

So I guess this issue can also be phrased as: Create a reducer for CategorizationResult

kaspersorensen:
Fixes #1095


ClaudiaPHI:
:+1:

PR #1097, opened 2016-01-24
Issue #1079: Implemented general purpose CLOB rendering for DCTables

### LINKED ISSUE #1079: Previewing a SQL Server table does not convert clobs to strings
DataCleaner results and previews will retrieve the string contents of clob types using MetaModelInputRow.convertValue(), but previewing the table uses normal MetaModel queries, so datsa will not be converted, making it look something like this:
image

This is worsened by the fact that jTDS makes the SQL type VARCHAR(MAX) into Clobs.

kaspersorensen:
Fixes #1079 

The approach was not to go "narrowly" just after the preview source table window, but to take all data sets displayed. To do this I made the fix in the DCTableCellRenderer which is applied to all tables in DC desktop.

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/widgets/table/DCTableCellRenderer.java
+++ b/desktop/api/src/main/java/org/datacleaner/widgets/table/DCTableCellRenderer.java
@@ -47,95 +47,95 @@ import org.slf4j.LoggerFactory;
 
 public class DCTableCellRenderer implements TableCellRenderer {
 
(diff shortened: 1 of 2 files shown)


kaspersorensen:
The class was using tabs instead of spaces, so I committed also a reformat. But if you want to review it without having to look at that, then just look at the initial/first commit. The second commit does not do any semantic changes - only reformat.

LosD:
LGTM! :)

PR #1099, opened 2016-01-26
Logging errors in DC monitor

### LINKED ISSUE #1085: Errors happening in GWT service calls are logged to System-out instead of normal log file
I am a bit confused about the reason for this, but it seems that when an exception is thrown in a GWT service method then it is not logged in our normal DC monitor log files. Except you can typically find it in catalina.out (if using Tomcat) which is what Tomcat is using to catch stuff from System.out.

We should ensure somehow that it is logged properly in our log file!

ClaudiaPHI:
Fixes #1085 

 Due to some timing issue the private static Logger would not log the
errors to the log file(it is possible the logger is created too early). Therefore the logger is created when it is needed.


kaspersorensen:
I can confirm that this fixed the issue and that after trying several other things, this was the only thing we could find that fixed the issue. I wish I knew why the static logger isn't working properly, but this is.

PR #1102, opened 2016-01-26
Issue #1081: Implemented UI preferences for Webstart/JNLP and more.

### LINKED ISSUE #1081: Tone down the "Analyze" (open with JNLP) options in DC monitor
Since JNLP launching of jobs is largely being scrutinized by the increased security and constraints of Java nowadays, let's no longer make the launching of DC via JNLP as prominent. It's a feature that we no longer seem to use so much and have a hard time supporting. And now when conf.xml is easier to exchange between a (normal) desktop client and monitor server, it also becomes less important.

kaspersorensen:
Fixes #1081 by introducing a spring bean with UI preference flags.

My idea is that in DC EE we will use the normal ```datacleaner-monitor.preferences``` to set these flags.


LosD:
LGTM... But Travis hates it.

PR #1106, opened 2016-01-31
Issue #1104: Fixed

### LINKED ISSUE #1104: Remove null from pattern finder optional properties
Many of the pattern finder's optional properties has a <null> choice, which doesn't really make sense. Let's get rid of it.

image

kaspersorensen:
Fixes  #1104


LosD:
So much complexity removed. I love it!

PR #1111, opened 2016-02-01
Feature/1110 reference data in conf.xml

### LINKED ISSUE #1110: Update <reference-data> in conf.xml when editing these in desktop UI
Today the conf.xml file's reference data section is entirely manually edited. As a user you can add reference data via the DC desktop UI, but it's saved in user preferences instead of conf.xml. This is unfortunate because it means that the reference data is harder to share or port to the DC monitor server.

We should use a mechanism like the DatastoreXmlExternalizer which takes care of this for datastores. The same should be the case for dictionaries, synonym catalogs and string patterns.

kaspersorensen:
Fixes #1110 

This merges the old class ```DatastoreXmlExternalizer``` into something a bit more general - a new class, ```DomConfigurationWriter```, which is capable of externalizing also reference data elements to conf.xml.

In the DC desktop the MutableReferenceDataCatalog is now backed by this writer class, making it so that any change to reference data in the UI is reflected in conf.xml. Hurray!

Because the old class is used as API in various extensions and other projects, I decided to keep it, but let it extend the new class and be deprecated.

I discovered a few missing XSD attributes while implementing this - both pertaining to externalizing the "case sensitive" flag.

I also noticed that the desktop dialog for reference data is usually way too narrow, so I widened it.

The code is admittedly not very heavily tested. I did some monkey testing myself of the DC desktop, and all is working fine. But if anyone says "you need to test this more", then I guess I would say ok :)


kaspersorensen:
Added some basic tests so that coverage of DomConfigurationWriter is much better.

ClaudiaPHI:
LGTM!

PR #1112, opened 2016-02-02
Bug/411-regexp-search-replace-empty-replacement

### LINKED ISSUE #411: Regexp search/replace does not allow empty string replacement.
If a user plans to replace some value with an empty string, it is not allowed with the regex version. However, it is possible with a simple search/replace transformer. It should be possible with both transformers.

arjansh:
I added a PatternProperty annotation which can be used to indicate that a Pattern property is allowed to contain an empty string.

I added this annotation the the RegexSearchReplaceTransformer's replacePattern field and extended the SinglePatternPropertyWidget so it allows such a property to be empty if the annotation is used.

### DIFF
--- a/api/src/main/java/org/datacleaner/api/PatternProperty.java
+++ b/api/src/main/java/org/datacleaner/api/PatternProperty.java
@@ -1,0 +1,47 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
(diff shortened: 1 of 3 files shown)


kaspersorensen:
LGTM! Maybe you can reformat the complete property widget class because right now the mix of spaces and tabs looks very strange. The semantic code change is good!

kaspersorensen:
Merging, hurray!

PR #1113, opened 2016-02-02
Added "Remove all columns" button to Create table dialog

kaspersorensen:
On a customer engagement today we faced that it is sometimes convenient to be able to remove all columns from the "Create table" dialog. This PR does that:

image


ClaudiaPHI:
:+1:

PR #1114, opened 2016-02-02
Removing unused tenant resolver

kaspersorensen:
This PR simply removes a class that we never use. In DC EE there is a similar class, which is maintained and has some bugfixes associated with it.


ClaudiaPHI:
:+1:

PR #1115, opened 2016-02-02
Feature/1011 rest support to elasticsearch

### LINKED ISSUE #1011: Add REST connector support for ElasticSearchDatastore
Currently the ElasticSearchDatastore is capable of supporting NODE and TRANSPORT clients. But with MetaModel 4.5.0 we can now also support a REST based client. We should add configuration support for that.

ClaudiaPHI:
Fixes #1011  

The REST connector has been added to the datastore.I hope is readable enough. 

I have tested in DC community and it works . This will affect the  elastic search extension. 

This PR is for review. <b>Not meargeble</b> until someone says the approach is ok.


desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
```diff
@@ -60,8 +60,7 @@
 
     private static final long serialVersionUID = 1L;
 
-    private static final ElasticSearchDatastore.ClientType DEFAULT_CLIENT_TYPE =
-            ElasticSearchDatastore.ClientType.TRANSPORT;
+    private static final ElasticSearchDatastore.ClientType DEFAULT_CLIENT_TYPE = ElasticSearchDatastore.ClientType.TRANSPORT;
```
kaspersorensen: I would suggest to have REST as the default client type actually.

engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
```diff
@@ -108,46 +114,90 @@ public PerformanceCharacteristics getPerformanceCharacteristics() {
...
+
+    private JestClient getClientForRestProtocol() {
+        final JestClientFactory factory = new JestClientFactory();
+        factory.setHttpClientConfig(new HttpClientConfig.Builder("http://" + _hostname + ":" + _port).multiThreaded(true)
+                .build());
+        final JestClient client = factory.getObject();
+        return client;
+    }
+
+    private Client getClientForJoingClusterAsNode() {
```
kaspersorensen: Spelling mistake here: "Joing"

engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
```diff
@@ -108,46 +114,90 @@ public PerformanceCharacteristics getPerformanceCharacteristics() {
     }
 
     @Override
-    protected UsageAwareDatastoreConnection<ElasticSearchDataContext> createDatastoreConnection() {
+    protected UsageAwareDatastoreConnection<UpdateableDataContext> createDatastoreConnection() {
+
+        Object client = null;
```
kaspersorensen: Client being an Object is a bit dirty and means that you need to type-cast it further down in the method. Maybe you can rather do something common for NODE and TRANSPORT and then a completely different code block for REST. Then you can use the ES Client type and a JestClient type for the other scenario.

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
```diff
@@ -891,7 +892,7 @@ private Datastore createDatastore(String name, ElasticSearchDatastoreType datast
             }
         }
 
-        return new ElasticSearchDatastore(name, ClientType.TRANSPORT, hostname, port, clusterName, indexName, tableDefs,
+        return new ElasticSearchDatastore(name, ClientType.valueOf(clientType), hostname, port, clusterName, indexName, tableDefs,
```
kaspersorensen: Ouch, looks like TRANSPORT was hardcoded before?! That's a bug that I guess we also fix now then. Or am I missing something?

kaspersorensen:
Looks like Travis build is failing - some NPE in the unittests, better check why that happens. I think it's maybe because of the ```.valueOf(...)``` method which may be called with a null-argument if the client type is not specified. So it should be able to find a default in such a case then (which would then be TRANSPORT for the sake of backwards compatibility).

kaspersorensen:
Mostly LGTM, but check the remarks :)

ClaudiaPHI:
I am working on the test. I have assured the backwards compatibility

ClaudiaPHI:
I mean now I am fixing the backwards compatibility

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
ClaudiaPHI: Yes, it was hard coded.

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
```diff
@@ -857,6 +857,10 @@ private Datastore createDatastore(String name, ElasticSearchDatastoreType datast
 
         final String indexName = getStringVariable("indexName", datastoreType.getIndexName());
 
+        String clientType = getStringVariable("clientType", datastoreType.getClientType());
+        if (clientType == null) {
+            clientType = ClientType.TRANSPORT.name().toString();
```
kaspersorensen: ```name()``` already returns a string, so the last ```toString()``` is redundant :)

ClaudiaPHI:
Ready to merge, I think. If everyone agrees.

arjansh:
LGTM!

desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
```diff
@@ -224,7 +226,7 @@ public void onSelected(final FilenameTextField filenameTextField, final File fil
 
         if (originalDatastore == null) {
             _hostnameTextField.setText("localhost");
-            _portTextField.setText("9300");
+            _portTextField.setText("9200");
```
kaspersorensen: Not a huge problem, but now I see "9200" and/or "9300" a handful of times in this class. That calls for two constants to make it easier to maintain and avoid making only partial changes.

kaspersorensen:
Seems the branch has conflicts with master now, sorry. Can you fix and also check the remark above? Then it should be good I think.

PR #1117, opened 2016-02-02
Allow spaces in (some) reference data

### LINKED ISSUE #1116: Make SynonymLookupTransformer and RemoveDictionaryMatches work with multiple words
As the input string is tokenized by SynonymLookupTransformer and RemoveDictionaryMatches, they do not support spaces in the dictionary entries and synonyms.

Example synonym:
```
a test, the test
```
transform("This is the test"), expected output: "This is a test", actual output: "This is the test"

Example dictionary:
```
the test
```
transform("This is the test"), expected output: "This is", actual output: "This is the test".

While the words "Expected" and "Actual" are used here, this is not really a bug, but a limitation.

LosD:
This allows spaces in dictionaries and synonym lists for RemoveDictionaryMatches and SynonymLookup transformers.

Fixes #1116

### DIFF
--- a/components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptTransformer.java
+++ b/components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptTransformer.java
@@ -52,9 +52,7 @@ public class JavaScriptTransformer implements Transformer {
 	private static final Logger logger = LoggerFactory
 			.getLogger(JavaScriptTransformer.class);
 
-	public static enum ReturnType {
-		STRING, NUMBER, BOOLEAN;
-	}
+	public enum ReturnType {STRING, NUMBER, BOOLEAN}
 
 	@Configured
 	InputColumn<?>[] columns;
--- a/components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptTransformer.java
+++ b/components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptTransformer.java
@@ -52,7 +52,9 @@ public class JavaScriptTransformer implements Transformer {
 	private static final Logger logger = LoggerFactory
 			.getLogger(JavaScriptTransformer.class);
 
-	public enum ReturnType {STRING, NUMBER, BOOLEAN}
+	public static enum ReturnType {
+		STRING, NUMBER, BOOLEAN;
+	}
 
 	@Configured
 	InputColumn<?>[] columns;
--- a/components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
+++ b/components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
@@ -20,7 +20,10 @@
 package org.datacleaner.beans.transform;
 
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.inject.Inject;
 import javax.inject.Named;
@@ -55,14 +58,12 @@ import com.google.common.base.Strings;
 @ExternalDocumentation({ @DocumentationLink(title = "Segmenting customers on messy data", url = "https://www.youtube.com/watch?v=iy-j5s-uHz4", type = DocumentationType.VIDEO, version = "4.0") })
 @Categorized(superCategory = ImproveSuperCategory.class, value = ReferenceDataCategory.class)
 public class RemoveDictionaryMatchesTransformer implements Transformer {
-
-    public static enum RemovedMatchesType implements HasName {
-
+    public enum RemovedMatchesType implements HasName {
         STRING("String"), LIST("List");
 
         private final String _name;
 
-        private RemovedMatchesType(String name) {
+        RemovedMatchesType(String name) {
             _name = name;
         }
 
@@ -70,9 +71,7 @@ public class RemoveDictionaryMatchesTransformer implements Transformer {
         public String getName() {
             return _name;
         }
-    };
-
-    private static final Splitter SPLITTER = Splitter.on(CharMatcher.WHITESPACE).omitEmptyStrings();
+    }
 
     public static final String PROPERTY_DICTIONARY = "Dictionary";
     public static final String PROPERTY_COLUMN = "Column";
@@ -96,6 +95,7 @@ public class RemoveDictionaryMatchesTransformer implements Transformer {
     DataCleanerConfiguration _configuration;
 
     private DictionaryConnection dictionaryConnection;
+    private List<String> dictionary;
 
     public RemoveDictionaryMatchesTransformer() {
     }
@@ -135,6 +135,12 @@ public class RemoveDictionaryMatchesTransformer implements Transformer {
     @Initialize
     public void init() {
         dictionaryConnection = _dictionary.openConnection(_configuration);
+        dictionary = new ArrayList<>();
+
+        final Iterator<String> allValues = dictionaryConnection.getAllValues();
+        while (allValues.hasNext()) {
+            dictionary.add(allValues.next());
+        }
     }
 
     @Close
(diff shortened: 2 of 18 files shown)


components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptTransformer.java
```diff
@@ -52,9 +52,7 @@
 	private static final Logger logger = LoggerFactory
 			.getLogger(JavaScriptTransformer.class);
 
-	public static enum ReturnType {
-		STRING, NUMBER, BOOLEAN;
-	}
+	public enum ReturnType {STRING, NUMBER, BOOLEAN}
```
kaspersorensen: Not sure we have a convention on this type of thing - but if we have, is this it? I would normally use the old format of defining the inner enum. In general I would always have newlines after/before curly braces.

LosD: Might just be my old C ways, but I'd never use the "class" format for Enums that are just simple enumerations without behavior.

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -135,6 +135,12 @@ public OutputColumns getOutputColumns() {
     @Initialize
     public void init() {
         dictionaryConnection = _dictionary.openConnection(_configuration);
+        dictionary = new ArrayList<>();
```
kaspersorensen: Naming wise I think it would make sense to call this "dictionaryValues" or so.
And shouldn't it be sorted so that the longest one is the first and so on?

LosD: Agreed with the name.
It is sorted, though. Look inside the dictionary implementations.

components/reference-data/src/main/java/org/datacleaner/beans/transform/SynonymLookupTransformer.java
```diff
@@ -123,27 +131,9 @@ public void close() {
...
-                    if (replacedToken == null) {
-                        sb.append(token);
-                    } else {
-                        sb.append(replacedToken);
-                    }
-                }
-            }
-            return new String[] { sb.toString() };
-
+            SynonymCatalogConnection.Replacement replacement = synonymCatalogConnection.replaceInline(originalValue);
```
kaspersorensen: final :smiling_imp:

LosD: Dangit

components/reference-data/src/main/java/org/datacleaner/beans/transform/SynonymLookupTransformer.java
```diff
@@ -90,7 +90,15 @@ public SynonymLookupTransformer(InputColumn<String> column, SynonymCatalog synon
 
     @Override
     public OutputColumns getOutputColumns() {
-        return new OutputColumns(String.class, new String[] { column.getName() + " (synonyms replaced)" });
+        if (lookUpEveryToken) {
+            return new OutputColumns(
+                    new String[] { column.getName() + " (synonyms replaced)", column.getName() + " (synonyms found)",
+                            column.getName() + " (master terms found)" },
+                    new Class[] { String.class, List.class, List.class });
```
kaspersorensen: For most users it will probably be difficult to work with Lists. So I suggest we introduce a configured enum like ```RemoveDictionaryMatchesTransformer.RemovedMatchesType```.

LosD: True. Will add.

components/reference-data/src/test/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformerTest.java
```diff
@@ -74,31 +74,30 @@ public void testJobTitleScenarioRemovedMatchesAsString() throws Throwable {
...
+        assertEquals("    Designer  ", transformer.transform("  Lead   Designer  ")[0]);
         assertEquals("[Lead]", transformer.transform("  Lead   Designer  ")[1].toString());
         
-        assertEquals("Software Engineer", transformer.transform("Software  Engineer")[0]);
+        assertEquals("Software  Engineer", transformer.transform("Software  Engineer")[0]);
         assertEquals("[]", transformer.transform("Software  Engineer")[1].toString());
         
         assertEquals("Guru of employees", transformer.transform("Principal Senior Lead Guru of Junior employees")[0]);
-        assertEquals("[Principal, Senior, Lead, Junior]", transformer.transform("Principal Senior Lead Guru of Junior employees")[1].toString());
+        assertEquals("[Principal, Junior, Senior, Lead]", transformer.transform("Principal Senior Lead Guru of Junior employees")[1].toString());
```
kaspersorensen: Damn. I don't think this side-effect is really good. It used to be a good thing that you could rely on having the same order of dictionary terms in the output as in the input.

Maybe this calls for an alteration of the strategy - just thinking aloud here:
 * Maybe we should only do the "replace inline" stuff on multi-word dictionary terms.
 * And then we could retain the old logic of tokenizing the input first and looking up every individual token in the order of the input string.

kaspersorensen: Correction - I said "replace inline" and I meant "regex based search thingie-m-bob".

engine/core/src/main/java/org/datacleaner/reference/SimpleDictionary.java
```diff
@@ -134,10 +137,14 @@ public boolean equals(Object obj) {
 
     @Override
     public DictionaryConnection openConnection(DataCleanerConfiguration configuration) {
+        final SortedSet<String> connectionValueSet = new TreeSet<>(Comparator.comparingInt(String::length).reversed().thenComparing(String::compareTo));
```
kaspersorensen: Shouldn't this somehow be part of consuming code - or at least somehow be revealed in the interface, because right now it takes form of a "hidden implementation detail".

kaspersorensen:
Generally looks good - some small remarks and open questions remaining :)

components/reference-data/src/test/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformerTest.java
LosD: I'm not really a fan of keeping the old way around; There's other changes in behaviour (i.e. what a boundary character is), and it messes with the string by tokenizing and stitching it together again.

I'll think a bit about how to retain it... Including if we even want to (somehow "I replaced them in this order" is more important than "I received them in this order" to me, but I see the compatibility part).

components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptTransformer.java
kaspersorensen: Maybe it's wrong to change the convention right now at least :-) And we do have a habit of also letting our enums implement interfaces such as ```HasName``` or adding other methods to them.

LosD: Hmmm, first point probably true, but the last point only really matters when we actually DO implement behaviour, it shouldn't affect how a traditional ("stupid") enum is represented... Actually it could be a nice way to tell them apart.

kaspersorensen: Honestly I don't care - it goes beyond my own needs for sanitizing code styling, I'm only wondering if anyone else would notice it and frown upon the unstandardized way of doing it :-D

engine/core/src/main/java/org/datacleaner/reference/SimpleDictionary.java
LosD: Yeah, I'll make a `getLengthSortedValues`

components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptTransformer.java
arjansh: I don't frown upon it, but when I select the line of code and apply the auto formatter to it, it formats it back to the three lines variant.

LosD: Damn Eclipse! :)

LosD:
Okay, most review comments addressed. For now, the terms order will not be addressed

LosD:
Travis failure seems unrelated, but is a bit worrying: org.datacleaner.connection.UsageAwareDatastoreConnectionTest.testCloseNoRaceConditions(UsageAwareDatastoreConnectionTest.java:161)

kaspersorensen:
Merging, thanks for the fix.

PR #1119, opened 2016-02-03
Notify user about software update available

arjansh:
Moved the mechanism which pings the Monitor from the MonitorConnectionDialog to the MonitorConnection in a newly introduct ping method.

Added a checkForNewVersion method to the WelcomePanel, which invokes this ping method and compares the returned version with DataCleaner's version and displays a message on the Welcome screen if DataCleaner Monitor's version is highter then DataCleaner's version.


kaspersorensen:
I'm afraid that the objective has been misunderstood. This PR is comparing the desktop version to a monitor installations version. But the idea was different: To get the latest published version on the DC website and suggest the desktop user to download an upgrade from website too.
Currently the website ping is done in the UsageLogger class.

kaspersorensen:
@arjansh do you agree to closing this PR then, or do you want to keep it for something?

Issue #1120, opened 2016-02-03
Remote components - Better message when the credit is low


khouzvicka:
Remote components.
Improve the message in DataCleaner when the credit is low. We show the 403 exception during job processing now.

PR #1121, opened 2016-02-03
Better catching exception

khouzvicka:
#1120

### DIFF
--- a/engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
+++ b/engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
@@ -75,6 +75,8 @@ public class RemoteTransformer extends BatchRowCollectingTransformer {
     private ComponentRESTClient client;
     private Map<String, Object> configuredProperties = new TreeMap<>();
 
+    private RuntimeException lastException = null;
+
     public RemoteTransformer(RemoteServerData serverData, String componentDisplayName) {
         this.serverData = serverData;
         this.componentDisplayName = componentDisplayName;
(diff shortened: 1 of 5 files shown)


engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -75,6 +75,8 @@
     private ComponentRESTClient client;
     private Map<String, Object> configuredProperties = new TreeMap<>();
 
+    private RuntimeException lastException = null;
```
kaspersorensen: I would suggest to use something that is concurrent / thread-safe instead of an assigned exception. So here I would have
```private final AtomicBoolean failed = new AtomicBoolean(false);```

File: engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java [hunk 2]
```diff
@@ -298,10 +300,18 @@ public void map(BatchSource<InputRow> source, BatchSink<Collection<Object[]>> si
         logger.debug("Processing remotely {} rows", size);
 
         if (client == null) {
+            if (lastException != null) {
+                throw lastException;
+            }
```
kaspersorensen: ... and then here I would have
```
if (failed.get()) {
  throw new PreviousErrorsExistException();
}
```

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -298,10 +300,18 @@ public void map(BatchSource<InputRow> source, BatchSink<Collection<Object[]>> si
...
             throw new RuntimeException("Remote transformer's connection has already been closed. ");
         }
-
-        ProcessStatelessOutput out = client.processStateless(componentDisplayName, input);
+        ProcessStatelessOutput out;
+        try {
+            out = client.processStateless(componentDisplayName, input);
+        } catch (RuntimeException e) {
+            lastException = e;
```
kaspersorensen: ... and here I would then have:
```
boolean alreadyFailed = failed.getAndSet(true);
if (!alreadyFailed) {
  throw e;
} else {
  throw new PreviousErrorsExistException();
}
```

khouzvicka:
I implemented all your suggestion and also I did small change to better log.

kaspersorensen:
LGTM!

PR #1126, opened 2016-02-05
Fix for preview filter issue.

### LINKED ISSUE #1047: Preview does not always respect filters
Sometimes, filters are not taken into consideration when running a preview of a component, making the preview result weird/wrong.

LosD:
Issue was that max rows filter was added using OR compound filter, meaning that all other filters was essentially nuked (as well as the max rows filter if the other filters has an outcome).

I extended the filter to also support ALL scenario, though it is a bit of weird hybrid now (normally ALL scenario is done by chaining requirements).

For some reason, this does NOT fix the test case, allthough it fixed a manual test in DC (I used the first example Kasper provided in fixed case).

This fix is obviously not mergeable with a broken unit test, but I'd like some feedback to my approach.

## DO NOT MERGE IN CURRENT STATE!

Fixes #1047

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/CompoundComponentRequirement.java
+++ b/engine/core/src/main/java/org/datacleaner/job/CompoundComponentRequirement.java
@@ -25,7 +25,6 @@ import java.util.LinkedHashSet;
 import java.util.Objects;
 import java.util.Set;
 
-import org.apache.metamodel.util.HasName;
 import org.datacleaner.api.InputRow;
 
 /**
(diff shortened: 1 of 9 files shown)


kaspersorensen:
Travis does not ser the unittest failure?

engine/core/src/main/java/org/datacleaner/job/CompoundComponentRequirement.java
```diff
@@ -144,11 +185,8 @@ public boolean equals(Object obj) {
...
             return false;
         final CompoundComponentRequirement other = (CompoundComponentRequirement) obj;
-        if (_outcomes == null) {
-            if (other._outcomes != null)
-                return false;
-        } else if (!_outcomes.equals(other._outcomes))
-            return false;
-        return true;
+
+        // TODO: This wasn't anything equivalent to deepEquals before... But should it have been?
```
kaspersorensen: I don't think deepEquals is necessary on collections?

LosD: Ah, yeah. True in most cases (including this). AbstractCollection makes no such guarantee, and it's not in the Collection contract, but AbstractSet and AbstractList does.

LosD:
Apparently I added @Ignore to the test in some misguided moment.

LosD:
To be quite honest, I'm not really a big fan of my approach, but the alternative seems to be either add a component directly on all source columns, so we have a place to do the filtering, or make another CompoundComponentRequirement, which I would have no idea have to make a sensible naming from, without renaming this (which may be feasible, I don't think it would be used outside DC core?).

engine/core/src/main/java/org/datacleaner/job/CompoundComponentRequirement.java
```diff
@@ -95,12 +130,21 @@ public boolean hasMultipleRequirementsFrom(HasFilterOutcomes producingComponent)
 
     @Override
     public boolean isSatisfied(InputRow row, FilterOutcomes outcomes) {
-        for (FilterOutcome outcome : outcomes.getOutcomes()) {
-            if (_outcomes.contains(outcome)) {
-                return true;
+        if(_compoundingType == CompundingType.ANY || row == null) {
```
LosD: Note the `row == null`. Without this, the flow analysis went pretty crazy (ended up in the "This should never happen" Exception). This may have some serious drawbacks, but I'm not familiar enough with the flow analysis to be sure.

engine/core/src/main/java/org/datacleaner/job/CompoundComponentRequirement.java
```diff
@@ -30,42 +33,74 @@
  * requires just any of the provided outcomes.
  */
 public class CompoundComponentRequirement implements ComponentRequirement {
-
     private static final long serialVersionUID = 1L;
 
+    public enum CompundingType implements HasName {
```
arjansh: Shouldn't this be `CompoundingType` instead of `CompundingType`?

LosD: Errrr, yes :)

engine/core/src/main/java/org/datacleaner/job/CompoundComponentRequirement.java
```diff
@@ -95,12 +130,21 @@ public boolean hasMultipleRequirementsFrom(HasFilterOutcomes producingComponent)
 
     @Override
     public boolean isSatisfied(InputRow row, FilterOutcomes outcomes) {
-        for (FilterOutcome outcome : outcomes.getOutcomes()) {
-            if (_outcomes.contains(outcome)) {
-                return true;
+        if(_compoundingType == CompoundingType.ANY || row == null) {
```
kaspersorensen: Reviving your comment here (was removed due to a commit I guess)...

The ```row == null``` is in deed a little worrying. I see that when I remove it, I get exceptions in the ```PreviewTransformedDataActionListenerTest``` tests because DC's engine cannot calculate a proper processing order. Wondering what that means - give me 10 minutes to ponder on it ;-)

kaspersorensen:
:+1:

PR #1129, opened 2016-02-08
Fixes #1107 - resolving tenant ID from serialized User object.

### LINKED ISSUE #1107: Monitor multi-tenant logins does not always get the tenant
First logins may in some cases fail because the tenant does not get resolved. It seems that tenant is null (or empty string) after resolving it from the MapBasedTenantResolver.

kaspersorensen:
Fixes #1107 

After much investigation I finally found out what this was about - that the tenant ID was actually set as null by ```UserBean.updateUser()``` in cases where the UserBean had been deserialized by Tomcat! The way this happened was that TenantResolver was transient and in the ```updateUser``` call there was a "if tenant resolver is null, then tenant ID is null" kind of logic. Now instead we do not keep the tenant resolver in the UserBean, so it should never be null.


arjansh:
LGTM!

PR #1132, opened 2016-02-09
bugfix - remote components URL without port

jakubneubauer:
If remote components URL is without port, using default port to check the connectivity.


kaspersorensen:
:+1:

Issue #1063, opened 2016-01-16
Make the HTML rendered results look more like the desktop result screen


kaspersorensen:
The desktop Analysis result screen has been refined quite a lot during the last year or two. But the HTML rendering of the same hasn't. Unfortunately the HTML rendering doesn't look as pretty (IMO) and it's not a consistent experience to have so different layouts.

The most distinguishing difference is probably on the vertical vs horizontal orientation of results. The HTML rendering should adopt the "vertical tabs" approach similar to the desktop. In addition it should create one item/tab per component rather than per component type.

PR #1136, opened 2016-02-09
Feature/resource consolidation

kaspersorensen:
OK this PR is a bit odd ... I actually wanted to take a peek at the issue #1063 but then found that it was impeded by the fact that resources (images) where located in ```desktop-ui``` while the result HTML rendering was in the ```html-rendering``` module. So I spun off a small separate task (represented by this PR) to isolate the resources needed in a separate JAR so that it can be depended on. I did this for two items:

 * Common images
 * the ```orderdb.script``` file which defines the 'orderdb' datastore.

The latter created a LOT of reprecautions. Which is why this PR has so many changed files. A lot of unittests depended on old versions of orderdb. But rest assured, no non-test code has been changed - it's just a lot of test changes because the "real" orderdb has been modified with more customer records, more dirt and funny stuff.


ClaudiaPHI:
LGTM!

kaspersorensen:
Looking into the test failure ...

PR #1139, opened 2016-02-10
Wrong property shown for input columns mapped to enum

arjansh:
When you map input columns to an enum type input field using the `@MappedProperty` annotation, the wrong property was rendered, the input columns should be rendered, not the property to which they are mapped.


kaspersorensen:
Beautiful :-) Merge when travis agrees.

Issue #1143, opened 2016-02-11
Remote component enumeration value looks ugly after job reload


jakubneubauer:
To reproduce:
- Use a remote transformer with some enumeration property (it must have a human - readable form with HasName interface)
- Check, that in the combo box you can see the human-friendly value, not the java constant name.
- Save the job
- Load the job again

Now, instead of the human-friendly value, a java enumeration constant name is used in the combo-box.

PR #1144, opened 2016-02-11
Bug/1143 remote enum after reload

jakubneubauer:
For the remote transformers we need to support "dynamic enumerations" for component properties. It means for cases where some property of remote transformer has not a corresponding java enum class on the client side (DC Desktop).

This branch fixes problem of deserialization of such enumeration from String. We needed a Serializer that "knows" about the property descriptor. Without breaking a Serializer API, we changed the way of creation of the Serializers - now the property descriptor is responsible for Serializer creation. This makes our fix possible: The property descriptor of a "remote" enumeration can then create a Serializer that converts a String to enumeration value using the stored values from the property descriptor.


jakubneubauer:
fixes #1143

kaspersorensen:
I like the approach! Good patch.

PR #1148, opened 2016-02-12
transformer icon returned in a REST api is limited to 32x32

jakubneubauer:
If the developer of extension provides the icon bigger (by accident), it is scaled.


kaspersorensen:
Hmm the PR looks fine in itself, but I'm wondering if it is future proof. We might increase the desired size of icons in the future. But OTOH that would require a lot of work in itself so by then also this aspect would be up for review.

Maybe we can at least use the ```IconUtils.ICON_SIZE_LARGE``` constant instead of literally "32" in the code here?

jakubneubauer:
Used the constant.
I have one argument why limiting the size could be a good idea - it is safety check, to not send big data over the wire in a response to rest API GET /repository/{tenant}/components?iconData=true
(In case some author of some extension puts a big image to the extension .jar)

PR #1149, opened 2016-02-13
Issue #461: Implemented "Plain search/replace" with a Map<String,String>

### LINKED ISSUE #461: Search/replace transformers could take multiple patterns
Today's "Plain Search/Replace", "Regex Parser" and "Regex Search Replace" work with one pattern  only.  In some cases, it would be handy to run a list of patterns on the input field in one go.
This spares the user work adding individual transformers for each search/replace, will reduce teh number of intermediate columns and will improve performance.

Sample use case: standardizing umläute in German:
In sed script:
s/ue/u/gi
s/ü/u/gi
s/ae/a/gi
s/ä/a/gi
s/oe/o/gi
s/ö/o/gi
s/ß/ss/g
Now takes 7 search/replace transformers.

kaspersorensen:
Fixes #461 which again caused me to think "uh, that's pretty limiting that you cannot have multiple search/replace terms" ;-)


engine/core/src/main/java/org/datacleaner/components/convert/ConvertToDateTransformer.java
```diff
@@ -65,14 +65,17 @@
 
     @Inject
     @Configured(order = 2)
+    @Description("Default time zone to use if the date mask does not itself specify the time zone.")
```
LosD: These all seems unrelated?

components/basic-transformers/src/test/java/org/datacleaner/beans/transform/PlainSearchReplaceTransformerTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+
+import org.datacleaner.api.InputColumn;
+import org.datacleaner.data.MockInputColumn;
+import org.datacleaner.data.MockInputRow;
+import org.junit.Test;
+
+public class PlainSearchReplaceTransformerTest {
```
LosD: This class seems to already exist as a Scala class. At least that's what Travis complains about.

LosD:
LGTM, except for the duplicate class, and the unrelated changes.

engine/core/src/main/java/org/datacleaner/components/convert/ConvertToDateTransformer.java
kaspersorensen: True. But hopefully still a valid (small) improvement.

PR #1158, opened 2016-02-16
Centralized component duplicate removal

kaspersorensen:
I have been reviewing the remote components implementation and found a few things that I didn't really like about the way descriptor providers was organized.

 * Removed the server priority flag from the java code. The priority should be globally decided by the composite descriptor provider - regardless if it is based on remote or local (or custom) descriptor providers.
 * Related: Removed the "showAll" attribute from the conf.xml format. It was only available for remote components while it could be applicable for _all_ descriptor providers. It also didn't make sense for a user to set to true, so if we want someway to do this in dev/test mode then I suggest introducing a magic system property. But I even suggest simply not to do it - read on ...
 * Made the removal of duplicate components happen in the engine layer instead of the UI layer. The engine layer (DescriptorProvider interface and more) specifies that there is only one descriptor per name, so cheating that rule by simply returning a list with more than one was a bad choice IMO.

I would like a review by @jakubneubauer to ensure I didn't screw anything completely up. I did confirm on my own machine that everything is still working as before.


jakubneubauer:
I like the approach that components conflict is solved on the DescriptorProvider instead of the UI layer. 
And I agree with removal of the "showAll" flag. It seems to be a reminiscent of previous discussions when we were not able to make a decision...
The "priority" flag was there only to decide between conflicting component names - to preserve the order as is defined in XML. Because previous code used comparator and for that having a number was easier. But if I read your code changes right, the order is still preserved, so no problem here.

I believe we can merge.

PR #1162, opened 2016-02-16
Bug/1160 reference data case sensitivity

### LINKED ISSUE #1160: SynonymCatalog + Dictionary filenames in conf.xml gets double scheme
There seems to be an issue with the resource path (```<filename>```) of the file-based dictionaries and synonym catalogs in conf.xml when being edited by the desktop dialogs for this.

After saving a synonym catalog, then restarting datacleaner and then editing it again, I get a filename like this:
```
<filename>file://file:/datastores/job_title_synonyms.txt</filename>
```

Note: This particular example is a relative path (datastores/job_title_synonyms.txt) within DataCleanerHome. I have not tried it with an absolute path.

kaspersorensen:
Fixes #1160 

This PR implements support for case-insensitivity in synonym lookup and remove dictionary matches. It's also going quite some way to do less regex-based matching and more hashmap based lookups when possible - and also to minimize memory by only keeping multi-word lookups in copied data structures during matching (I'm sure @LosD will have his fun making remarks about that because it adds two ways of matching).

Compared to previously (prior to the multi-word support) the tokenization has been improved by doing the splitting based on word boundaries with Regex. And the same wrt. evaluating if a token is a "single word" - this is now based on a 80/20 rule where most terms will be found as "single words" but if there is any special characters or so that could act as word boundaries - then the lookup will be treated like a multi-word match.


LosD:
I don't really mind, as long as there is some level of guarantee that the matching isn't different depending on the path. If there isn't, I really think that it is a step backwards.

(I originally placed this comment in the issue instead of here, sorry for any confusing emails)

kaspersorensen:
Depending on which path do you mean?

LosD:
I mean, if results suddenly change because we find a term with a space in (and therefore taking another code path), it would be a (IMHO) very big issue.

kaspersorensen:
Agreed. So I'm hoping the tests cover that situation, but if I'm missing some test case for that then maybe that's a good thing for the reviewer to judge.

components/reference-data/src/main/java/org/datacleaner/beans/transform/SynonymLookupTransformer.java
```diff
@@ -156,7 +156,11 @@ public void close() {
         final String originalValue = inputRow.getValue(column);
 
         if (originalValue == null) {
-            return new String[1];
+            if (lookUpEveryToken) {
+                return new String[3];
```
LosD: Not really related, but this reminds me that during a demo, it was a bit of a pain that we don't have the synonyms found and masterterm found column when not doing every "token" (I guess that word is actually wrong in some cases now).

Would there be any drawbacks to always returning them?

components/reference-data/src/main/java/org/datacleaner/beans/transform/RemoveDictionaryMatchesTransformer.java
```diff
@@ -175,9 +194,26 @@ public void close() {
                     }
 
                     value = value.substring(0, start) + value.substring(end);
-                    removedParts.add(entry);
+                    removedParts.add(entry.getKey());
+                }
+            }
+            
+            // do word-by-word dictionary lookups
```
LosD: Ahhh! Misunderstood in first overview. This doesn't change the entire working method when a multiword entry is found, but is only different for those few words.

That makes it much better. My worry was that the working method of everything would change as soon as a multi word was inserted into the dictionary.

In reality, if there is any difference (besides the wanted difference), it is limited to those entries, which is much, much less critical. :+1:

engine/core/src/main/java/org/datacleaner/reference/SimpleSynonymCatalog.java
```diff
@@ -67,15 +68,37 @@ public SimpleSynonymCatalog(String name, Map<String, String> synonyms, boolean c
...
+            // synonym map
+            return _synonymMap;
+        }
+        final Map<String, String> synonymMap = new HashMap<>();
+        final Set<Entry<String, String>> entries = _synonymMap.entrySet();
+        for (Entry<String, String> entry : entries) {
+            final String synonym = entry.getKey();
+            final String masterTerm = entry.getValue();
+            if (StringUtils.isSingleWord(synonym)) {
+                synonymMap.put(synonym.toLowerCase(), masterTerm);
```
LosD: How come no _isCaseSensitive check here (like in the multiword dict)?

components/reference-data/src/test/java/org/datacleaner/beans/transform/SynonymLookupTransformerTest.java
```diff
@@ -29,13 +29,33 @@
...
 
-public class SynonymLookupTransformerTest extends TestCase {
+public class SynonymLookupTransformerTest {
 
     private final DataCleanerConfiguration configuration = new DataCleanerConfigurationImpl();
     private final SynonymCatalog sc = new TextFileSynonymCatalog("my synonyms",
             "src/test/resources/synonym-countries.txt", true, "UTF8");
 
+    @Test
+    public void testCaseInsensitiveMathing() throws Exception {
```
LosD: Small spelling mistake

LosD: And, since we've had issues in the case sensitiveness, maybe consider repeating the test without lookupEveryToken

engine/core/src/main/java/org/datacleaner/reference/SimpleSynonymCatalog.java
LosD: Ah. I'm a blind moron. It's coveret 12 lines up....

LosD:
LGTM, but see comments

components/reference-data/src/main/java/org/datacleaner/beans/transform/SynonymLookupTransformer.java
kaspersorensen: No I think it would be good to do that. Let's change add it always.

PR #1163, opened 2016-02-17
Issue #1160: Fixed by adding "file://" prefix support in FileResolver

### LINKED ISSUE #1160: SynonymCatalog + Dictionary filenames in conf.xml gets double scheme

kaspersorensen:
Fixes #1160 

Actually the solution is not very pretty because it is a file-only (as in - not any other resource types) fix. But it seems to be the minimum-impact resolution of the bug. The more general issue is that loading of dictionary and synonym resource is using the deprecated ```createFilename(...)``` method on the ```ConfigurationReaderInterceptor```.

It would be better (but a separate issue IMO) to use a real resource implementation but that would leak then also into standardizing how we write resource paths in XML (see ```DomConfigurationWriter.toFilename(Resource)``` which only does a partial job at this - candidate for a refactoring I would say).

### DIFF
--- a/engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
+++ b/engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
@@ -512,7 +512,7 @@ public final class JaxbConfigurationReader implements ConfigurationReader<InputS
         return environment.getStorageProvider();
     }
 
-    @SuppressWarnings("deprecation")
+    @Deprecated
     private String createFilename(String filename) {
         return _interceptor.createFilename(filename);
     }
@@ -559,8 +559,8 @@ public final class JaxbConfigurationReader implements ConfigurationReader<InputS
 
                         addVariablePath(name);
 
-                        String filenamePath = getStringVariable("filename", tfdt.getFilename());
-                        String filename = createFilename(filenamePath);
+                        final String filenamePath = getStringVariable("filename", tfdt.getFilename());
+                        final String filename = createFilename(filenamePath);
                         String encoding = getStringVariable("encoding", tfdt.getEncoding());
                         if (encoding == null) {
                             encoding = FileHelper.UTF_8_ENCODING;
--- a/engine/xml-config/src/test/resources/example-configuration-reference-data-resource-paths.xml
+++ b/engine/xml-config/src/test/resources/example-configuration-reference-data-resource-paths.xml
@@ -8,11 +8,23 @@
 				<encoding>UTF-8</encoding>
 				<case-sensitive>false</case-sensitive>
 			</text-file-dictionary>
+			
+			<text-file-dictionary name="dictionary2">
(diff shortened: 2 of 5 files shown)


engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
```diff
@@ -559,8 +559,8 @@ private ReferenceDataCatalog createReferenceDataCatalog(ReferenceDataCatalogType
 
                         addVariablePath(name);
 
-                        String filenamePath = getStringVariable("filename", tfdt.getFilename());
-                        String filename = createFilename(filenamePath);
+                        final String filenamePath = getStringVariable("filename", tfdt.getFilename());
+                        final String filename = createFilename(filenamePath);
```
LosD: Nooooooooooooooooooooooooooooooooooooooooooooooooooo NOT IN HERE! Then I'll need to merge when I do my commit. :scream: :sob: **whine**

engine/xml-config/src/test/resources/example-configuration-reference-data-resource-paths.xml
```diff
@@ -0,0 +1,22 @@
...
+		<synonym-catalogs>
+
+			<text-file-synonym-catalog name="synonyms">
+				<filename>file://relative/path/to/synonyms.txt</filename>
+				<encoding>UTF8</encoding>
+				<case-sensitive>false</case-sensitive>
+			</text-file-synonym-catalog>
+		</synonym-catalogs>
+	</reference-data-catalog>
+
```
LosD: Maybe add another one that isn't using `file://`just to make sure that those aren't screwed with (they probably aren't, but still).

LosD:
While true that it is a "lazy" fix, LGTM, but see comment (only the last one :smile:)

Will you add the issue?

kaspersorensen:
Created issue #1165 to refactor the way resources are externalized and read.

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
ClaudiaPHI: sissi :-1:

ClaudiaPHI: I am mean  :-P :-D

LosD: :D

PR #1164, opened 2016-02-17
Feature/1159 tokenizer num tokens

### LINKED ISSUE #1159: Tokenizer sets number of tokens to 10000 when switching to "Rows"
The tokenizer component switches the "Number of tokens" flag to 10000 when the user selects "rows" as token target. That's in itself not a problem, but if he switches back to "Columns" then he gets 10000 output columns! That's enough to give a serious issue in terms of blocking the user experience - and it does not make sense for anyone to tokenize into that many columns.

Rather we should hide the "Number of tokens" input text field and not set any value when the token target is "Rows".

kaspersorensen:
Fixes #1159

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/panels/tokenizer/TokenizerJobBuilderPresenter.java
+++ b/desktop/ui/src/main/java/org/datacleaner/panels/tokenizer/TokenizerJobBuilderPresenter.java
@@ -59,21 +59,26 @@ class TokenizerJobBuilderPresenter extends TransformerComponentBuilderPanel {
 	@Override
 	protected PropertyWidget<?> createPropertyWidget(ComponentBuilder componentBuilder,
 			ConfiguredPropertyDescriptor propertyDescriptor) {
-		PropertyWidget<?> propertyWidget = super.createPropertyWidget(componentBuilder, propertyDescriptor);
-		String propertyName = propertyDescriptor.getName();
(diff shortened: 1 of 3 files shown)


LosD:
LGTM. Has it been manually tested?

desktop/ui/src/main/java/org/datacleaner/panels/tokenizer/TokenizerJobBuilderPresenter.java
```diff
@@ -59,21 +59,26 @@ public TokenizerJobBuilderPresenter(TransformerComponentBuilder<?> transformerJo
...
+                    if (item == TokenTarget.ROWS) {
+                        if (!_numTokensPropertyWidget.isSet()) {
+                            _numTokensPropertyWidget.onValueTouched(2);
+                        }
+                        _numTokensPropertyWidget.setEnabled(false);
+                    } else {
+                        _numTokensPropertyWidget.setEnabled(true);
+                    }
+                }
 			});
```
LosD: Something looks a bit screwy here. I guess it's the tabs->space change. In this case the change is IMHO a large enough part to warrant a whole-file reformatting (otherwise I'd probably reformat the method, just to make sure that braces match)

kaspersorensen:
Fixed indentation and yes it has been manually tested. Merging.

Issue #1142, opened 2016-02-11
Sign in to DataCloud - Welcome screen.


khouzvicka:
DataCleaner 5 should by default have a remote connection to DataCloud configured.
The username/password should be blank by default. At the first startup, DataCleaner should ask the user for his datacleaner.org credentials and use these for the DataCloud connection.

PR #1168, opened 2016-02-18
Feature/1142 sign to datacloud

khouzvicka:
#1142 

New Datacloud dialog:
datacleaner welcome version 3

Default datacloud server name is "DataCloud" with url: "https://services.datacleaner.org".
In configuration will be only server name and url will be paar it in code.


api/src/main/java/org/datacleaner/configuration/RemoteServerData.java
```diff
@@ -33,4 +33,8 @@
     public String getUsername();
 
     public String getPassword();
+
+    public void setUsername(String username);
```
kaspersorensen: I did really like that this was immutable... It gives a lot of guarantees that mutability breaks.

So to explain a bit of background: Normally when you have an immutable structure, you don't try to mutate the existing objects - you rather just create a NEW object that represents the change.

desktop/api/src/main/java/org/datacleaner/windows/AbstractDialog.java
```diff
@@ -69,7 +73,18 @@ public AbstractDialog(WindowContext windowContext, Image bannerImage) {
         // make dialog itself focusable and add a listener for closing it when
         // ESC is typed.
         setFocusable(true);
-        addKeyListener(createDialogKeyListener());
+
+        getRootPane().getInputMap(JComponent.WHEN_IN_FOCUSED_WINDOW).put(KeyStroke.getKeyStroke(KeyEvent.VK_ESCAPE, 0), "EscapeAction");
+        getRootPane().getActionMap().put("EscapeAction", createEscapeAction());
```
kaspersorensen: Nice change. Could ideally have been in a separate PR, but OK :-)

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersConfigUtils.java
```diff
@@ -0,0 +1,100 @@
...
+import org.datacleaner.configuration.DataCleanerConfigurationImpl;
+import org.datacleaner.configuration.DataCleanerConfigurationUpdater;
+import org.datacleaner.repository.RepositoryFile;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+/**
+ * Utils for work with information about remote servers in configuration file
+ */
+public class RemoteServersConfigUtils {
```
kaspersorensen: I think actually we should avoid using this class (and DataCleanerConfigurationUpdater) because they stray away from the unified changes to conf.xml that we recently tried to implement in the class ```DomConfigurationWriter```. It would be best to only update conf.xml in a single way, which that class does well right now for datastores, dictionaries, synonym catalogs and string patterns. I would prefer if we simply add some remote component server awareness to that class.

kaspersorensen: ... and thereby also making it a non-UI class (located in ```xml-config``` module instead of ```desktop-ui```)

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
```diff
@@ -0,0 +1,106 @@
...
+ * Utilities for better work with remote servers.
+ */
+public class RemoteServersUtils {
+    final private DataCleanerConfiguration _configuration;
+
+    public RemoteServersUtils(final DataCleanerConfiguration configuration) {
+        _configuration = configuration;
+    }
+
+    public RemoteServerData getServerConfig(String serverName) {
```
kaspersorensen: This method should be added on ```RemoteServerConfiguration``` in my opinion - to allow general retrieval of server data based on name.

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
```diff
@@ -0,0 +1,106 @@
...
+    public void createRemoteServer(String serverName, String serverUrl, String userName, String password) {
+        String url = serverUrl;
+        if (RemoteDescriptorProvider.DATACLOUD_SERVER_NAME.equals(serverName) && url == null) {
+            url = RemoteDescriptorProvider.DATACLOUD_URL;
+        }
+        List<RemoteServerData> serverList =
+                _configuration.getEnvironment().getRemoteServerConfiguration().getServerList();
+        RemoteServerData remoteServerData =
+                new RemoteServerDataImpl(url, serverName, userName, password);
+        serverList.add(remoteServerData);
```
kaspersorensen: I would not add to a (immutable) list - rather: create a new list and a new ```RemoteDescriptorProvider```

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
```diff
@@ -0,0 +1,106 @@
...
+
+        final CompositeDescriptorProvider descriptorProvider =
+                (CompositeDescriptorProvider) _configuration.getEnvironment().getDescriptorProvider();
```
kaspersorensen: You can't be sure that it's a ```CompositeDescriptorProvider``` here, so the class cast is unsafe IMO.

desktop/ui/src/main/java/org/datacleaner/windows/DataCloudLogInWindow.java
```diff
@@ -0,0 +1,294 @@
...
+        getRootPane().getActionMap().put("EnterAction", new AbstractAction() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                signIn();
+            }
+        });
+
+    }
+
+    public static boolean mayIShowIt(UserPreferences userPreferences, DataCleanerConfiguration configuration) {
```
kaspersorensen: I would rename to something like ```isRelevant(...)``` or something like that. In code there is no "I" :-)

desktop/ui/src/main/java/org/datacleaner/windows/DataCloudLogInWindow.java
```diff
@@ -0,0 +1,294 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DataCloudLogInWindow extends AbstractDialog {
+    private static final Logger logger = LoggerFactory.getLogger(DataCloudLogInWindow.class);
+    public static final String SHOW_DATACLOUD_DIALOG_USER_PREFERENCE = "show.datacloud.dialog";
+
+    final private DataCleanerConfiguration _configuration;
+    final private UserPreferences _userPreferences;
+    final private JComponent _contentPanel;
```
kaspersorensen: styling-wise, "private" is always coming before "final".

desktop/ui/src/main/java/org/datacleaner/windows/DataCloudLogInWindow.java
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+
+    protected boolean isWindowResizable() {
+        return false;
+    }
+
+    @Override
+    public String getWindowTitle() {
+        return "DataCloud Sign In";
```
kaspersorensen: Wording wise I think it makes more sense to formulate it like: "Sign in to DataCloud"

desktop/ui/src/main/java/org/datacleaner/windows/DataCloudLogInWindow.java
```diff
@@ -0,0 +1,294 @@
...
+
+        @Override
+        public void changedUpdate(DocumentEvent e) {
+            invalidCredentialsLabel.setText("");
+        }
+    }
+
+    @Override
+    protected String getBannerTitle() {
+        return "Sign In / Register";
```
kaspersorensen: just call ```getWindowTitle()``` :)

engine/core/src/test/java/org/datacleaner/configuration/DataCleanerConfigurationUpdaterTest.java
```diff
@@ -29,13 +29,12 @@
...
 
 public class DataCleanerConfigurationUpdaterTest {
 
-    private String originalPassword = encode("admin");
-    private String newPassword = encode("newPassword");
     private String tagName = "password";
-    private String nodePathString = "descriptor-providers:remote-components:server:" + tagName;
+    private String nodePathString = "configuration/descriptor-providers/remote-components/server/" + tagName;
 
     private final Resource originalConfigurationResource = new FileResource("src/test/resources/conf.xml");
```
kaspersorensen: I think the filename here (conf.xml) should be named so that it is recognizable which test is using it, or else I look at src/test/resources and think it is just a general conf.xml file.

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -313,7 +313,7 @@ public void map(BatchSource<InputRow> source, BatchSink<Collection<Object[]>> si
         } catch (RuntimeException e) {
             boolean alreadyFailed = failed.getAndSet(true);
             if (!alreadyFailed) {
-                throw e;
+                throw new RuntimeException("Remote transformer failed: " + e.getMessage(), e);
```
kaspersorensen: This seems like a step back? I would prefer getting the cause of the error thrown rather than a wrapped exception (unless that wrapped exception is able to provide better details, which it doesn't look like).

api/src/main/java/org/datacleaner/configuration/RemoteServerData.java
khouzvicka: Each remote server provider has its data. In my case, when password is changed. I can do changes only in this object, and call refresh on descriptor provider. Everything is working.

But in your case. I will must find actual remote data. Remove it. Create new data, and after that set this data to right provider. Call refresh on providers.

If you wish, i will do this change, but my code is simple and elegant.

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
khouzvicka: I did changes in creating descriptor providers. Always is CompositeDescriptorProvider on top level

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersConfigUtils.java
khouzvicka: I will move DataCleanerConfigurationUpdater and RemoteServersConfigUtils to confix-xml modul. It is ok, for now?

api/src/main/java/org/datacleaner/configuration/RemoteServerData.java
kaspersorensen: Maybe we can reach a compromise by removing the setters from the interface at least. Then you need to do an ```instanceof RemoteServerDataImpl``` check and in case it returns false throw an UnsupportedOperationException or so. But that makes the mutability "hidden" for the moment which is good for the future if we want to enforce immutability.

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersConfigUtils.java
kaspersorensen: I would really like to see these classes go away. We recently spent a decent amount of time to ensure that ```DomConfigurationWriter``` is the only class with responsibility of changing/writing conf.xml. And it has all the utility-methods needed to do it well and safely. So a refactor here would really be helpful for maintainability.

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
kaspersorensen: But a ```DataCleanerConfiguration``` can actually be created in other ways than reading conf.xml. And then by default it will not be a composite. That could be in simple testing scenarios, in DC embedded inside Pentaho, in DataHub or in dataimprover.com ... And anyways the DataCleanerConfiguration interface only returns the ```DescriptorProvider``` interface so that is the contract we have to live up to. The point just being: It's not safe to assume that DataCleanerConfiguration is within your control.

jakubneubauer: To the im/mutability discussion: We thought that it will be better to not change the "main" Descriptor provider reference in the Environment. If we create a new composite Descr.Provider, then we need to replace the previous one in Environment. Is it good idea Kasper?

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
jakubneubauer: This is intentional. Previously, the REST Client error message contained something like "Remote transformer failed: ....". But the client can be used in another situations that from the transformer. So we removed the "Remote transformer" part from the error message in REST client. But we want that message "prefix" when doing a call in the RemoteTransformer.
So the new RuntimeException here is used to enhance the error message.

api/src/main/java/org/datacleaner/configuration/RemoteServerData.java
khouzvicka: Ok. :-)

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersConfigUtils.java
jakubneubauer: Just talked with Karel - he will put some effort to put the modification to DomConfigurationWriter. So working on it...

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
kaspersorensen: Ok :)

desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
kaspersorensen: I see ... That's in deed not so easy. OK let's keep it as is for now and see if it holds. Maybe I'm worrying just about scenarios that will never materialize because this method is only called from the regular desktop UI.

jakubneubauer:
So, what about merging this PR and work on graphics and text on separate branch?

kaspersorensen:
Ah yes, I will give it another look. In general: When you've finished fixing review remarks, then please give a comment here too so that notification emails are sent out. We don't get a notification for every commit.

desktop/ui/src/main/java/org/datacleaner/windows/DataCloudLogInWindow.java
```diff
@@ -0,0 +1,290 @@
...
+                        " them to improve your data quality." +
+                        " (You will get free credits to spend.)" +
+                        "<p>Enter your credentials to:" +
+                        "<ul style=\"list-style-type:none\">" +
+                        "   <li>\u2022 Validate your contacts addresses/emails." +
+                        "   <li>\u2022 Parse names, phones, emails." +
+                        "   <li>\u2022 Ensure you have recent addresses of your contacts." +
+                        "</ul>" +
+                        "<b>Your <a href=\"http://datacleaner.org\">datacleaner.org</a> credentials:</b>");
+        editorPane.setEditable(false);
```
kaspersorensen: Minor remark: DCHtmlBox is not editable by default so this is not needed.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
```diff
@@ -22,19 +22,40 @@
...
-    public RemoteServerConfigurationImpl() {
-        remoteServerDataList = new ArrayList<>();
+    public RemoteServerConfigurationImpl(List<RemoteServerData> serverData) {
+        remoteServerDataList = new ArrayList<>(serverData);
     }
-
     @Override
     public List<RemoteServerData> getServerList() {
-        return remoteServerDataList;
+        return UnmodifiableList.decorate(remoteServerDataList);
```
kaspersorensen: Normally I would use ```Collections.unmodifyableList(...)``` here. I think the apache-commons variant is there only for JDK4 users.

engine/xml-config/src/main/java/org/datacleaner/configuration/DomConfigurationWriter.java
```diff
@@ -872,4 +873,51 @@ private void appendElement(Element parent, String elementName, Object value) {
...
+
+        if(!StringUtils.isNullOrEmpty(url)){
+            createElement(serverElement, "url", url);
+        }
+        createElement(serverElement, "username", username);
+        createElement(serverElement, "password", SecurityUtils.encodePasswordWithPrefix(password));
+        onDocumentChanged(getDocument());
+    }
+
+    private void createElement(Element parent, String elementName, String textContext) {
```
kaspersorensen: See existing method ```appendElement(...)``` which has same purpose.

khouzvicka:
Everything is done :-) @kaspersorensen

kaspersorensen:
I think it looks and feels good. I wasn't able to log in using it, but that's due to network security constraints currently I guess (I get 403 forbidden). I'm in favor of merging.

I'm making a few notes here, but those are NOT to be fixed in this PR - I'll rather make a new issue about it, so just for reference:

 * I'm missing a cancel (or "Skip for now") button. I see that I can close the dialog but that feels less intuitive.
 * The Sign in button's icon is a DC logo. It should rather be a DataCloud logo (or a general login icon).
 * We need to rename "Remote components" tab in Options dialog to "DataCloud" - and the text accordingly.
 * The graphical banner that we've talked about should be added and the text in dialog should be revised a bit accordingly.
 * The user will always use his email address as username, so we should ask for his email rather than "Name".

khouzvicka:
I agree. I will make new branch for cahnges in "sign in" dialog. There will be new Datacloud banner and cancel button. And also other your suggestions will be there.

kaspersorensen:
I did make issue #1184 for it. I'm already implementing a few small things. Maybe you can wait for my PR ... It will only take an hour or so.

khouzvicka:
Ok. I will wait for it.

PR #1175, opened 2016-02-19
new remote component overlay

### LINKED ISSUE #1173: remote component icon overlay looks ugly
Especially for components with blue or green icons.

jakubneubauer:
fixes #1173 
Changes:

1. new icons for big and small variant
2. moved the overlaying from server to client. Previously, the server sent the image already overlayed. Now it sends pure image. This allows client overlay with different icons in different contexts (small/big size). Overlaying is presentation logic that should go to client anyway.


kaspersorensen:
Note, please do not merge yet - see discussion on the story #1173 :)

PR #1177, opened 2016-02-19
Remove unwanted characters cannot be added to canvas

### LINKED ISSUE #1153: Remove unwanted characters cannot be added to canvas
Something seems to have gone wrong with the remove unwanted characters transformer in 5.0-RC2: When adding it to the canvas, it just dissappears.

Grab:
image

Drag:
image

Release. Now it should be added, but instead... Nothing:
image

jhorcicka:
Fixes #1153 

RemoveUnwantedCharsTransformer was not declared as a public class. I didn't find the particular reason why package-private classes don't work, but making the class public (as other transformers) solves the problem.


kaspersorensen:
Very interesting fix! :-)
I'm wondering if we could detect this during annotation checking. Would be nice with a big warning in the logs in case this happens on other components.

PR #1179, opened 2016-02-24
Possibility to specify remote server in spring context

jakubneubauer:
Datacleaner monitor should be able to use remote components feature. (Other candidate is MDM).
This PR added a setter for additional descriptor provider in ConfigurationFactory. So now it is possible for example this:

```xml
	<bean name="configurationFactory" lazy-init="false"
		class="org.datacleaner.monitor.configuration.ConfigurationFactory">
		<property name="scannedPackages">
			<list>
				<value>org.datacleaner</value>
				<value>com.hi</value>
				<value>com.neopost</value>
			</list>
		</property>
		<property name="numThreads" value="30" />
		<property name="scanWebInfFolder" value="true" />
		<property name="additionalDescriptorProvider">
			<bean class="org.datacleaner.monitor.configuration.RemoteDescriptorProviderFactory">
				<property name="url" value="https://services.datacleaner.org"/>
				<property name="username" value="demo"/>
				<property name="password" value="password"/>
			</bean>
		</property>
	</bean>

```

 Note: Such an application must also add the DataCleaner-remote-components to dependencies.


kaspersorensen:
LGTM but an improvement idea would be to also create FactoryBean so that you can configured it in XML via named properties instead of constructor-arg (those are dependent on sequence which is harder to understand from the outside IMO).

jakubneubauer:
implemented the factory bean for RemoteDescriptorProvider (and updated description of this PR with the example XML snippet).

kaspersorensen:
LGTM!

PR #1180, opened 2016-02-24
Remote component output format

jakubneubauer:
This PR adds a possibility for a nicer output when calling transformer via REST API. Appending a GET parameter to the URL like this:
PUT /repository/tenant/components/MyTransformer?**outputFormat=columnMap**
will output for example this for Address correction (a row is a map, where key is an output column name):
```
{
  "rows": [
    [
      {
        "Traffic light (address)": "GREEN",
        "Thoroughfare": "Bizetstraat",
        "Country ISO code": "NL",
        "Sub premise": "",
        "Messages (address)": [],
        "Organization": "",
        "Sub administrative area": "Heemskerk",
        "Locality": "",
        "Dependent thoroughfare": "",
        "Country name": "Netherlands",
        "Post box": "",
        "House number": "10",
        "Administrative area": "Noord-Holland",
        "Premise": "",
        "Postal code": "1962 BX",
        "Delivery service qualifier": "",
        "Post town": "HEEMSKERK",
        "Name": "",
        "Department": "",
        "Dependent locality": "",
        "House number addition": ""
      }
    ]
  ]
}
```

instead of this:
```
{
  "rows": [
    [
      [
        "GREEN",
        "NL",
        "Netherlands",
        "Bizetstraat",
        "",
        "10",
        "",
        "1962 BX",
        "HEEMSKERK",
        "",
        "Noord-Holland",
        "Heemskerk",
        "",
        "",
        "",
        "",
        "",
        "",
        "",
        "",
        []
      ]
    ]
  ]
}
```

Default is to use the non-map variant, using array.


engine/rest-client/src/main/java/org/datacleaner/restclient/ComponentController.java
```diff
@@ -47,12 +47,8 @@
 
     /**
      * It creates a new component with the provided configuration, runs it and returns the result.
-     * @param tenant
-     * @param name
-     * @param processStatelessInput
-     * @return
      */
-    public ProcessStatelessOutput processStateless(final String tenant, final String name,
+    public ProcessStatelessOutput processStateless(final String tenant, final String name, String outputFormat,
```
kaspersorensen: Maybe would be better with an enum here so we control which outputFormats are applicable.

kaspersorensen:
I like the idea of this PR a lot. I think it would be good to do a quick API review before we merge it though. I am thinking here particularly about naming. Maybe also @losd can pitch in on that...

 * The parameter name is "outputFormat" which sounds reasonable, but it's also a little bit generic. It could easily be misunderstood as also "is it XML or JSON" or something like that. I don't have a better suggestion though, but thinking on it ...
 * The parameter value is "columnMap" which I somehow don't think fits. And I also think we need to think of what is the equivalent name of the default format. I'm thinking maybe to have these two values:
  * "tabular" (default) - this would be the row/array oriented format that we know. An alternative name could be "row-oriented".
  * "document" - this could be the new format which is more (key/value) document oriented. An alternative name could be "map-oriented" or so.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -29,12 +29,8 @@
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
LosD: This seems to break our formatting rules? (Same with the java.awt.* above)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -210,18 +207,41 @@ public OutputColumns getOutputColumns(@PathVariable(PARAMETER_NAME_TENANT) final
...
-        output.rows = getJsonNode(handler.runComponent(processStatelessInput.data, _maxBatchSize));
+        output.rows = getOutputJsonNode(handler, handler.runComponent(processStatelessInput.data, _maxBatchSize), outputFormat);
         output.result = getJsonNode(handler.closeComponent());
 
         return output;
     }
 
+    private JsonNode getOutputJsonNode(ComponentHandler handler, Collection<List<Object[]>> data, String outputFormat) {
+        if("columnMap".equals(outputFormat)) {
+            org.datacleaner.api.OutputColumns columns = handler.getOutputColumns();
```
LosD: This should probably be imported (I guess it is one of the * imports above that creates conflict)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -210,18 +207,41 @@ public OutputColumns getOutputColumns(@PathVariable(PARAMETER_NAME_TENANT) final
...
+            int columnCount = columns.getColumnCount();
+            List<List<Map<String, Object>>> columnMapOutput = new ArrayList<>(data.size());
+            for(List<Object[]> rowGroup: data) {
+                List<Map<String, Object>> columnMapRowGroup = new ArrayList<>(rowGroup.size());
+                for(Object[] row: rowGroup) {
+                    Map<String, Object> columMapRow = new HashMap<>(columnCount);
+                    for(int i = 0; i < columnCount; i++) {
+                        columMapRow.put(columns.getColumnName(i), row[i]);
```
LosD: I could imagine cases where legal DC column names could end up as illegal JSON, though I must admit I'm not sure.

LosD:
I don't really mind the outputFormat name. The document type (i.e. JSON or XML) would normally be done by mime content negotiation (the Accept header), so that shouldn't really clash.

But I do think the tabular/document change is better

jakubneubauer:
I agree with renaming outputFormat parameter. It can be really confused with the lower-level XML/JSON format. Even if we could use HTTP headers, it still can make confusion of users. So what about:

`outputStyle=tabular` (default)
`outputStyle=map`

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
jakubneubauer: The Jackson mapping library takes care of escaping special chars.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
jakubneubauer: I prefer the full class name on this place because there is used the same class name from different packages just on the same place (even on 2 consecutive lines). So because of better readability I kept the full name of the class.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -210,18 +213,41 @@ public OutputColumns getOutputColumns(@PathVariable(PARAMETER_NAME_TENANT) final
     @RequestMapping(value = "/{name}", method = RequestMethod.PUT, consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
     public ProcessStatelessOutput processStateless(@PathVariable(PARAMETER_NAME_TENANT) final String tenant,
             @PathVariable(PARAMETER_NAME_NAME) final String name,
+            @RequestParam(value = PARAMETER_NAME_OUTPUT_STYLE, required = false, defaultValue = "tabular") OutputStyle outputStyle,
```
jakubneubauer: Guys, is there a way how to use a enum constant here for the annotation attribute "defaultValue"? I tried also creating a static final String "constant" initialized with the enum.name(), but then compiler still complaints that "Attribute value must be constant".

I guess that compiler is clever enough to recignize that even static final field is not a real constant if it is initialized with a method call...

kaspersorensen: At this level I would not use the enum because that introduces an unfortunate link between java enum naming conventions (uppercase and underscore etc) which does not map well to http paramo. Rather I would make the string-to-enum conversion immediately after in your controller, and then pass it on to a method which takes the enum. That way you separate http interface (unsafe, dynamic) with java interface (safe, typed).

kaspersorensen:
I'm good with naming of "outputStyle". And let's make map a synonym for document too (as for string-to-enum mapping)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
LosD: D'oh. Of course. :)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
LosD: That might be a good reason for renaming the class.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -474,4 +501,30 @@ static void setPropertyType(ComponentDescriptor<?> descriptor, ConfiguredPropert
...
     }
+
+    public enum OutputStyle {
+
+        TABULAR,
+        MAP;
+
+        public static final String HTTP_PARAM_TABULAR = "tabular";
+        public static final String HTTP_PARAM_MAP = "map";
+        public static final String HTTP_PARAM_DOCUMENT = "document";
```
kaspersorensen: public static fields inside an enum ... That'll be pretty confusing when one tries to figure out what is what in this class. What I mean is that when a user in his IDE types "OutputStyle (dot)" then he will see a list of 5 elements: The two enums and the three public static fields. Somehow this conceptually feels wrong to me...

Do we need to declare them like this? I would be OK with just an inline string declaration below...

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -474,4 +501,30 @@ static void setPropertyType(ComponentDescriptor<?> descriptor, ConfiguredPropert
...
+
+        public static OutputStyle forString(String outputStyle) {
+
+            if(outputStyle == null) {
+                return TABULAR;
+            }
+            switch(outputStyle) {
```
kaspersorensen: maybe do a `.trim().toLowerCase()` on the string just to make it more fail-safe.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -474,4 +501,30 @@ static void setPropertyType(ComponentDescriptor<?> descriptor, ConfiguredPropert
...
+        TABULAR,
+        MAP;
+
+        public static final String HTTP_PARAM_TABULAR = "tabular";
+        public static final String HTTP_PARAM_MAP = "map";
+        public static final String HTTP_PARAM_DOCUMENT = "document";
+
+        public static OutputStyle forString(String outputStyle) {
+
+            if(outputStyle == null) {
```
kaspersorensen: Maybe also check for empty string. If the client is a tool it might easily submit an empty HTTP parameter instead of not including the param.

kaspersorensen:
I've made my adaptions - please review (and merge if you agree) @jakubneubauer :)

jakubneubauer:
thanks for repairs. Let's merge

PR #1181, opened 2016-02-25
Input Rewriter in Remote Component API

jakubneubauer:
Sorry for a long description, but this PR brings so many improvements... But it is actually simpler than it looks ;-)

This PR adds a hook just at the beginning of a remote transformer execution call, which allows to rewrite the user input.
One hook implementation was added, which allows users to provide simplified input and the rewriter modifies the input to a canonical form.
What is it good for? For example user doesn't need to specify a value for InputColumn property - in case the transformer has only one. All columns from data are assigned for the InputColumn property. Also, in case of one-column input user can provide simple value instead of array. Or, in case of one-row input, no need to provide array of rows, simple value is enough.

So for **Hostname resolver** transformer, instead of:
```json
{
  "configuration": {
    "properties": {
      "Hostname column": "c1"
    },
    "columns": ["c1"]
  },
  "data": [
    ["localhost"]
  ]
}
```
User can now send only this payload:
```json
{
  "data":"localhost"
}
```

Also, the rewriter hook recognizes "mapped property" (annotation @MappedProperty). In case the input has only one InputColumn and its corresponding mapped property associated with it, it is possible for example instead of this input for **Address correction** transformer:
```json
{
  "configuration": {
    "properties": {
      "Address part columns": ["c1", "c2", "c3"],
      "Address parts": ["POSTAL_CODE", "HOUSE_NUMBER",  "COUNTRY_ISO_CODE"]
    },
    "columns": ["c1", "c2", "c3"]
  },
  "data": [
    ["1962BX", "10", "NL"]
  ]
}
```

use this:
```json
{
  "data": {
    "POSTAL_CODE": "1962BX",
    "HOUSE_NUMBER": "10",
    "COUNTRY_ISO_CODE": "NL"
  }
}
```


kaspersorensen:
From your description it sounds like an awesome improvement! Will look into the code a bit later.

monitor/services/src/main/java/org/datacleaner/monitor/server/components/ComponentHandlerFactory.java
```diff
@@ -77,4 +88,23 @@ private void initialize() {
             analysisListener = new CompositeAnalysisListener(listeners.toArray(new AnalysisListener[listeners.size()]));
         }
     }
+
+    public ComponentDescriptor<?> resolveDescriptor(DataCleanerEnvironment env, String componentName) {
+        ComponentDescriptor<?> descriptor = env.getDescriptorProvider()
+                .getTransformerDescriptorByDisplayName(componentName);
+        if (descriptor == null) {
```
kaspersorensen: Yuck, maybe we should add a new method on ```DescriptorProvider``` like ```getComponentDescriptorByDisplayName(...)```. It should then check for all component types (incl. Filters which are missing here).

kaspersorensen:
Only a small suggestion, otherwise looks good.

kaspersorensen:
I'm gonna try and make that change myself ;)

kaspersorensen:
I've made my corrections as suggested - which also lead to a small but nice improvement/simplification in `JaxbJobReader`. Please review (and merge if you're happy too) @jakubneubauer :)

PR #1182, opened 2016-02-26
Case insensitive "Remove substring"

### LINKED ISSUE #1151: Add flag for case sensitiveness in "Remove substring" transformation
Today the ```Remove substring``` transformation is case sensitive. It would be nice to allow case-insensitive removal with a flag somehow.

We just had to do it and it required us to use two separate ```Text case transformers``` before passing data into the ```Remove substring``` component. Kinda annoying.

kaspersorensen:
Fixes #1151

The approach to removing substrings has been changed a bit so that the code flow is not too nested into a lot of if-else's.

I think the test case is now rather good.


jhorcicka:
LGTM

PR #1194, opened 2016-03-02
Fixed problems with Swagger JSON controller

jakubneubauer:
Fixes mean:
- only "live" controllers are reported, not all from classpath
- host,path in response is properly generated, also supporting our custom header X-Original-Url in case monitor is running behind reverse proxy with different URL.


monitor/documentation/src/main/java/org/datacleaner/documentation/swagger/SecuritySchemeObject.java
```diff
@@ -0,0 +1,28 @@
...
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.documentation.swagger;
+
+import java.util.HashMap;
+
+/**
+ * Added by jakub on 1.3.16
+ */
+public class SecuritySchemeObject extends HashMap {
```
kaspersorensen: Hmm, why do we need to extend HashMap? If it's needed, please document what the role of this class is.

kaspersorensen: And should we not add stuff to avoid compiler warnings ... such as generic arguments and the serialVersionUID?

monitor/documentation/src/main/java/org/datacleaner/documentation/swagger/SwaggerInfo.java
```diff
@@ -70,11 +70,11 @@ public void setContact(SwaggerContact contact) {
...
-    public SwaggerLicense getLicense() {
-        return license;
-    }
-
-    public void setLicense(SwaggerLicense license) {
-        this.license = license;
-    }
+//    public SwaggerLicense getLicense() {
+//        return license;
+//    }
```
kaspersorensen: If it's not needed, let's delete it fully rather than commenting it.

jakubneubauer:
all comments were processed, we can merge.

kaspersorensen:
LGTM.

PR #1203, opened 2016-03-04
Enable custom execute buttons and analysis runners

arjansh:
Make it possible to add extra execute buttons to the job builder window and make it possible to have these jobs run using custom `AnalysisRunner` implementations.

You can add your own `ExecutionMenuItem` implementations and register them through the "datacleaner.ui.desktop.extra.execute.buttons" system property so they will be added to the execute buttons in the job builder window.

Such an `ExecutionMenuItem` implementation can run in a custom implemented `ResultWindow` using a custom `AnalysisRunner`.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/ExecuteButtonBuilder.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/ExecuteButtonBuilder.java
@@ -21,6 +21,8 @@ package org.datacleaner.widgets;
 
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
 import java.util.List;
 
 import javax.swing.JButton;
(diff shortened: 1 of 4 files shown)


LosD:
Code looks good, but I'm not super fond of the focus on "button", which seems to be a UI detail (and maybe even a wrong detail; a menu item is not really a button). I'd prefer something that more focuses on that it's a new way to execute.

LosD:
I'm not super sure about the extensibility of ResultWindow... For ourselves, we'll only use it temporarily, so it seems a little useless, OTOH the flexibility doesn't hurt.

Let's keep it unless @kaspersorensen or someone else has objections.

arjansh:
I understand your reservation about the extensibility of `ResultWindow`, but I saw no other (easy) manner to inject a custom `AnalysisRunner` into the window. If possible however I would prefer not having to extend `ResultWindow` to be able to do that.

LosD:
If anyone has objections, we can keep it non-final until the rest of the task is complete, then seal it again. Otherwise I think I really don't mind. I can't see any long-term reason to use a non-standard ResultWindow for long, but I can't really see issues with it either.

ClaudiaPHI:
Tested and it works nice.

desktop/ui/src/main/java/org/datacleaner/widgets/ExecuteButtonBuilder.java
```diff
@@ -38,11 +40,21 @@
...
  */
 public class ExecuteButtonBuilder {
+    private static final Logger LOG = LoggerFactory.getLogger(ExecuteButtonBuilder.class);
+    
+    /**
+     * Property which may contain a comma separated vector of class names
+     * implementing the ExecutionMenuItem interface, which are added as extra
+     * execute buttons to the job builder window of DataCleaner.
+     */
+    public static final String EXTRA_BUTTON_CLASSES = "datacleaner.ui.desktop.extra.execute.buttons";
```
kaspersorensen: Shouldn't this be called something with "menuitem" instead of "button"? Or maybe just "options" to make it less widget-specific?

kaspersorensen:
I checked an extra time and have a more fundamental question: The ```ExecuteButtonOptions``` class already has static methods like `addMenuItem(...)`. Why is a new system property even needed? I think what you want to do (in your custom distribution) is simply to call the static method in your launcher? No?

arjansh:
The problem here is that using the static `addMenuItem(...)` methods I have no control over the manner in which the `ExecuteMenuItem' is constructed and, as far as I know, I cannot inject instances of objects into them.

kaspersorensen:
That's the reason that important arguments are passed in the `createActionListener(...)` method of `ExecutionMenuItem`. Maybe we need to make it a bit more elaborate in terms of arguments. But they idea was that this method would be able to serve the same need. I think that what we have now instead is two ways of doing the same thing.

PR #1206, opened 2016-03-04
EnumerationProvider conversion from string to EnumerationValue

jakubneubauer:


LosD:
A shame with the duplicate code, but it's only a few lines, so LGTM

PR #1210, opened 2016-03-07
Bug/1205 dc banner appers too many times

### LINKED ISSUE #1205: DataCloud sign in screen appears one to many times
DataCloud sign in/welcoming screen appears while opening jobs as well.

ClaudiaPHI:
Fixes #1205

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/bootstrap/Bootstrap.java
+++ b/desktop/ui/src/main/java/org/datacleaner/bootstrap/Bootstrap.java
@@ -69,7 +69,9 @@ import org.datacleaner.util.VFSUtils;
 import org.datacleaner.util.WidgetUtils;
 import org.datacleaner.util.http.MonitorHttpClient;
 import org.datacleaner.util.http.SimpleWebServiceHttpClient;
+import org.datacleaner.windows.AbstractWindow;
 import org.datacleaner.windows.AnalysisJobBuilderWindow;
+import org.datacleaner.windows.DataCloudLogInWindow;
 import org.datacleaner.windows.MonitorConnectionDialog;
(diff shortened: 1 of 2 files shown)


kaspersorensen:
I would like to suggest an alternative solution: The dialog should not be tied to the ```AnalysisJobBuilderWindow``` _at all_. It has nothing to do with showing that window - it was a wrong idea to begin with. Rather it only has something to do with application startup. So the instruction to do it (without needing any counter or anything like that) should be done in the ```Bootstrap``` class which eventually also opens the (first) ```AnalysisJobBuilderWindow```

kaspersorensen:
A good side-effect of that is also that we decrease the amount of code in ```AnalysisJobBuilderWindow``` - which is already a too complex class as it is.

ClaudiaPHI:
oki doki

ClaudiaPHI:
Moved DataCloud banner to ```Bootstrap``` class.

desktop/ui/src/main/java/org/datacleaner/bootstrap/Bootstrap.java
```diff
@@ -249,8 +251,17 @@ public void run() {
                 });
             }
 
-            final WindowContext windowContext = injector.getInstance(WindowContext.class);
-
+            WidgetUtils.invokeSwingAction(new Runnable() {
+                @Override
+                public void run() {
+                    if (DataCloudLogInWindow.isRelevantToShow(userPreferences, configuration)) {
+                        DataCloudLogInWindow dataCloudLogInWindow = new DataCloudLogInWindow(configuration,
```
LosD: final

LosD:
Nice and simple. LGTM (except a single missing final)

PR #1214, opened 2016-03-08
Add AnalysisJobBuilderWindow to ExecutionMenuItem methods

arjansh:
Add AnalysisJobBuilderWindow to createActionListener (and run) methods on ExecutionMenuItem interface and abstract SimpleExecutionMenuItem implementation to make it available to custom ExecutionMenuItem implementations (which typically extend SimpleExecutionMenuItem).


LosD:
What do we need the AnalysisJobBuilderWindow for?

arjansh:
This way custom ExecutionMenuItems have access to the `WindowContext` which the `AnalysisJobBuilder` doesn't provide access to.

LosD:
Okay. LGTM then.

PR #1215, opened 2016-03-08
Fixes #1196: Only check there are enough (but not same amount) columns

### LINKED ISSUE #1196: Synonym lookup changes has broken old jobs
The changes in #1116 and/or #1152 has introduced 2 new output columns to the synonym lookup transformer. Unfortunately that has broken old DC jobs, as it makes the job validation fail.

kaspersorensen:
Fixes #1196 

Does so by using `<` instead of `!=` when comparing expected vs actual output columns on loading a job. The effect is that it's OK for component authors to add output columns, but not to remove them. A corner case (which also existed before) is if the meaning of an output column is changed - in that case the component developer has (still) buried a hole for himself ;-)


ClaudiaPHI:
LGTM!

PR #1216, opened 2016-03-09
Launching spark job with resultfile URI fails

arjansh:
When you provide a location for the results of the job which is executed on the Hadoop cluster through the SparkLauncher, a temporary properties file is created containing the location for the results. This properties file is then copied to the Hadoop file system, but this doesn't work (and silently fails), because the hdfs path which is created (and to which the properties file is copied) contains the absolute location of the properties file as it has been created on your local machine. This is easily fixed by just adding the properties file name to the hdfs path, instead of its absolute location.


kaspersorensen:
LGTM. Merge when travis is green.

PR #1217, opened 2016-03-09
Fixes #1161 - text cases transformers

### LINKED ISSUE #1161: Provide a default "Mode" for TextCaseTransformer
The "Text case transformer" currently has no default setting for which case should be applied. That makes it a less fluent experience to configure it if you're just looking to standardize values to "something". It's a common case that you just want to either uppercase or lowercase or whatever is enough to bring diverting values in line.

So let's make a default selection for the user. I vote for UPPER case as the default.

ClaudiaPHI:
Fixes #1161 . 

I have also added another scope to this story. I have seen that the transformation mode ```CAPITALIZE_WORDS ``` doesn't capitalize the words separated by '-' .  So I fixed that as well. 

capitalize


kaspersorensen:
LGTM!

PR #1218, opened 2016-03-09
Support for enumeration aliases in remote transformer

jakubneubauer:
Is important in UI, when suggesting mapping for @MappedProperty property.


engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
```diff
@@ -250,11 +251,25 @@ public void acceptJsonFormatVisitor(JsonFormatVisitorWrapper visitor, JavaType t
         }
 
         protected String enumValueToSchemaString(Enum<?> value) {
+            StringBuilder serialized = new StringBuilder();
+            serialized.append(value.name());
+            serialized.append("::");
```
kaspersorensen: Since "::" is referred to both here and in JsonSchemaConfiguredPropertyDescriptorImpl, then I suggest to share a common constant so that it never comes out of sync.

jakubneubauer:
added constant for aliases separator.

Issue #1211, opened 2016-03-08
ClassCastException thrown when logging in to DataCloud


kaspersorensen:
This happened to me (on master branch) when logging in to DataCloud:

![image](https://cloud.githubusercontent.com/assets/291450/13599198/0483b704-e522-11e5-9f08-e025d611c3ba.png)

Before this I had deleted all `userhome/.datacleaner` folders to ensure that everything was completely clean.

Stack trace:

```
ERROR 11:34:41 DCUncaughtExceptionHandler - Thread AWT-EventQueue-0 threw uncaught exception
java.lang.ClassCastException: org.datacleaner.descriptors.ClasspathScanDescriptorProvider cannot be cast to org.datacleaner.descriptors.CompositeDescriptorProvider
	at org.datacleaner.util.RemoteServersUtils.addRemoteServer(RemoteServersUtils.java:54)
	at org.datacleaner.windows.DataCloudLogInWindow$2.run(DataCloudLogInWindow.java:292)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:744)
	at java.awt.EventQueue.access$400(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:697)
	at java.awt.EventQueue$3.run(EventQueue.java:691)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$1.doIntersectionPrivilege(ProtectionDomain.java:75)
```

kaspersorensen:
Note: It does look like the conf.xml file was properly updated afterwards.

PR #1221, opened 2016-03-10
Fix of problem  - ClassCastException thrown when logging in to DataCloud

khouzvicka:
#1211  
Fix of problem - ClassCastException thrown when logging in to DataCloud

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
+++ b/desktop/ui/src/main/java/org/datacleaner/util/RemoteServersUtils.java
@@ -51,6 +51,10 @@ public class RemoteServersUtils {
         }
 
         RemoteServerData remoteServerData = new RemoteServerDataImpl(serverUrl, serverName, userName, password);
+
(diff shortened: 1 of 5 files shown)


kaspersorensen:
For now this does the trick, but I'm honestly not so happy with the fix because it just strengthens an "assumption based interface" where `DescriptorProvider` _has_ to be a composite implementation, while the `DataCleanerEnvironment` interface prescribes that it is just _any_ implementation (it returns the interface). I do however realize that time is slight and making a proper fix for this (which would entail making a _new_ environment and putting that to work in the desktop app's session) is a bit cumbersome.

For the sake of future-proofing, I'd like to make @losd aware of this issue, since he will be paying the technical debt of it ... some day.

Can we please also add a more managed exception (`IllegalStateException` is fine) in `RemoteServersUtils` if we find that the `DescriptorProvider` is NOT a composite? The least we can do at this point is to inform the next developer (via the exception message) that there is a pitfall here which we have not covered completely.

kaspersorensen:
LGTM now :)

PR #1226, opened 2016-03-10
Bug/1219 value distribution isdistributable

### LINKED ISSUE #1219: Spark submit to Hadoop doesn't work with Value Distribution Analyzer
When you use Spark to submit a job to a Hadoop cluster, it fails if the job contains a Value Distribution Analyzer, with this stacktrace:

```
16/03/09 12:55:56 ERROR ApplicationMaster: User class threw exception: java.lang.IllegalStateException: Cannot assign an array-value ([Lorg.datacleaner.api.InputColumn;@1177c16) to a non-array property (ConfiguredPropertyDescriptorImpl[name=Column])
java.lang.IllegalStateException: Cannot assign an array-value ([Lorg.datacleaner.api.InputColumn;@1177c16) to a non-array property (ConfiguredPropertyDescriptorImpl[name=Column])
	at org.datacleaner.lifecycle.AssignConfiguredPropertiesHelper.assignProperties(AssignConfiguredPropertiesHelper.java:50)
	at org.datacleaner.lifecycle.LifeCycleHelper.assignConfiguredProperties(LifeCycleHelper.java:125)
	at org.datacleaner.job.builder.AbstractComponentBuilder.getComponentInstanceForQuestioning(AbstractComponentBuilder.java:761)
	at org.datacleaner.job.builder.AbstractComponentBuilder.isDistributable(AbstractComponentBuilder.java:280)
	at org.datacleaner.job.builder.AnalysisJobBuilder.isDistributable(AnalysisJobBuilder.java:1419)
	at org.datacleaner.spark.SparkAnalysisRunner.run(SparkAnalysisRunner.java:93)
	at org.datacleaner.spark.Main.main(Main.java:76)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.spark.deploy.yarn.ApplicationMaster$$anon$2.run(ApplicationMaster.scala:542)
```

kaspersorensen:
Fixes #1219

... by making `getComponentInstanceForQuestioning()` work with `AnalyzerComponentBuilder`s that build multiple component. This prevented the `isDistributable()` call from working before.


kaspersorensen:
Finally (at least on my machine) this PR builds ... Still waiting for Travis.
But the reviewer will undoubtly discover that this fix did not only deal with Pattern finder and Value distribution. In fact while fixing it, I discovered a strange thing in String analyzer, namely that it had the `@ColumnProperty(escalateToMultipleJob=true)` flag set. I was quite confused by that and pursued to find out what was going on...

Turned out that with this flag (and unfortunately _only_ with this flag) it was made possible to have the configured `InputColumn`-array act as a "escalating mechanism" for having columns from many tables configured on the String analyzer:
image

What this then does is to create two instances of the String analyzer - one for each table. But in fact I would expect that to be possible with _any_ component having such an `InputColumn[]` property and it should certainly not be bound to the `escalateToMultipleJob` flag.

So when I tried to resolve that, I ran into new issues because it seemed that all kinds of special behaviour was in place for enabling the escalation mechanism. I've spent some (bit too much now) effort to clean that up. It eventually also revealed other issues, such as it was possible to wire input columns of the wrong type to such a property which is "escalating".

kaspersorensen:
Nasty after-remark ... The test failures that sort of led me through this where good ... but ... they where shattered over the following projects:

 * engine-core
 * basic-analyzers
 * value-distribution
 * pattern-finder
 * xml-config
 * desktop-ui

Ideally they would all have been found (fail fast) in engine-core since that's where the real changes where made. Good thing though is that collectively our coverage of this seems very good.

components/basic-analyzers/src/test/java/org/datacleaner/job/builder/AnalyzerComponentBuilderTest.java
```diff
@@ -57,15 +57,15 @@ public void testNoOriginatingTableBecauseOfMockColumns() throws Exception {
...
 
-        Table table1 = new MutableTable("table1");
+        final Table table1 = new MutableTable("table1");
         jobBuilder.addInputColumn(new MetaModelInputColumn(
                 new MutableColumn("foo", ColumnType.VARCHAR, table1, 0, true)));
         jobBuilder.addInputColumn(new MetaModelInputColumn(
                 new MutableColumn("bar", ColumnType.VARCHAR, table1, 1, true)));
 
-        Table table2 = new MutableTable("table2");
+        final Table table2 = new MutableTable("table2");
```
LosD: Very minor, and don't take them out now, but IMHO final fixes in irrelevant code do more harm than good: They take focus away from the real changes.

LosD:
Completely irrelevant to this PR, but since you've been looking so much into isDistributable: The default  rule seems to say filters and transformers are distributable, analyzers are not... But what if a transformer implements HasAnalyzerResult... Will/should that change the default? Somehow it should, I guess, but a cursory look over the code doesn't seem to agree.

desktop/ui/src/test/java/org/datacleaner/job/runner/RowProcessingConsumerSorterTest.java
```diff
@@ -76,7 +77,7 @@ public void testCreateProcessOrderedConsumerListWithMergedOutcomes() throws Exce
         fjb1.setName("fjb1");
 
         // 2: trim (depends on filter)
-        TransformerComponentBuilder<TransformerMock> tjb1 = ajb.addTransformer(TransformerMock.class);
+        TransformerComponentBuilder<MockTransformer> tjb1 = ajb.addTransformer(MockTransformer.class);
```
LosD: No need to do it now, but should the `TransformerMock` class die completely, or does it serve a purpose in the rest of `RowProcessingConsumerSorterTest`? It seems to be the only class that doesn't use the normal `MockTransformer`

LosD:
LGTM, but it's pretty hairy stuff, so I'm not sure I would have caught any blunders.

I have a few questions, but none of them blocks a merge, so... Merging :)

desktop/ui/src/test/java/org/datacleaner/job/runner/RowProcessingConsumerSorterTest.java
kaspersorensen: I did look if it was used still, and still 2 unittests use it... But probably it can be replaced easily if we want to.

components/basic-analyzers/src/test/java/org/datacleaner/job/builder/AnalyzerComponentBuilderTest.java
kaspersorensen: Good point... I was touching or at least monitoring closely the test method so I decided to fix it, but maybe not nice.

kaspersorensen:
Regarding `HasAnalyzerResult` and `isDistributable()` conventions ... You might be right. I simply didn't consider it, so no conscious decision was made on that from my side.

PR #1227, opened 2016-03-10
Fixes #1191 by doing an eager check if displayed TableModel is OK

### LINKED ISSUE #1191: Give up when trying to preview an input datastore with errors
If a user tries to preview a datastore with errors, the preview will keep trying to read it, even if there is an error (i.e. wrong column count for a CSV row). This means that the preview dialog will keep putting up a blocking error dialog, which makes DC unusable until killed.

kaspersorensen:
Fixes #1191

See #1191 for investigation details.
I also updated the error dialog shown when a DataSetWindow is unable to render. This was to add a little bit of detail around the error (but still referring to the log file for more details).

screen shot 2016-03-10 at 23 01 10

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/widgets/table/DCTable.java
+++ b/desktop/api/src/main/java/org/datacleaner/widgets/table/DCTable.java
@@ -307,17 +307,11 @@ public class DCTable extends JXTable implements MouseListener {
 
     @Override
     public Object getValueAt(int row, int column) {
-        try {
-            Object value = super.getValueAt(row, column);
-            if (value == null) {
-                value = LabelUtils.NULL_LABEL;
-            }
-            return value;
-        } catch (Exception e) {
-            // TODO
-            System.out.println("!!! " + e.getMessage());
-            return null;
+        Object value = super.getValueAt(row, column);
+        if (value == null) {
+            value = LabelUtils.NULL_LABEL;
         }
+        return value;
     }
 
     public void setVisibleColumns(int min, int max) {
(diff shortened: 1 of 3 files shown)


desktop/api/src/main/java/org/datacleaner/widgets/table/DCTable.java
```diff
@@ -307,11 +307,17 @@ public String getTextValueAt(int row, int column) {
 
     @Override
     public Object getValueAt(int row, int column) {
-        Object value = super.getValueAt(row, column);
-        if (value == null) {
-            value = LabelUtils.NULL_LABEL;
+        try {
+            Object value = super.getValueAt(row, column);
```
LosD: You could do that with final if you just return NULL_LABEL directly :)

desktop/api/src/main/java/org/datacleaner/widgets/table/DCTable.java
```diff
@@ -307,11 +307,17 @@ public String getTextValueAt(int row, int column) {
...
+            if (value == null) {
+                value = LabelUtils.NULL_LABEL;
+            }
+            return value;
+        } catch (Exception e) {
+            // TODO
```
LosD: Details of what TODO would be nice

LosD:
Minor issues, otherwise LGTM

desktop/api/src/main/java/org/datacleaner/widgets/table/DCTable.java
kaspersorensen: Oh that try block shouldn't have been committed ... It was an experiment that I obviously forgot to clean up!

PR #1228, opened 2016-03-11
Add remote components to Spark deps

### LINKED ISSUE #1225: Add remote components to Spark jar
The Spark jar currently does not support remote components, as the module is not included in the env-spark dependencies.

Add dependency and check that remote components works as intended.

LosD:
This adds the remote components module to the dependencies of the Spark module.

Fixes #1225


arjansh:
LGTM!

PR #1232, opened 2016-03-14
Adding security headers to cookie retrieval request for CAS when comm…

MennoB:
…unicating with DataHub

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/http/CASMonitorHttpClient.java
+++ b/engine/core/src/main/java/org/datacleaner/util/http/CASMonitorHttpClient.java
@@ -131,13 +131,13 @@ public class CASMonitorHttpClient implements MonitorHttpClient {
         // now we request the spring security CAS check service, this will set
         // cookies on the client.
         final HttpGet cookieRequest = new HttpGet(_requestedService + "?ticket=" + ticket);
+        addSecurityHeaders(cookieRequest);
         final HttpResponse cookieResponse = executeHttpRequest(cookieRequest, context);
         EntityUtils.consume(cookieResponse.getEntity());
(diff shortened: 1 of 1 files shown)


LosD:
Code (all one lines of it :)) looks good to me, but I know nothing about CAS.

I guess we can merge if/when Travis is happy.

arjansh:
I also know nothing about CAS, but I see that you now apply the addSecurityHeaders method to both the `cookieRequest` variable and the `request` argument. Can you be certain that both these requests go to the same host and need the same security headers?

MennoB:
@arjansh The headers contain the information that is necessary to get CAS authorisation. Every call needs to be verified by CAS, therefore every call will go through CAS first and for this it will need those same headers every time.

arjansh:
LGTM then!

LosD:
Okay, so I guess it is to make sure the payload (including HTTP headers) is the same?

LosD:
Would it maybe be a good idea to check the response, so errors like this are caught?

MennoB:
@LosD In the case of DataHub's CAS the ticket and user are relayed in the headers. So that's why they're necessary there. I can add a status code check for the cookie request, that might be useful.

LosD:
Ah, makes sense.

Github doesn't send notifications when a commit is added, so if you make a comment when the check has been added, we'll react a bit faster :)

MennoB:
Added a check for the status code when retrieving cookies. If we should fail harder here, please let me know.

LosD:
Let's leave it like this for now... I'm not sure how stuff reacts on a hard failure, and the warning should at least help track down problems.

Merge when Travis is happy

LosD:
Travis happy. Merging! :)

PR #1233, opened 2016-03-14
Bug/filter transitivity

### LINKED ISSUE #1192: Preview still has issues
Even though we did a lot of preview-fixing, we're still having issues at a customer. The job looks like
Table->Equals filter->Concatenator->Whitespace trimmer

Previewing the concatenator gives fine results, but previewing the whitespace trimmer gives `<null>` for all inputs and outputs.

kaspersorensen:
Fixes #1192

Approach taken was to check whether the transformer being previewed was already under some filter condition. If so, we only apply Max rows condition to the existing filters. That ensures that it becomes an AND filtering scenario instead of a logical OR.

The __second__ approach taken was not related to the preview per se, but at the engine level (`AbstractRowProcessingConsumer`). Here I found that the `_sourceJobsOfInputColumns` which is used to determine if filtered sources of columns (transformers with filter requirements on them) are handled in a logical OR fashion. In the case of the reproducing .analysis.xml file, a filter was added to this list and thus it acted as a gate-opener for the OR condition which shouldn't have happened.

So, actually this PR fixes two bugs :-)

(also a small commit to make internet-dependent unittests pass automagically if you're offline).

Note from my side: I think this change is testament that our "flattening" of the graph into a list of components is starting to show it's weaknesses. I believe that it is fixed now, but I could imagine that with more and more complex jobs we will see an increase in complexity also in the calculation of requirements for processing a particular item in the list. It does seem like we soon should refactor the engine to do true graph processing - it starts to seem that it would simplify rather than make more complex.

### DIFF
--- a/components/network-tools/src/test/java/org/datacleaner/extension/networktools/ResolveHostnameTransformerTest.java
+++ b/components/network-tools/src/test/java/org/datacleaner/extension/networktools/ResolveHostnameTransformerTest.java
@@ -24,6 +24,7 @@ import static org.junit.Assert.assertEquals;
 import org.datacleaner.data.MockInputColumn;
 import org.datacleaner.data.MockInputRow;
 import org.datacleaner.test.TestHelper;
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -53,9 +54,8 @@ public class ResolveHostnameTransformerTest {
     
     @Test
     public void testTransformInternet() throws Exception {
-        if (!TestHelper.isInternetConnected()) {
-            return;
-        }
+        Assume.assumeTrue(TestHelper.isInternetConnected());
+        
         assertEquals("94.142.215.39", t.transform(new MockInputRow().put(col, "eobjects.org"))[0]);
     }
 
--- a/components/network-tools/src/test/java/org/datacleaner/extension/networktools/ResolveHostnameTransformerTest.java
+++ b/components/network-tools/src/test/java/org/datacleaner/extension/networktools/ResolveHostnameTransformerTest.java
@@ -19,38 +19,48 @@
  */
 package org.datacleaner.extension.networktools;
 
+import static org.junit.Assert.assertEquals;
+
 import org.datacleaner.data.MockInputColumn;
 import org.datacleaner.data.MockInputRow;
+import org.datacleaner.test.TestHelper;
+import org.junit.Before;
+import org.junit.Test;
 
-import junit.framework.TestCase;
+public class ResolveHostnameTransformerTest {
 
-public class ResolveHostnameTransformerTest extends TestCase {
+    private final ResolveHostnameTransformer t = new ResolveHostnameTransformer();
+    private final MockInputColumn<String> col = new MockInputColumn<String>("host", String.class);
 
-	public void testTransform() throws Exception {
-		ResolveHostnameTransformer t = new ResolveHostnameTransformer();
-		MockInputColumn<String> col = new MockInputColumn<String>("host",
-				String.class);
-		t.hostnameColumn = col;
+    @Before
+    public void setUp() {
+        t.hostnameColumn = col;
+    }
 
-		assertEquals("OutputColumns[host (ip address)]", t.getOutputColumns()
-				.toString());
+    @Test
+    public void testTransformLocal() throws Exception {
+        assertEquals("127.0.0.1", t.transform(new MockInputRow().put(col, "localhost"))[0]);
 
-		assertEquals("127.0.0.1",
-				t.transform(new MockInputRow().put(col, "localhost"))[0]);
+        assertEquals("127.0.0.1", t.transform(new MockInputRow().put(col, "127.0.0.1"))[0]);
+        
+        assertEquals(null, t.transform(new MockInputRow().put(col, ""))[0]);
+        
+        assertEquals(null, t.transform(new MockInputRow().put(col, null))[0]);
+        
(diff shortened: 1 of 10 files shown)


desktop/api/src/main/java/org/datacleaner/actions/PreviewTransformedDataActionListener.java
```diff
@@ -162,6 +167,8 @@ protected PreviewJob createPreviewJob() {
                     ajb.removeSourceTable(otherTable);
                 }
             }
+            alreadyFiltered = sourceColumnFinder.findAllSourceJobs(tjb).stream().filter(
+                    o -> o instanceof HasFilterOutcomes).findAny().isPresent();
```
LosD: Here, the extra scoping is starting to do more harm than good. It would be quite a bit more readable if alreadyFiltered was declared and used where it is used.

LosD:
Code LGTM (except maybe the 50+ lines distance between alreadyFiltered declaration and usage, though no biggie).

I'll start testing now.

I might merge the test used to determine the issue in the first (if that's okay with @kaspersorensen).

LosD:
Just to be pedantic: Branch is misnamed, missing issue number :)

LosD:
Unfortunately, the fix seems to have no effect on the issue at all:
image

Confirmed by test case (See https://github.com/datacleaner/DataCleaner/commit/c348c1bbdf71024e650c2951f8df9cca55b99545)

LosD:
Description of issue seems a bit wrong, as far as my tests seemed to indicate: It wasn't a logical OR... I.e in the case above the "inherited" (equals) filter is simply not consulted at all when the preview max rows filter is attached to the whitespace transformer.

LosD:
(Or, not so much the test, as the effect is more or less indistinguishable with that few rows, but the debugging I did)

kaspersorensen:
I did add a unittest (for the Preview action) to reproduce. And before fixing that test showed the failure.

LosD:
Yeah, it's pretty weird, it seems to be exactly the same issue (F->T->T).

LosD:
Yeah, tried again from home... The programmaticly created test case succeeds, the one I tried (which is based on a job file) fails. Another difference is that the job-file based one has a convert-to-number in front... However, that is not the case for the manually tested one.

kaspersorensen:
I did this coding on the plane so I didn't have access to the issue - thus no way to name the branch properly :-)

I will run the test to confirm. Strange stuff. Lets see if we can nail it, but if nothing else this branch should improve the situation.

LosD:
Ah ☺

That's the weirdest part. In real usage, the change doesn't seem to do anything at all, so there's no improvement (but seemingly no downsides either)

kaspersorensen:
OK, a bit of a note here... Have been debugging this based on the unittest provided. It is in deed another scenario. But both scenarios are valuable IMO, so let's keep both to strengthen the preview even further.

I don't have time right now to finish it, but I have identified that for the whitespace trimmer, `AbstractRowProcessingConsumer.satisfiedOutcomesForConsume(...)` returns `true` (line 140) on records where it should return false due to the transitive filter condition. That's at least my initial feeling when I step through it.

kaspersorensen:
Or actually it might also be a few lines lower `AbstractRowProcessingConsumer.satisfiedInputsForConsume(...)` (line 144) where it goes wrong.

engine/core/src/main/java/org/datacleaner/job/runner/AbstractRowProcessingConsumer.java
```diff
@@ -121,7 +119,7 @@ private boolean isAlwaysSatisfiedRequirement() {
         final Set<Object> sourceJobsOfInputColumns = sourceColumnFinder.findAllSourceJobs(inputColumnSinkJob);
         for (Iterator<Object> it = sourceJobsOfInputColumns.iterator(); it.hasNext();) {
             final Object sourceJob = it.next();
-            if (sourceJob instanceof HasComponentRequirement) {
+            if (sourceJob instanceof HasComponentRequirement && sourceJob instanceof InputColumnSourceJob) {
```
kaspersorensen: For reference - this change was what made the integration test finally pass ... It seems that `buildSourceJobsOfInputColumns` was creating a list with any kind of source component. The naming indicates however that it's only sources of columns (so transformers).

kaspersorensen:
Updated the PR description in the top. And the build passed on my local machine - I expect Travis to pass too shortly :)

components/network-tools/src/test/java/org/datacleaner/extension/networktools/ResolveHostnameTransformerTest.java
```diff
@@ -19,38 +19,48 @@
...
+        assertEquals(null, t.transform(new MockInputRow().put(col, null))[0]);
+        
+        assertEquals(null, t.transform(new MockInputRow().put(col,
+                "lmdslfsm flskmf lskmfls kmslf kdmlfsk mflsk fmsl kfdmsl"))[0]);
+    }
+    
+    @Test
+    public void testTransformInternet() throws Exception {
+        if (!TestHelper.isInternetConnected()) {
+            return;
```
LosD: This is wrong-ish. You should use junit.Assume.assumeTrue(TestHelper.isInternetConnected()) instead.

Then the test gets properly skipped instead of claiming that the test passed, which we now nothing about.

kaspersorensen:
Fixed review remarks.

LosD:
LGTM! Merge when Travis is happy.

PR #1236, opened 2016-03-14
Validation in remote transformers

### LINKED ISSUE #1235: Component validation not working as expected for remote components
The NLDC remote component is hard to configure: it is unclear which fields need to be mapped for it to give output columns.   Only by trial and error one can discover this, which is not the user-friendliness we want to have.

This may be due to the component configuration validation not taking place, or the results of that validation not being put on screen.  Or because of validations lacking in the NLDC component.

**Expected Behavior**
User is guided to map the required fields for the component and gets immediate feedback if it's incorrectly configured.

Previewing or running a job with a misconfigured component is impossible.

**Observed behavior**
No feedback on incorrect config for NLDC

Attempt to preview leads to 500 internal server errors.

jakubneubauer:
fixes #1235
Added validation method into the remote transformer. The validation method tries to get OutputColumns from the server (which includes the validation part as well). The OutputColumns is cached, so subsequent call of "getOutputColumns" doesn't perform a query, so this enhancement has almost no performance impact.
We needed a cache implementation which is able to remember not only a value, but also an exception in case the value retrieval thrown an exception. I didn't find anything like that in Guava, so a simple error-aware cache implementation was introduced within this PR.


jakubneubauer:
fixes #1235

jakubneubauer:
Please, do not merge yet, I forgot that I want also do a little change in Remote components API on the server - return different HTTP response code in case of validation error (some 3xx - because it is the client side error, providing wrong input)

jakubneubauer:
Ready to merge

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -100,55 +105,28 @@ public void initClient() throws RemoteComponentException {
...
+    }
+
+    @Validate
+    public void validate() throws Exception {
+        CreateInput createInput = new CreateInput();
+        createInput.configuration = getConfiguration(getUsedInputColumns());
+        try {
+            cachedOutputColumns.getCachedValue(createInput);
+        } catch(RESTClientException e) {
+            if(e.getCode() == 422) {
```
LosD: 422 is a weird status code to use here. Protocol-wise there's nothing wrong with the request.

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
```diff
@@ -322,4 +300,39 @@ public void map(BatchSource<InputRow> source, BatchSink<Collection<Object[]>> si
         }
         convertOutputRows(out.rows, sink, size);
     }
+
+    private OutputColumns getOutputColumnsInternal(CreateInput createInput) throws Exception {
+        logger.debug("Getting output columns from server");
+        boolean wasInit = false;
```
LosD: I realize this is just moved code, but something like initializedLocally seem a little more proper.

LosD:
Code looks good, except the used status code confuses me a bit.

By the way, Guava Cache seems to support storing exceptions fine, just wrapped in an ExecutionException... However, AFAIK their abstract implementations are just skeletons, with the implementer providing both fetching _and_ storage, possible making it more work to make the concrete implementations than just using our own, so no objections from me.

LosD:
No wait, the Guave CacheBuilder built LoadingCache actually provides storage, as well as exception support... You just need to provide a CacheLoader.

LosD:
http://guava-libraries.googlecode.com/files/JavaCachingwithGuava.pdf

jakubneubauer:
@LosD , I want to return to the Guava debate. What I wanted achieve is to cache a value or exception. I mean really caching the exception, not just rethrowing it while retrieving value. And to make it more complicated ;-) I want to have different expiration time for the value (never) and the exception (1 second).

The reason behind it is that I want to prevent repetitive value retrieval even in case of error. I don't want to try to retrieve the value again if previous retrieval failed. It calls HTTP request so is pricy. But in case of error I want to give it a chance after 1 second in case it was temporary network failure.

jakubneubauer:
I couldn't find such a pattern in Guava caches portfolio.

LosD:
Hmmm... That might make sense, although never is a _very_ long time. :)

I'm pretty sure it is possible to wrap into the Guava cache but we're probably back in "more trouble than it's worth" territory :)

jakubneubauer:
"more trouble than it's worth" yeah :-) I considered some combination of more Guava's Suppliers, but I think we don't gain much code lines savings...

LosD:
Agreed :)

engine/remote-components/src/main/java/org/datacleaner/components/remote/RemoteTransformer.java
jakubneubauer: On the server side, we implemented returning 422 in case a validation fails in the transformer. I've recently read some suggestions in REST API design articles for cases when the request and also the payload is syntactically correct (if not, 400 is the candidate), but its content is not processable.

> The 422 (Unprocessable Entity) status code means the server understands the content type of the request entity (hence a 415(Unsupported Media Type) status code is inappropriate), and the syntax of the request entity is correct (thus a 400 (Bad Request) status code is inappropriate) but was unable to process the contained instructions.

LosD: Yeah, I've seen it before... But I don't really agree. It's processable, we just don't agree on the with the columns. Unprocessable to me is something like "a gdsfgsedt field? What's does that mean?" :)

Anyway, not really important, it just puzzled me.

kaspersorensen: Without a better alternative I vote to take the 422 that @jakubneubauer suggests :-)

jakubneubauer: To me it just seems like the best fit (looking on all http codes). What would you suggest instead, @LosD ?

LosD: Since RFC 7231, where the meaning of status codes were expanded, a normal (and admittedly bland) 400 seems to be the correct error code: https://tools.ietf.org/html/rfc7231#section-6.5.1

kaspersorensen:
Branch is no longer auto-mergeable. And I'm not sure if there is any review remarks still pending?

jakubneubauer:
If you don't have a really big problem with using the 422 status code for generic validation error, I would stick with it. I like that we can recognize it from the 400 returned on the "Tomcat" level (really mangled HTTP syntax etc.). If so, we can merge.

kaspersorensen:
Quick talk with @LosD and we agreed that we don't wanna be religious about it and that this is a very good start - we may always reconsider later.

Issue #170, opened 2015-01-20
Introduce a @Caution annotation to make user aware of components that should not be invoked without caution


kaspersorensen:
We have a couple of components that have side-effects or that may introduce costs for the user or so. At the same time we make it easy for users to preview their jobs, which is generally good, but may be unfortunate if the preview session triggers the side-effects.

For instance a transformer might do a remote method or web service call, and as such it would introduce potential side-effects and even privacy concerns.

Another example is that some component may write data to disk or some other destination without the user being aware of it.

So, to make the user aware we first need to have metadata about these components. I suggest we introduce an annotation like this:

```
public enum CautionType {
  REMOVE_SERVICE, PRIVACY, PAID_SERVICE, DATA_MODIFICATION
}

public @interface Caution {
  public CautionType[] value();
}
```

Not sure if the enum 'CautionType' is exchaustive yet, but those are a few types of cautions that I could come up with. The value() method of Caution is an array-type so that multiple types of cautions can be signaled.

kaspersorensen:
@losd this is the bomb

PR #1240, opened 2016-03-15
Bug/1239 smaller set in preview

### LINKED ISSUE #1239: Preview should work on a smaller set for remote components
We should reduce the preview size for DataCloud components to a much smaller number like 10.

rposkocil:
Fixes #1239

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/panels/TransformerComponentBuilderPanel.java
+++ b/desktop/api/src/main/java/org/datacleaner/panels/TransformerComponentBuilderPanel.java
@@ -19,6 +19,20 @@
  */
 package org.datacleaner.panels;
 
+import java.awt.BorderLayout;
+import java.awt.FlowLayout;
+import java.awt.Image;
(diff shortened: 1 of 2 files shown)


LosD:
Code LGTM, but is it possible to generalize this somehow without too much work? We have the same "costs money" issue with other transformers that can preview (admittedly not currently a lot, since the Satori services are implemented as analyzers), so it somehow seems wrong to make a special case of RemoteTransformer.

I'm thinking maybe a new annotation (`@DefaultPreviewSize(10)` or something like that).

It could also be something to be done for later.

kaspersorensen:
I think the idea you have @losd is described in a bit more general terms in #170 - so I just wanted to add that as a remark :)

kaspersorensen:
I'm merging this as a short term fix before the 5.0 release. I suggest to refactor it later while looking at the annotation API suggested in #170.

LosD:
:+1:

Issue #1243, opened 2016-03-16
The remote components do not have a tooltip with a description.


khouzvicka:
The remote components do not have a tooltip with a description. They should, because the tooltip explains the purpose of the component.

ClaudiaPHI:
Clousing becauce issue was fixed by https://github.com/datacleaner/DataCleaner/pull/1244

PR #1244, opened 2016-03-16
Refactor for all annotation in transformer class

khouzvicka:
#1243
Tooltip problem - remote transformers


jakubneubauer:
Looks good.

PR #1245, opened 2016-03-17
Bug/1237 confirmation of logging

### LINKED ISSUE #1237: No confirmation of successful login
When logging to DataCloud a feedback is expected. A creation of a confirmaion on welcome window and on More -> Options -> DataCloud tab.

rposkocil:
fixes #1237

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/options/DataCloudOptionsPanel.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/options/DataCloudOptionsPanel.java
@@ -22,18 +22,11 @@ package org.datacleaner.widgets.options;
 import static org.datacleaner.descriptors.RemoteDescriptorProvider.DATACLOUD_SERVER_NAME;
 import static org.datacleaner.descriptors.RemoteDescriptorProvider.DATACLOUD_URL;
 
-import java.awt.Color;
-import java.awt.Component;
-import java.awt.GridBagConstraints;
+import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 
-import javax.swing.JButton;
-import javax.swing.JComponent;
-import javax.swing.JEditorPane;
-import javax.swing.JLabel;
-import javax.swing.JPasswordField;
-import javax.swing.JTextField;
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/java/org/datacleaner/widgets/options/DataCloudOptionsPanel.java
```diff
@@ -160,11 +162,15 @@ private boolean updateDcConfiguration() {
         try {
             RemoteServersUtils.checkServerWithCredentials(DATACLOUD_URL, username, password);
         } catch (Exception ex) {
+            invalidCredentialsLabel.setForeground(new Color(170,10,10));
```
LosD: How come this doesn't use the normal color scheme in WidgetUtils (the old apparently didn't either, so it may not be anything new)?

desktop/ui/src/main/java/org/datacleaner/widgets/options/DataCloudOptionsPanel.java
```diff
@@ -160,11 +162,15 @@ private boolean updateDcConfiguration() {
...
             invalidCredentialsLabel.setText("Sign in to DataCloud failed: " + ex.getMessage());
+            logger.warn("Sign in to DataCloud failed for user '{}'", username, ex);
             return false;
         }
 
-        invalidCredentialsLabel.setText("&nbsp;<br>&nbsp;");
+        invalidCredentialsLabel.setForeground(new Color(10,170,10));
```
LosD: Same here

LosD:
Code looks good, though I'd like a little better use of our styling class (maybe adding the new colors to the class, if the current color scheme doesn't apply). However, this is apparently not new, so not a blocker.

Test went fine, but I don't really like that the user has to click an OS element to dismiss the dialog. I'd prefer it to auto-dismiss after a few seconds, but at the very least a close button should be available.

LosD:
We agreed to skip the timeout. It wont matter much (and might even confuse), and there is quite some time pressure at the moment.

LosD:
Aaaaaaand my eyes apparently doesn't work. The close button is right there in the lower right corner.

Merging! :)

PR #1246, opened 2016-03-17
Lower timeout for HdfsUrlChooser

### LINKED ISSUE #1150: HdfsUrlChooser does not time out when trying to load from an unresponsive machine
Whenever a user tries to browse an unresponsive Hadoop cluster/namenode, the timeout will take forever (I'm not sure if that is literal or not, at least much, much longer than reasonable).

What makes it worse, is that there's no way to cancel the wait, so the DC Window is essentially useless while waiting.

Try to make a timeout on the connection, so that the user doesn't have to wait forever. This probably shouldn't apply to the actual execution of jobs, as we don't want to kill those unless the server is really, truly dead. But maybe make sure that the cancel button works also when dealing with an unresponsive server.

arjansh:
Fixes #1150.

Changed the max number of retries for trying to connect with an HDFS to 1 (where the default is 45) when selecting a File from an HDFS and refactored the code a bit, so no more attempts to retrieve files from the HDFS are done if the scanning of the Hadoop config files isn't successful.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/util/HdfsUtils.java
+++ b/desktop/ui/src/main/java/org/datacleaner/util/HdfsUtils.java
@@ -25,13 +25,15 @@ import java.net.URI;
 import javax.ws.rs.core.UriBuilder;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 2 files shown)


kaspersorensen:
This only seems to work correctly the first time around, but NOT if I put a bad hostname in the popup dialog that shows after the first attempt. Then it just hangs:

image

I think you may need to set this property somewhere else also? I think in `HdfsUtils.getHadoopConfiguration(...)` although that may be used for (too) many other things.

kaspersorensen:
LGTM.

PR #1248, opened 2016-03-18
Catch errors in component close methods

### LINKED ISSUE #1247: CloseTaskListener should handle exceptions thrown by @Close
CloseTaskListener doesn't handle exceptions thrown by the cleanup action, which prevents it from signaling to the JoinTaskListener that it has terminated. This is causing the execution to hang indefinitely when an upstream multi-stream component throws an exception in `@Close`.

The issue is easy to reproduce by implementing a `@Close` method that throws a RuntimeException in `MockOutputDataStreamAnalyzer`, and running the `MultipleChainsOfOutputDataStreamTest` test case.

LosD:
To avoid hangs, closing tasks must always continue, even with an error in a component's `@Close`-annotated methods.
We may want to do more elaborate reporting, but for now an error is just logged.

Fixes #1247

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
+++ b/engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
@@ -90,6 +90,7 @@ public class CloseTaskListener implements TaskListener {
             cleanup();
         } catch (Exception e) {
             onErrorInternal(task, e, false);
+            return;
         }
         if (_nextTaskListener != null) {
             _nextTaskListener.onComplete(task);
@@ -102,6 +103,8 @@ public class CloseTaskListener implements TaskListener {
     }
 
     private void onErrorInternal(Task task, Throwable throwable, boolean doCleanup) {
+        final boolean previouslySuccessful = _success.getAndSet(false);
+
         if (doCleanup) {
             try {
                 cleanup();
@@ -110,7 +113,6 @@ public class CloseTaskListener implements TaskListener {
             }
         }
 
-        final boolean previouslySuccessful = _success.getAndSet(false);
         if (previouslySuccessful) {
             // only report the first such error
             _analysisListener.errorUnknown(_analysisJob, throwable);
--- a/engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
+++ b/engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
@@ -19,7 +19,6 @@
  */
 package org.datacleaner.job.tasks;
 
-import java.util.Collection;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.datacleaner.descriptors.ComponentDescriptor;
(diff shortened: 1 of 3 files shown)


LosD:
@edwinchoi Could I get you to verify that this fixes the problem for you (thanks a million for the well-analyzed report by the way)?

edwinchoi:
Thanks for the quick turn-around.

To be honest, I would prefer that the execution hang on error over swalling of the errors. Now it seems that an exception in `@Close` won't propagate downstream or upstream.

What if you did something like this instead...
```java
public void onComplete(Task task) {
    try {
        cleanup();
    } catch(Exception e) {
        failed(task, e, false);
        return;
    }
    ...
}
public void onError(Task task, Throwable cause) {
    failed(task, cause, true);
}
private void failed(Task task, Throwable cause, boolean releaseResources) {
    // notify listener
    ...
    if (releaseResources) {
        try {
            cleanup();
        } catch(Exception e) {
            cause.addSuppressed(e);
        }
    }
    ...
}
```

This would ensure the task properly sends the termination signal downstream, while maintaining the error condition.

LosD:
I'm going to have to consider this. I'm no fan of swallowing errors either, but OTOH, I don't really like starting to call a succesful job a failure, even less halfway through a chain of listeners.

LosD:
Ok, there's a real good reason to do it your way: OutputDataStream jobs may be flushing here, meaning that a failure could be a real failure.

edwinchoi:
It's not really limited to that either. The way I see it, `@Close` is the complement of `@Initialize`. If a job fails to flush out the state changes its made in `@Initialize` on termination (whether it be releasing of resources or whatever), it should be assumed that the system is in an inconsistent state and that it is unsafe to proceed.

LosD:
Implemented the recommended change, with a few modifications (i.e. adding the suppressed exceptions before notifying, so the listener has a chance of handling them).

Thanks again! :)

engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
```diff
@@ -90,7 +86,11 @@ public void onBegin(Task task) {
 
     @Override
     public void onComplete(Task task) {
-        cleanup();
+        try {
+            cleanup();
+        } catch (Exception e) {
+            onErrorInternal(task, e, false);
```
kaspersorensen: I think we should return here?

LosD: Yep

engine/core/src/main/java/org/datacleaner/job/tasks/CloseTaskListener.java
```diff
@@ -98,12 +98,23 @@ public void onComplete(Task task) {
...
+    private void onErrorInternal(Task task, Throwable throwable, boolean doCleanup) {
+        if (doCleanup) {
+            try {
+                cleanup();
+            } catch (Exception e) {
+                throwable.addSuppressed(e);
+            }
+        }
+
         final boolean previouslySuccessful = _success.getAndSet(false);
```
LosD: Stupid mistake here. That flag needs to be set before calling cleanup();

kaspersorensen:
LGTM. What do you think @edwinchoi ? If you agree we will merge.

kaspersorensen:
Merging, please let us know if you found any omissions @edwinchoi :-)

edwinchoi:
Look goods, thanks!

PR #1249, opened 2016-03-21
Add new partition index to Spark component key

### LINKED ISSUE #1242: Spark submit to Hadoop merges analysis results for Value Distribution Analyzer
When you create a job with a Value Distribution Analyzer with more than one column selected in the input columns, when running the job normally the analysis results shows result tabs for each selected column. If you however use spark submit to run the job on a Hadoop cluster, it seems to merge the analyzed values of all columns into on set an renders the one result tab for all values in all input columns (when you download the result file from the hdfs and open it in DataCleaner). Something similar happens when you use the Pattern Finder Analyzer.

LosD:
To avoid escalated component jobs being sent to a single reducer, a new partition index is introduced, which makes it possible to see that the partitions are not the same job.

Fixes #1242

### DIFF
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
@@ -73,7 +73,6 @@ public class SparkJobContext implements Serializable {
     // cached/transient state
     private transient DataCleanerConfiguration _dataCleanerConfiguration;
     private transient AnalysisJobBuilder _analysisJobBuilder;
-    private AnalysisJobBuilder _analysisJobBuilder1;
 
     public SparkJobContext(final URI dataCleanerConfigurationPath, final URI analysisJobXmlPath,
             final URI customPropertiesPath, final JavaSparkContext sparkContext) {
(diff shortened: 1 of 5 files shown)


engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
```diff
@@ -72,6 +73,7 @@
     // cached/transient state
     private transient DataCleanerConfiguration _dataCleanerConfiguration;
     private transient AnalysisJobBuilder _analysisJobBuilder;
+    private AnalysisJobBuilder _analysisJobBuilder1;
```
kaspersorensen: This looks like it is unused ... if only your IDE would tell you ... :-1:

kaspersorensen:
Just one remark, otherwise LGTM.

engine/env/spark/src/main/java/org/datacleaner/spark/SparkJobContext.java
LosD: It will... It even warns when committing, but it often drowns in other warnings from to silly coding practices and bad IDEs :p

PR #1250, opened 2016-03-21
Added multiple possible 'primary' spark JAR filename prefixes

kaspersorensen:
Allows the "primary" JAR of spark to be prefixed with _either_ `DataCleaner-env-spark` or `DataCleaner-spark`

### DIFF
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/ApplicationDriver.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/ApplicationDriver.java
@@ -218,6 +218,7 @@ public class ApplicationDriver {
                 for (String prefix : PRIMARY_JAR_FILENAME_PREFIXES) {
                     if (filename.startsWith(prefix)) {
                         primaryJarRef.set(path.toString());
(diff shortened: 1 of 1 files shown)


engine/env/spark/src/main/java/org/datacleaner/spark/ApplicationDriver.java
```diff
@@ -214,16 +214,21 @@ private String toHadoopPath(String path) {
                 final LocatedFileStatus file = files.next();
                 final Path path = file.getPath();
                 final String filename = path.getName();
-                if (filename.startsWith(PRIMARY_JAR_FILENAME_PREFIX)) {
-                    primaryJarRef.set(path.toString());
-                } else {
+                boolean primaryJar = false;
```
LosD: This doesn't seem to ever be set to anything else?

LosD: (... if only your IDE would tell you ;))

LosD:
LGTM. Merge when Travis is happy

PR #1252, opened 2016-03-21
Feature/1140 add dialog for choosing server

ClaudiaPHI:
There are some things left to do: 

- [x] There may be some clean up 
- [x]  Save the new configurations (serverInformationCatalog) on disk (it doesn't automatically save it when elements are removed ) 
- [x]  In the select configuration dialog I need to add a link towards the "Hadoop options".  

Feel free to test and suggest any improvements for wording, etc.


desktop/ui/src/main/java/org/datacleaner/windows/HadoopConfigurationsOptionsDialog.java
```diff
@@ -0,0 +1,282 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    private final ImageManager imageManager = ImageManager.get();
+    private final UserPreferences _userPreferences;
+    private final DataCleanerConfiguration _configuration;
+    private final MutableServerInformationCatalog _serverInformationCatalog;
+    private DCPanel _connectionsConfigurationsPanel;
+    private final WindowContext _windowContext;
+    private final List<HadoopConnectionPanel> _connectionsPanels;
```
LosD: Very minor, but I think connectionPanels makes more sense.

desktop/ui/src/main/java/org/datacleaner/panels/AddDatastorePanel.java
```diff
@@ -68,7 +69,7 @@ public AddDatastorePanel(final DatastoreCatalog datastoreCatalog,
             final ServerInformationCatalog serverInformationCatalog, final DatabaseDriverCatalog databaseDriverCatalog,
             final DCModule dcModule,
             final DatastoreSelectedListener datastoreSelectedListener, UserPreferences userPreferences,
-            boolean showExistingDatastoresButton) {
+            boolean showExistingDatastoresButton, WindowContext wizardContext) {
```
LosD: Since you have the dcModule here, does it really make sense to propagate the WindowContext (Which you call wizardContext for some reason... A DI habit, maybe? :smile:) all the way up? You could just create an injector here instead.

desktop/ui/src/main/java/org/datacleaner/windows/HadoopConnectionToNamenodeDialog.java
```diff
@@ -0,0 +1,282 @@
...
+    public HadoopConnectionToNamenodeDialog(WindowContext windowContext,
+            DirectConnectionHadoopClusterInformation directConnection,
+            MutableServerInformationCatalog serverinformationCatalog) {
+        super(windowContext);
+
+        _statusLabel = DCLabel.bright("Please specify connection name");
+        _directConnection = directConnection;
+        _nameTextField = WidgetFactory.createTextField("MyConnection");
+        _hostTextField = WidgetFactory.createTextField("localhost");
+        _portTextField = WidgetFactory.createTextField("9000");
```
LosD: I know that we use 9000 on our test server, but the correct default namenode port is 8020.

desktop/ui/src/main/java/org/datacleaner/windows/HadoopConnectionToNamenodeDialog.java
```diff
@@ -0,0 +1,282 @@
...
+
+        final String saveButtonText = directConnection == null ? "Register connection" : "Save connection";
+        _saveButton = WidgetFactory.createPrimaryButton(saveButtonText, IconUtils.ACTION_SAVE_BRIGHT);
+        _saveButton.addActionListener(new ActionListener() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+
+                URI nameNodeUri;
+                try {
+                    nameNodeUri = new URI("hdfs", null, _hostTextField.getText(), Integer.parseInt(_portTextField
```
LosD: I'm not sure that hdfs is always a given. Maybe consider adding a scheme field with a default of HDFS? I think we have a (probably non-exhaustive) list of possible Hadoop schemes somewhere.

desktop/ui/src/main/java/org/datacleaner/windows/HadoopConnectionToNamenodeDialog.java
```diff
@@ -0,0 +1,282 @@
...
+        });
+
+        _portTextField.getDocument().addDocumentListener(new DCDocumentListener() {
+            @Override
+            protected void onChange(DocumentEvent event) {
+                validateAndUpdate();
+            }
+        });
+
+        _descriptionTextField.getDocument().addDocumentListener(new DCDocumentListener() {
```
LosD: You should probably reuse a singe DCDocumentListener instance instead of creating multiple ones.

desktop/ui/src/main/java/org/datacleaner/panels/HadoopDirectoryConfigurationDialog.java
```diff
@@ -0,0 +1,348 @@
...
+                final List<String> paths = new ArrayList<>();
+                for (int i = 0; i < _pathPanels.size(); i++) {
+                    final DirectoryPathPanel directoryPathPanel = _pathPanels.get(i);
+                    final File directory = directoryPathPanel.getDirectory();
+                    if (directory != null && directory.isDirectory()) {
+                        paths.add(directory.getPath());
+                    }
+                }
+                // We do not save a connection if there are no paths
+                if (paths.size() == 0) {
```
LosD: Shouldn't this validation be done in `validateAndUpdate()`?

LosD:
Minor things from me, but looks good so far!

One thing that I'd like to see is validation that the paths exists in the `HadoopDirectoryConfigurationDialog` and contains configuration files. I think a simple way to do that is simply to try to get a configuration from the instantiated `DirectoryBasedHadoopClusterInformation`. 

Later, we might also want to validate that an actual connection can be established, but that's probably too much for now.

desktop/ui/src/main/java/org/datacleaner/windows/HadoopConnectionToNamenodeDialog.java
ClaudiaPHI: I will do that.

ClaudiaPHI:
image


image


image


image


image


coloredadddatabsedriverbutton

ClaudiaPHI:
What it doesn't work:  
-[ ] Removing configuration from conf.xml.  
-[ ] Someone might wanna work on the wording

LosD:
Haven't looked at the new code, but the screenshots looks good now! :)

Some small things:
- I think the configuration dialog is so simple now that it could rather easily fit into the options dialog. Especially when it isn't loaded with all the test values.
- (This may make the other point less valid again) For the overview, it would be nice with a bit of hints what type a configuration is. This is currently a major shortcoming of this way compared to the other one.
- The `(default)` is pretty nasty. It should look nice and friendly, not weird. Especially since in most cases it will be the only one used.

kaspersorensen:
Update: I've given this PR/branch a good massage and I think it's now ready for review. Some screenshots:

image

image

ClaudiaPHI:
Ready to merge  :-D

PR #1254, opened 2016-03-22
Add a listener interface to SparkRunner

### LINKED ISSUE #1253: Support progress listener in SparkRunner
To track the progress that happen internally in `SparkRunner` (at minimum upload of files, and submitting the job), some sort of listener interface is needed.

LosD:
To support the listener interface, a launch method that returns the Process instead of waiting for it has been added.

Fixes #1253


kaspersorensen:
LGTM, but shouldn't we also have some event when the job finishes (so when `waitFor()` returns)?

LosD:
We could, but it isn't really needed. That the method returns should be a pretty good indicator :)

LosD:
After discussion, it was agreed to move some concerns about the role or `SparkRunner`and `ApplicationDriver` to #1255, and keep this a bit dirty (but working) :)

LosD:
Merging

PR #1256, opened 2016-03-29
Preventing "ordering" in MaxRowsFilter when "Running on N rows"

kaspersorensen:
A small (but important) fix to the "Preview N rows" functionality on the execute button. This ensures that ordering is not applied by the "Max rows" filter - which means that on the query execution plan, nothing is done to apply an "ORDER BY" clause.

(I just ran out of memory due to this - because I was previewing a job based on a very big CSV file ... DOH!)


ClaudiaPHI:
LGTM!

PR #1261, opened 2016-03-30
Utility class for checking the validity of a job ran on Hadoop

ClaudiaPHI:
Utility class for checking the validity of a job ran on Hadoop

### DIFF
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopConfigurationUtils.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopConfigurationUtils.java
@@ -1,0 +1,63 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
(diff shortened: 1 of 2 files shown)


ClaudiaPHI:
Do not merge!

ClaudiaPHI:
Ready for review and merge

engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopConfigurationUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import org.datacleaner.util.StringUtils;
+
+/**
+ * Utility class to check the validity of a datastore when trying to run job on
+ * Hadoop with Spark.
+ *
+ */
+public class HadoopConfigurationUtils {
+
+    public static boolean isValidConfiguration(Datastore datastore) {
```
kaspersorensen: Since the method takes a datastore and has some assumptions about what it is used for (the source of a job) then I would rename the method to: `isValidSourceDatastore(Datastore)`.

kaspersorensen:
Something went bad in Travis it seems.

And a small naming thingie, other LGTM.

PR #1264, opened 2016-03-30
Issue #1263: Added NULL and BLANK_OR_NULL conditions to Completeness

### LINKED ISSUE #1263: Completeness analyzer: Ability to set that NULL is required for particular fields
Just encountered an interesting case today, and shortcoming of our existing Completeness analyzer. A customer wants to ensure that (given certain filter conditions) certain fields are filled AND that certain other fields are NULL. The first part she can do with the Completeness analyzer, but NOT the part about ensuring that other fields are NULL.

It actually seems like such a natural requirement and something we should add.

kaspersorensen:
Fixes #1263

### DIFF
--- a/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
+++ b/components/basic-analyzers/src/main/java/org/datacleaner/beans/CompletenessAnalyzer.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


arjansh:
I would prefer it if the added logic in the switch statement was put in a separate "validate" method (which might even be added to the Condition enum), so it doesn't increased the complexity of the run method.

Outside of that (and outside of the failing travis build) it looks good to me.

kaspersorensen:
Ok let me give it a massage :)

kaspersorensen:
Please re-review.

arjansh:
LGTM! Merge freely if Travis approves.

PR #1278, opened 2016-04-04
Black remote indicator

### LINKED ISSUE #1277: Repaint remote indicator to black color
As a user I like to use the remote components. It's simple and easy I am not even aware that they are working remote. Even the icons on these components are not immediately refllecting that. Theyu do it in a subtle way by providing a wifi element on the icon. That element is in green and hardly visible on green and blue colored icons. If we make it black its more visible that you are using remote components ...

khouzvicka:
Fixes #1277

New black color:

new remotewifi


khouzvicka:
Please wait with merge. I have to check this story with Manuel. Maybe it is obsolete

manuelvandenberg:
I think it is not necessary anymore to change to black. The green is equally well visible in the new color scheme and with the new positioning of the overlay.

@LosD, if you agree then let's close this PR without merging.

LosD:
Fine by me

PR #1280, opened 2016-04-05
Fix reference to Salesforce duplicate detection job

### LINKED ISSUE #1269: Salesforce duplicate detection demo job is missing
There is a `Salesforce duplicate detection.analysis.xml` job in `desktop/ui/src/main/resources/datacleaner-home/jobs`. This job is however not available in the example jobs which are created in your user's .datacleaner folder after DataCleaner is started. This is most probably caused by the fact the the `org.datacleaner.user.DemoConfiguration` doesn't have a reference to this file, while it does have a reference to a file named `Salesforce dedup training.analysis.xml` which doesn't exist on disk, so there probably has been some kind of mixup here.

arjansh:
Fixes #1269.

Replaced reference to non existent "Salesforce dedup training" job with one to the existent "Salesforce duplicate detection" job.


ClaudiaPHI:
LGTM!

PR #1286, opened 2016-04-07
Fix errors caused by missing .png files

### LINKED ISSUE #1285: Removed png files may cause errors
The removal of the images/actions/add.png, images/actions/close.png, images/actions/remove.png, images/actions/save.png and images/actions/scrolldown.png from the datacleaner-desktop-ui module (e.g. in the context of https://github.com/datacleaner/DataCleaner/pull/1252) can result in errors in the DataCleaner desktop ui when using components which have been compiled against a version of DataCleaner which still contained these png files.

arjansh:
Fixes #1285.

Copy images/actions/*_dark.png to images/actions/*.png to make sure components compiled against older versions which use the .png files still work at run-time.

### DIFF
--- a/desktop/ui/pom.xml
+++ b/desktop/ui/pom.xml
@@ -58,6 +58,35 @@
 					<source>${project.build.directory}/lib</source>
 				</configuration>
 			</plugin>
+
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


LosD:
Ugh! Ant!

No it's fine. Good simple solution. :)

Issue #1287, opened 2016-04-07
Scheduled task runner and refactor of remote server status.


khouzvicka:
I would like to add new funcionality to MultiThreadedTaskRunner. 
Scheduled tasks. There will be one thread to run scheduled tasks.

And also refactor of get method for remote server status.

ClaudiaPHI:
Closing. Fixed by https://github.com/datacleaner/DataCleaner/pull/1288

PR #1288, opened 2016-04-07
Scheduled tasks. There will be one thread to run scheduled tasks. And…

khouzvicka:
#1287   Scheduled tasks. There will be one thread to run scheduled tasks. And also refactor of get method for remote server status.


khouzvicka:
Any suggestion? @kaspersorensen @LosD

LosD:
I really like to have this (it actually came up recently here as well), but can't the scheduled threads use the same ThreadPoolExecutor (i.e. replace the ThreadPoolExecutor with a ScheduledThreadPoolExecutor, it should be capable of both) as the normally executed threads?

LosD:
Furthermore, this PR seems to do A LOT more than just providing this functionality. That really should go into its own PR (with a proper description on what it does, and what problem it solves).

engine/remote-components/src/main/java/org/datacleaner/descriptors/RemoteDescriptorProviderImpl.java
```diff
@@ -266,16 +219,13 @@ private void downloadDescriptors() {
         return annotations;
     }
 
-    @Override
-    public Collection<DescriptorProviderStatus> getStatus() {
-        final Set<DescriptorProviderStatus> statusSet = new HashSet<>();
+    private class RemoteServerStateListenerImpl implements org.datacleaner.configuration.RemoteServerStateListener{
```
LosD: org.datacleaner.configuration.RemoteServerStateListener is already imported

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
```diff
@@ -19,20 +19,51 @@
...
+import org.slf4j.LoggerFactory;
 
 /**
  * Implementation of {@link RemoteServerConfiguration}.
  */
 public class RemoteServerConfigurationImpl implements RemoteServerConfiguration {
 
+    private static final Logger logger = LoggerFactory.getLogger(RemoteServerConfigurationImpl.class);
+    private static final int TEST_CONNECTION_TIMEOUT = 15 * 1000; // [ms]
+    private static final long DELAY_MIN = 5;
```
LosD: Is the check expensive? 5 minutes seems very slow to reconnect after a loss, or will that be taken care of another way?

Naming-wise, it would be nice with a little more precise name, this instantly makes me think "delay for what?". Maybe also make a constant for the timeunit.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
```diff
@@ -56,5 +87,103 @@ public RemoteServerData getServerConfig(String serverName) {
...
+        }
+        listeners.add(listener);
+    }
+
+    @Override
+    public void removeListener(RemoteServerStateListener listener) {
+        listeners.remove(listener);
+    }
+
+    private synchronized void scheduledTask() {
```
LosD: This seems to schdule a task, so it's scheduleTask not scheduledTask. And it would again be nice with a little more precise name, i.e. scheduleServerStatusTask. Not super important for private methods, but it still makes the code a little more readable, and everything helps :).

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
```diff
@@ -56,5 +87,103 @@ public RemoteServerData getServerConfig(String serverName) {
         return null;
     }
 
+    @Override
+    public RemoteServerState getActualState(String remoteServerName) {
+        if (serverStatusTask == null) {
```
LosD: `scheduledTask()` makes this check, so I guess there's no need to repeat it (same below)

jakubneubauer:
Overall architecture seems ok for me: ServerData holds connection data for a remote server, ServerConfiguration uses it and builds a regular checker for remote servers, RemoteDescriptorProvider provides components for specific server and uses the status for serverConfiguration to provide info about "health" of its components.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
jakubneubauer: What about different check interval for health server and for error state?

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
khouzvicka: There is two ways, how use it this class.
Active getting status and add Listener for future notification.
This two actions have create/check task.
If you will call only actual status again again. The task must be created.

khouzvicka:
@LosD 
> I really like to have this (it actually came up recently here as well), but can't the scheduled threads use the same ThreadPoolExecutor (i.e. replace the ThreadPoolExecutor with a ScheduledThreadPoolExecutor, it should be capable of both) as the normally executed threads?
> 

My vision is one dedicated pool with one thread for service tasks. I mean that is better for standard performance of threads.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
LosD: That would certainly also make sense, though it may make it hard to get back up if the server gets pounded when it is already unstable.

- But that is probably not a worry until we have a lot more customers, and then we'll probably need to change the architecture to better support scale-out anyway.

LosD:
That would also mean that service tasks would have to wait for each other... I'd much rather have them completely independent of each other, especially since the main CPU pressure will come from job executions anyway.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
```diff
@@ -19,20 +19,51 @@
...
 
-    public RemoteServerConfigurationImpl(List<RemoteServerData> serverData) {
+    public RemoteServerConfigurationImpl(List<RemoteServerData> serverData, TaskRunner taskRunner) {
         remoteServerDataList = new ArrayList<>(serverData);
+        for (RemoteServerData remoteServerData : serverData) {
+            actualStateMap.put(remoteServerData.getServerName(), RemoteServerState.UNKNOWN);
+        }
+
+        if (taskRunner == null || !(taskRunner instanceof ScheduledTaskRunner)) {
+            logger.error("Task runner isn't ScheduledTaskRunner. Remote server status task won't be scheduled.");
```
LosD: Shouldn't this throw instead of just logging? Then you can also make scheduledTaskRunner final and drop the check in scheduledTask()

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
LosD: I mean, there's a check for `if (serverStatusTask == null)` in `scheduledTask()` (along with a check for the runner), so you don't need to check it here as well, you can just call scheduledTask() directly.

Not that it's much of an issue.

khouzvicka:
I added dynamic intervals for check.
Every 1 minute - only servers with connection problems.
Every 5 minutes - all servers.

jakubneubauer:
I have bad news. Quick look on java ScheduledThreadPoolExecutor class shows, that it doesn't allow to use custom task BlockingQueue, which is needed for the normal (non-scheduled) tasks -- see to comments in the MultiThreadedTaskRunner constructor and its inner class AlwaysBlockingQueue. Thus, we cannot use java ScheduledThreadPoolExecutor to run both scheduled and unscheduled tasks through one thread pool. We have to use 2 pools - one ThreadPoolExecutor instance for normal tasks and one ScheduledThreadPoolExecutor for the scheduled tasks.

khouzvicka:
@arjansh Could you do code review, please?

LosD:
Could it reuse the work queue?

arjansh:
@khouzvicka I will do a code review.

api/src/main/java/org/datacleaner/descriptors/DescriptorProvider.java
```diff
@@ -46,12 +46,6 @@
 public interface DescriptorProvider {
 
     /**
-     * @return a collection of {@link DescriptorProviderStatus} describing
-     *         current status.
-     */
-    public Collection<DescriptorProviderStatus> getStatus();
-
-    /**
```
arjansh: This pull request effectively removes all usages of the `DescriptorProviderStatus` interface, why not remove that interface as well?

api/src/main/java/org/datacleaner/job/concurrent/ScheduledTaskRunner.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.datacleaner.job.tasks.Task;
+
+/**
+ * Extended TaskRunner for scheduled tasks.
+ */
+public interface ScheduledTaskRunner extends TaskRunner {
+
+    /**
+     * Submits a {@link Task} and a {@link ScheduledTaskRunner} to the
```
arjansh: Should this say `{@link TaskListener}` instead of `{@link ScheduledTaskRunner}`?

desktop/ui/src/main/java/org/datacleaner/guice/DCModuleImpl.java
```diff
@@ -300,7 +300,10 @@ public final DataCleanerConfiguration getDataCleanerConfiguration(@Undecorated D
                     final StorageProvider storageProvider = c.getEnvironment().getStorageProvider();
 
                     final TaskRunner taskRunner = c.getEnvironment().getTaskRunner();
-                    final RemoteServerConfiguration remoteServerConfiguration = new MutableRemoteServerConfigurationImpl(c.getEnvironment().getRemoteServerConfiguration().getServerList(), configurationWriter);
+                    final RemoteServerConfiguration remoteServerConfiguration =
+                            new MutableRemoteServerConfigurationImpl(
+                                    c.getEnvironment().getRemoteServerConfiguration().getServerList(),
+                                    c.getEnvironment().getTaskRunner(), configurationWriter);
```
arjansh: Why do `c.getEnvironment().getTaskRunner()`, when in the line above (302), it is already retrieved and assigned to a variable?

api/src/main/java/org/datacleaner/descriptors/DescriptorProvider.java
LosD: That's a nope. This is part of the public API, and we don't remove public API methods without warning.

arjansh: Then this method also shouldn't be removed.

LosD: That's what I meant :)

arjansh: Could both be deprecated instead then?

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
```diff
@@ -19,20 +19,54 @@
...
 public class RemoteServerConfigurationImpl implements RemoteServerConfiguration {
 
+    private static final Logger logger = LoggerFactory.getLogger(RemoteServerConfigurationImpl.class);
+    private static final int TEST_CONNECTION_TIMEOUT = 15 * 1000; // [ms]
+    private static final long ERROR_DELAY_MIN = 1;
+    private static final long OK_DELAY_MIN = 5;
+    private final Map<String, RemoteServerState> actualStateMap = Collections.synchronizedMap(new HashMap<>());
+    private ServerStatusTask serverStatusTask;
+    private ScheduledTaskRunner scheduledTaskRunner;
+    private List<RemoteServerStateListener> listeners = Collections.synchronizedList(new ArrayList());
```
arjansh: My IDE complains about a `Type safety: The expression of type List needs unchecked conversion to conform to List<RemoteServerStateListener>` here. Maybe add <> to the new ArrayList() object instanciation.

engine/core/src/main/java/org/datacleaner/configuration/RemoteServerConfigurationImpl.java
```diff
@@ -19,20 +19,54 @@
...
 
-    public RemoteServerConfigurationImpl(List<RemoteServerData> serverData) {
+    public RemoteServerConfigurationImpl(List<RemoteServerData> serverData, TaskRunner taskRunner) {
         remoteServerDataList = new ArrayList<>(serverData);
+        for (RemoteServerData remoteServerData : serverData) {
+            actualStateMap.put(remoteServerData.getServerName(), RemoteServerState.UNKNOWN);
+        }
+
+        if (taskRunner == null || !(taskRunner instanceof ScheduledTaskRunner)) {
+            logger.warn("Task runner isn't ScheduledTaskRunner. Remote server status task won't be scheduled.");
```
arjansh: Maybe log this at info level, because now a warning logged each time I start my vanilla DataCleaner install.

api/src/main/java/org/datacleaner/descriptors/DescriptorProvider.java
khouzvicka: @LosD @arjansh This method was relative new, we used it only for remote components. This is reason, why I could delete it. But if you wish, I will return it back with inotation.

LosD: Okay, let's take it out. It's not so nice, but you're right that that the chance anyone us using it is _very_ slim. Then let's delete `DescriptorProviderStatus` as well, then.

arjansh:
From my point of view I have no further comments.

LosD:
I'm no fan of the single threaded scheduled queue, but let's leave it at that for now. If/when more services starts using it, we'll probably have to use more threads to make sure scheduled tasks happen roughly when wanted.

jakubneubauer:
@LosD , don't be afraid, the settings for ScheduledThreadPoolExecutor is:

- core pool size = 1, but
- max pool size = 50 (with the last commit we changed it to 50, default was Integer.MAX_VALUE, which seems danger to me, so we changed it)
- keep-alive time = 0

It means that in peak season new threads are created to satisfy the demand of scheduled tasks (up to maximum 50), but immediately (because keep alive time = 0) after their finish the threads are closed and only one thread is kept (core size = 1)

LosD:
Oh, I remembered it as a `newSingleThreadScheduledExecutor`, but this is fine :)

PR #1289, opened 2016-04-07
Fix backwards compatibility issue in plain search/replace transformer

### LINKED ISSUE #1284: Plain search/replace transformer is not backwards compatible
In the context of https://github.com/datacleaner/DataCleaner/pull/1149 the `Plain search/replace transformer` has been refactored and jobs configured to use that transformer configured in DataCleaner versions older than 5.0 can no longer be loaded in DataCleaner 5.0. If you try to load such a job, you get a `No such property: Search string` error.

You should however be able to load the job without problems and it should fix/upgrade its own configuration to comply with its 5.0 variant.

arjansh:
Fixes #1284.

Added some logic to the JaxbJobReader#applyProperties method which checks if a property has been removed from the component which is being read, stores these values and uses these to process the removed properties, i.e. assign their values to another newly introduced property if necessary.

Implemented this logic specifically for the PlainSearchReplaceTransformer from which the "Search string" and "Replacement string" properties have been removed in favor of a "Replacements" property.

### DIFF
--- a/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
+++ b/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
@@ -48,6 +48,7 @@ import org.apache.metamodel.util.FileHelper;
 import org.datacleaner.api.Converter;
 import org.datacleaner.api.InputColumn;
 import org.datacleaner.api.OutputDataStream;
+import org.datacleaner.beans.transform.PlainSearchReplaceTransformer;
 import org.datacleaner.configuration.DataCleanerConfiguration;
 import org.datacleaner.configuration.SourceColumnMapping;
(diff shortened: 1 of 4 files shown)


LosD:
Code looks good, but a quick test loading an old job would be nice.

engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
```diff
@@ -878,38 +879,49 @@ private void applyProperties(final ComponentBuilder builder,
         if (configuredPropertiesType != null) {
             final List<Property> properties = configuredPropertiesType.getProperty();
             final ComponentDescriptor<?> descriptor = builder.getDescriptor();
+
+            Map<String, String> removedProperties = new HashMap<>();
```
LosD: final

PR #1292, opened 2016-04-08
Bug/1290 remote transformer exposed as remote transformer

### LINKED ISSUE #1290: exposing remote component as remote component doesn't work
If DC monitor connects to DataCloud to get remote components and then the DC monitor itself is used as a server for remote components, the remote components exposed as remote components don't work.
Specifically, the enumerations types that are only in DataCloud classpath (not in monitor classpath) don't work.

jakubneubauer:
fixes #1290

### DIFF
--- a/engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
+++ b/engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
@@ -252,25 +252,33 @@ public class Serializator {
         }
 
         protected String enumValueToSchemaString(Enum<?> value) {
-            StringBuilder serialized = new StringBuilder();
-            serialized.append(value.name());
-            serialized.append(ENUM_ALIAS_SEPARATOR);
+            String enumValue = value.name();
+            String enumName;
+            String[] aliases = null;
             if(value instanceof  HasName) {
-                serialized.append(((HasName)value).getName());
+                enumName = ((HasName)value).getName();
             } else {
-                serialized.append(String.valueOf(value));
+                enumName = String.valueOf(value);
             }
             if(value instanceof HasAliases) {
-                String[] aliases = ((HasAliases)value).getAliases();
-                if(aliases != null) {
-                    for(String alias: aliases) {
-                        if(alias != null && !alias.isEmpty()) {
-                            serialized.append(ENUM_ALIAS_SEPARATOR).append(alias);
-                        }
-                    }
+                aliases = ((HasAliases)value).getAliases();
+            }
+            return Serializator.enumValueToSchemaString(enumValue, enumName, aliases);
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
```diff
@@ -252,25 +252,33 @@ public void acceptJsonFormatVisitor(JsonFormatVisitorWrapper visitor, JavaType t
         }
 
         protected String enumValueToSchemaString(Enum<?> value) {
-            StringBuilder serialized = new StringBuilder();
-            serialized.append(value.name());
-            serialized.append(ENUM_ALIAS_SEPARATOR);
+            String enumValue = value.name();
```
LosD: This could be final

engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
```diff
@@ -252,25 +252,33 @@ public void acceptJsonFormatVisitor(JsonFormatVisitorWrapper visitor, JavaType t
...
-                        }
-                    }
+                aliases = ((HasAliases)value).getAliases();
+            }
+            return Serializator.enumValueToSchemaString(enumValue, enumName, aliases);
+        }
+    }
+
+    public static String enumValueToSchemaString(String enumValue, String enumName, String[] aliases) {
+        StringBuilder serialized = new StringBuilder();
```
LosD: Same here

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -498,11 +504,40 @@ private static void setComponentAnnotations(ComponentDescriptor<?> componentDesc
...
-        // "Class" ? )
 
-        SchemaFactoryWrapper visitor = new SchemaFactoryWrapper();
+        // Special support for enumerations in case we are publishing info about a remote transformer from another server.
+        // (Client -> Server1 -> Server2)
+        // In this case if Server1 doesn't have the enumeration class on its classpath, the json schema cannot
+        // cannot be generated from a class => We do it "by hand" here.
+        if(EnumerationValue.class.isAssignableFrom(propertyDescriptor.getBaseType())
+                && propertyDescriptor instanceof EnumerationProvider) {
+            EnumerationValue[] values = ((EnumerationProvider) propertyDescriptor).values();
```
LosD: As well as most of this method (I know, I'm being a big anal about it... I blame @kaspersorensen :D)

LosD:
Code looks good to me, except for a few finals.

I'd like to test it a bit, but I have no idea how to set the monitor up to server remote components (I should probably get a crash course someday), so I'll just have to to trust you on that part :)

jakubneubauer:
no problem, I tried it and some other member of our team will cross-test it too.

PR #1298, opened 2016-04-14
change the sentence for processing finished message

### LINKED ISSUE #1156: Wording of "finished" in Progress Information may be misleading
I am running a UK suppression job an noticing this message in the Progress Information panel:

```
10:13:25 INFO: Processing of UK-sample-10000.csv finished. Generating results...
```

But actually the job has not finished because the UK suppression analyzer only returned a result future object. So in this case the wording should be more sensitive to whether or not all analyzer results are actually gathered or not.

ClaudiaPHI:
Fixes #1156

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
@@ -563,7 +563,7 @@ public final class ResultWindow extends AbstractWindow implements WindowListener
                 logger.info("rowProcessingSuccess: {}", job.getDatastore().getName());
                 _progressInformationPanel.updateProgressFinished(metrics.getTable());
                 _progressInformationPanel.addUserLog(
-                        "Processing of " + metrics.getTable().getName() + " finished. Generating results...");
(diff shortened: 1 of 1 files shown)


desktop/ui/src/main/java/org/datacleaner/windows/ResultWindow.java
```diff
@@ -563,7 +563,7 @@ public void rowProcessingSuccess(AnalysisJob job, final RowProcessingMetrics met
                 logger.info("rowProcessingSuccess: {}", job.getDatastore().getName());
                 _progressInformationPanel.updateProgressFinished(metrics.getTable());
                 _progressInformationPanel.addUserLog(
-                        "Processing of " + metrics.getTable().getName() + " finished. Generating results...");
+                        "Processing of " + metrics.getTable().getName() + " is completed. Generating results...");
```
arjansh: I'm not sure if "is completed" is really a change here, because I think it carries the same meaning as "finished". Maybe it should be something like "successful".

ClaudiaPHI: You are right. It's probably better.

arjansh:
LGTM!

Issue #1296, opened 2016-04-13
Datacloud Icon in notification panel and status panel


khouzvicka:
We can make the connection visible, for instance by adding a DataCloud icon in the status bar of DataCleaner (where it now reads "Community edition'). That icon can reflect the connection status (green if all's fine, yellow if credits are out, grey if not configured). A mouseover could display username and credit balance, and we can provide a link directly to the dc.org credit balance or history page (including the login - we know the password  ). This will make the integration even more tight.

ClaudiaPHI:
Closing because the issue was solved in https://github.com/datacleaner/DataCleaner/pull/1300

PR #1300, opened 2016-04-15
Feature/1296 datacloud notification icon

khouzvicka:
#1296 
Hi.
This is new information label about DataCloud status. It is on notification panel before information about edition of DC.
Size of glass panel is same like edition panel. Window is shown after click.
There is 4 states:
datacloudnotconfigured
datacloudok
datacloudnocredit
dataclouderror


khouzvicka:
@LosD @jakubneubauer 
Any suggestion?

api/src/main/java/org/datacleaner/configuration/RemoteServerState.java
```diff
@@ -22,6 +22,55 @@
 /**
  * States for connection to remote components.
  */
-public enum RemoteServerState {
-    UNKNOWN, OK, ERROR;
+public class RemoteServerState {
+
+    public enum State {
+        UNKNOWN, OK, ERROR, NO_CREDIT;
```
LosD: Isn't `UNKNOWN` more like `NOT_CONNECTED`?

khouzvicka: UNKNOWN is base state before real check or when user isn't connect to DataCloud.
NOT_CONNECTED sounds like error state for me.

LosD: Okay. I don't really mind, but NOT_CONNECTED sounds like the user is... well... Not connected :)

LosD:
Icons looks fine to me. Are they using the GMC colors (where applicable)? If not, I'd tweak them to do so.

khouzvicka:
Only green is gmc. I will check all colors with UX.

LosD:
Of course, if the GMC colors are so far off that the traffic light like presentation is lost, just leave them as-is :)

desktop/ui/src/main/java/org/datacleaner/widgets/DataCloudStatusLabel.java
```diff
@@ -0,0 +1,100 @@
...
+        });
+    }
+
+    private void setIcon(RemoteServerState.State state) {
+        switch (state) {
+        case UNKNOWN:
+            setIcon(ImageManager.get().getImageIcon("images/datacloud/cloudNotUsed.png", IconUtils.ICON_SIZE_SMALL));
+            break;
+        case NO_CREDIT:
+            setIcon(ImageManager.get().getImageIcon("images/datacloud/cloudNoCredit.png", IconUtils.ICON_SIZE_SMALL));
```
LosD: It'd maybe use a little more generic name here. There's a good chance that this icon could be reused for other warning-level problems later.

desktop/ui/src/main/java/org/datacleaner/panels/DataCloudInformationPanel.java
```diff
@@ -0,0 +1,185 @@
...
+        String formattedCredit = "";
+
+        for (int i = creditStr.length() - 1; i >= 0; i--) {
+            int index = creditStr.length() - 1 - i;
+            if (index != 0 && index % 3 == 0 && creditStr.charAt(i) != '-') {
+                formattedCredit = " " + formattedCredit;
+            }
+            formattedCredit = creditStr.charAt(i) + formattedCredit;
+        }
+        return formattedCredit;
```
LosD: That's pretty low-level. Can't a standard number or string formatter do this?

desktop/ui/src/main/java/org/datacleaner/panels/DataCloudInformationPanel.java
```diff
@@ -0,0 +1,185 @@
...
+
+            if (!remoteServerState.isEmailConfirmed()) {
+                panelContent = addLine(panelContent, "Email is not confirmed.");
+            }
+        }
+
+        if (remoteServerState.getActualState() == RemoteServerState.State.UNKNOWN) {
+            panelContent = addLine(panelContent, "Datacloud is not configured.");
+            panelContent = addLine(panelContent, "You can set your credentials in");
+            panelContent = addLine(panelContent, "<b>Options dialog - DataCloud</b>");
```
LosD: All of this is very different from what we usually do. How come this isn't using standard Swing components instead of inline HTML?

LosD:
A few issues, but mostly minor stuff.

desktop/ui/src/main/java/org/datacleaner/panels/DataCloudInformationPanel.java
khouzvicka: You are right. String.format("%,d", n) is more better.

khouzvicka:
new names od icon:
cloudRed, cloudGrey, cloudGreen, cloudOrange and new colors. GMC :-)
cloudgreen
cloudgrey
cloudorange
cloudred

desktop/ui/src/main/java/org/datacleaner/panels/DataCloudInformationPanel.java
khouzvicka: @LosD 
In this way I have only one component and I can change it more flexible. Sometimes I have only two lines, sometimes four. Red - Black credit balance. Bold text.
And I am sure, that I have place for it. I do not have to use complex layout. I create only html.

LosD: Hmmmm... But at the same time you break consistency with other components, increase overall complexity and move styling away from the proper place and into this component for IMHO very limited advantages.

Anyway, it's a pretty shielded off component, so I'll let it be for now, but in general this is the wrong way to go, especially for what is supposed to just be an simple status display.

api/src/main/java/org/datacleaner/configuration/RemoteServerState.java
khouzvicka: you got me. I will change it to NOT_CONNECTED.

LosD:
Code mostly LGTM. Merging! :)

PR #1302, opened 2016-04-15
Fixed a few minor code style issues that gave compiler warnings.

kaspersorensen:
These issues showed up when I checked out DC into my new eclipse. Felt that they would be easy and good to fix :)


arjansh:
LGTM!

Issue #1309, opened 2016-04-20
DataCloud Status icon unrespons on change credentials


khouzvicka:
DataCloud icon unrespons on change of credentials in DataCloud options dialog.

PR #1310, opened 2016-04-20
#1309 - Check status of DataCloud also after credentials update.

khouzvicka:
#1309 DataCloud icon unrespons on change of credentials in DataCloud options dialog.
I fixed this bug


jakubneubauer:
LGTM

PR #1318, opened 2016-04-22
Remote^2 transformer icon was not working

jakubneubauer:
Icon of a remote transformer being exposed as a remote transformer was not working. Fixed.


jakubneubauer:
was reviewed by peer face to face :-)

PR #1320, opened 2016-04-26
Added a validate method to the create staging table analyzer which checks if the configure…

### LINKED ISSUE #1319: Preview of staging table produces error in case datastore name conflicts
When you configure a create staging table analyzer with a datastore name, which is already use by an existing non H2 database datastore or by a H2 database datastore which is not located in your "Written datastores" directory (which you configure in the "Files & directories" field set in the "General" tab of the "Options" dialog), you get an unexpected error when clicking the "Preview table" button in the analysis results screen.

createstagingtableerror

This is caused by the fact that the create staging table analyzer always writes its results to the configured "Written datastores" directory using the configured datastore name (and creates a datastore using that name if none exists), but the "Preview table" button retrieves the datastore from the datastore catalog, which may point to a different source in case the datastore name configured in the create staging table already exists pointing to a source which is not an H2 database in the "Written datastores" directory.

arjansh:
Fixes #1319.

Added a validate method to the analyzer which checks if the configured datastore name doesn't conflict with a datastore in the datastore catalog, that is there should either not be a datastore for that name in the datastore catalog, or if there is one, it has to be an H2 database located in the user's saveDatastoreDirectory.

Also added some more exception catching to the AbstractComponentBuilder, to avoid a com.google.inject.internal.util.$ComputationException: java.lang.ArrayIndexOutOfBoundsException when opening a job containing a create staging table analyzer with a conflicting datastore name.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/extension/output/CreateStagingTableAnalyzer.java
+++ b/desktop/ui/src/main/java/org/datacleaner/extension/output/CreateStagingTableAnalyzer.java
@@ -30,10 +30,13 @@ import org.datacleaner.api.Configured;
 import org.datacleaner.api.Description;
 import org.datacleaner.api.Distributed;
 import org.datacleaner.api.HasLabelAdvice;
+import org.datacleaner.api.Validate;
 import org.datacleaner.beans.writers.WriteDataResult;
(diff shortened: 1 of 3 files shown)


LosD:
LGTM!

PR #1324, opened 2016-05-03
Use credentials for REST mode.

### LINKED ISSUE #1323: REST ElaticSearch client does not set credentials
If credentials are set in the dialog for REST ElasticSearch datastores, those are not set on the data context's HttpClientConfig.

LosD:
Fixes #1323

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
+++ b/engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
@@ -19,8 +19,6 @@
  */
 package org.datacleaner.connection;
 
-import static org.elasticsearch.node.NodeBuilder.nodeBuilder;
-
 import java.util.List;
 
 import org.apache.metamodel.DataContext;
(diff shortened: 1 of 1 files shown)


engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
```diff
@@ -169,10 +169,15 @@ private Client getClientForNodeAndTransportProtocol() {
 
     private JestClient getClientForRestProtocol() {
         final JestClientFactory factory = new JestClientFactory();
-        factory.setHttpClientConfig(new HttpClientConfig.Builder("http://" + _hostname + ":" + _port).multiThreaded(
-                true).build());
-        final JestClient client = factory.getObject();
-        return client;
+        HttpClientConfig.Builder builder =
```
ClaudiaPHI: final

ClaudiaPHI:
LGTM!

engine/core/src/main/java/org/datacleaner/connection/ElasticSearchDatastore.java
LosD: No can do: builder is overwritten if a username is set.

PR #1327, opened 2016-05-04
Feature/1322 register a hadoop datastore

### LINKED ISSUE #1322: DC monitor: Register a Hadoop datastore
Add registration for a file located to Hadoop cluster to this form. 

image

This might include: 
- [ ] Selection of the server
- [ ] Selection of the file

ClaudiaPHI:
Fixes #1322  

image

image

image

image

### DIFF
--- a/monitor/extensions/csv-datastore-wizard/src/main/java/org/datacleaner/monitor/server/wizard/CsvDatastoreWizardSession.java
+++ b/monitor/extensions/csv-datastore-wizard/src/main/java/org/datacleaner/monitor/server/wizard/CsvDatastoreWizardSession.java
@@ -23,25 +23,25 @@ import java.io.File;
 
 import javax.xml.parsers.DocumentBuilder;
 
-import org.datacleaner.configuration.DatastoreXmlExternalizer;
+import org.apache.metamodel.csv.CsvConfiguration;
+import org.apache.metamodel.util.FileHelper;
+import org.apache.metamodel.util.FileResource;
+import org.apache.metamodel.util.Resource;
+import org.datacleaner.configuration.DomConfigurationWriter;
 import org.datacleaner.connection.CsvDatastore;
 import org.datacleaner.monitor.shared.model.DCUserInputException;
 import org.datacleaner.monitor.wizard.WizardPageController;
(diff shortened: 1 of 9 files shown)


ClaudiaPHI:
Does anyone want to review/test this PR ?

LosD:
I'll take a look

monitor/extensions/csv-datastore-wizard/src/main/java/org/datacleaner/monitor/server/wizard/CsvDatastoreWizardSession.java
```diff
@@ -86,26 +88,31 @@ protected WizardPageController nextPageControllerExisting() {
...
-                                    // only .csv and .tsv files are allowed to
-                                    // be referenced on the server, for security
-                                    // reasons.
-                                    throw new DCUserInputException(
-                                            "For security reasons, only existing .csv, .tsv or .txt files can be referenced on the server");
+                    protected WizardPageController nextPageController(Resource resource) {
+                        if (resource instanceof FileResource) {
+                            final File file = ((FileResource) resource).getFile();
+                            final String filepath = file.getAbsolutePath();
+                            if (!filepath.toLowerCase().endsWith(".csv")) {
```
LosD: That's a pretty nasty construct. Can we use ORing instead (I'm also pretty convinced we have are tools for checking extensions, maybe they can take varargs or an array of extensions to check for)?

Why the special case for txt files?

ClaudiaPHI: I will improve the with the "OR" and check about the extensions. Why the special cases ? I have no idea.

LosD: Oh, didn't notice it in the old code.

LosD:
Only very minor stuff in the code (and just now realised that it's not in new code, but in reformatted old code).

However, I'm pretty sure that that a .tsv is just as legal as a .csv file for Hadoop? It's the same format, just using tab and 0x10 for EOL while CSV is a little more free to use other characters.

PR #1332, opened 2016-05-11
ElasticSearch: Remember connection type and don't use cluster name

### LINKED ISSUE #1311: ElasticSearch datastore dialog requires cluster name even for REST connections
When connecting to an ElasticSearch datastore via the REST protocol, the dialog still requires me to specify cluster name, although this is to my knowledge not needed or even used when connecting via REST. But I have to specify it or else the "Register datastore" button is disabled.

image

### LINKED ISSUE #1331: ElasticSearchDatastore forgets REST setting
The ElasticSearchDatastores does not store the client type. That only affects REST clients, as loading has a backup way to detech the datastore type by looking if a port has been set (cluster connections does not use ports.

LosD:
This will store client connection types (none was remembered, but only REST was affected). Will also disable cluster name for REST clients, as it is not used.

 Fixes #1331
 Fixes #1311

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
@@ -220,7 +220,7 @@ public class ElasticSearchDatastoreDialog extends AbstractDatastoreDialog<Elasti
         if (originalDatastore == null) {
             _hostnameTextField.setText("localhost");
             _portTextField.setText(""+ElasticSearchDatastore.DEFAULT_PORT);
-            _clientTypeComboBox.setSelectedItem(ClientType.REST);
+            _clientTypeComboBox.setSelectedItem(DEFAULT_CLIENT_TYPE);
         } else {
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/java/org/datacleaner/windows/ElasticSearchDatastoreDialog.java
```diff
@@ -82,7 +80,6 @@ public ElasticSearchDatastoreDialog(WindowContext windowContext, MutableDatastor
         setSaveButtonEnabled(false);
 
         _clientTypeComboBox = new JComboBox<>(ClientType.values());
-        _clientTypeComboBox.setSelectedItem(DEFAULT_CLIENT_TYPE);
```
arjansh: With the removal of this line the private constant DEFAULT_CLIENT_TYPE is no longer used.

LosD: That's actually more of a failing in line 223, where I used ClientType.REST instead.

arjansh:
Outside of my small inline comment it looks good to me.

arjansh:
LGTM!

PR #1333, opened 2016-05-11
Hot folder monitoring

### LINKED ISSUE #1329: Hot folder monitoring
There's a request for some kind of watch service in the monitor, which can trigger a job execution when a new file is added to the folder. 

To keep this simple, we've decided to skip on advanced features: Just use a single trigger file. If it changes or appears, trigger the job straight away.
To do and not to do for now:
- Change the scheduling GUI to add the option to watch a file or folder for triggering.
- If folder: Any file added or changed (not removed) should trigger the job.
(low prio) Maybe look at some way to simplify the GUI. It's getting a bit messy.
- Do NOT implement reading zip files.
- Do NOT use any kind of configuration for the file.

(That last one will probably change soon, but lets keep it simple for now)

arjansh:
Fixes #1329.

Some things which may need special attention when reviewing this pull request:

- The ScheduleServiceImpl now spawns a `org.apache.commons.io.monitor.FileAlterationMonitor` instance on construction, which spawns a monitoring thread in the initialize method. Please check if the implementation doesn't introduce memory leaks or keeps unwanted references to objects.
- The hot folder monitor is quite trigger happy, when a hot folder trigger is added to as a schedule to a job it instantly starts the job for all initial changes it detects, which may be more than one initially. Please test this and provide some feedback whether or not we want to suppress these first time triggers.
- The Customize schedule dialog in DataCleaner monitor starts with a blue paragraph which describes some of different ways in which you can configure the schedule. It may be desirable to also add the hot folder trigger to this paragraph, which as yet has not been done.

### DIFF
--- a/monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinition.java
+++ b/monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinition.java
@@ -43,7 +43,8 @@ public class ScheduleDefinition implements Comparable<ScheduleDefinition>, Seria
     private boolean _distributedExecution;
     private String _dateForOneTimeSchedule;
 	private Map<String,String> _jobMetadataProperties;
-    private boolean _runOnHadoop; 
+    private boolean _runOnHadoop;
+    private String _hotFolder;
 
     // no-args constructor
     public ScheduleDefinition() {
@@ -138,8 +139,9 @@ public class ScheduleDefinition implements Comparable<ScheduleDefinition>, Seria
             return TriggerType.PERIODIC;
         } else if(_dateForOneTimeSchedule!= null){
         	return TriggerType.ONETIME;
-        }
-        else {
+        } else if (_hotFolder != null) {
+            return TriggerType.HOTFOLDER;
+        } else {
             return TriggerType.MANUAL;
         }
     }
@@ -159,7 +161,15 @@ public class ScheduleDefinition implements Comparable<ScheduleDefinition>, Seria
 	public Map<String,String> getJobMetadataProperties(){
 		return _jobMetadataProperties;
 	}
-    
+
+    public String getHotFolder() {
+        return _hotFolder;
+    }
+
+    public void setHotFolder(String hotFolder) {
+        _hotFolder = hotFolder;
+    }
+
     @Override
     public int hashCode() {
         final int prime = 31;
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
@@ -19,6 +19,7 @@
  */
 package org.datacleaner.monitor.server;
 
+import java.io.File;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.text.ParseException;
@@ -35,6 +36,9 @@ import java.util.Set;
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 
+import org.apache.commons.io.monitor.FileAlterationListener;
+import org.apache.commons.io.monitor.FileAlterationMonitor;
+import org.apache.commons.io.monitor.FileAlterationObserver;
 import org.apache.metamodel.util.Action;
 import org.apache.metamodel.util.CollectionUtils;
 import org.apache.metamodel.util.Func;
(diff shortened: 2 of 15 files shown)


monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinition.java
```diff
@@ -43,7 +43,8 @@
     private boolean _distributedExecution;
     private String _dateForOneTimeSchedule;
 	private Map<String,String> _jobMetadataProperties;
-    private boolean _runOnHadoop; 
+    private boolean _runOnHadoop;
+    private String _hotFolder;
```
LosD: Not on you, but why in the world does Git/Github think that the _runOnHadoop line has changed?

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -665,4 +703,62 @@ public String getServerDate() {
...
+        private final JobIdentifier job;
+        private final TenantIdentifier tenant;
+
+        private HotFolderAlterationListener(final JobIdentifier job, final TenantIdentifier tenant) {
+            this.job = job;
+            this.tenant = tenant;
+        }
+
+        public void onStop(FileAlterationObserver observer) {
+            // Do nothing.
```
LosD: Ew. A lot of these. Isn't there an adapter version of FileAlterationListener?

monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinition.java
arjansh: Because it actually has changed (either I or my IDE removed an empty space from the end of the line).

LosD: Oh. The beauty of not printing whitespaces (didn't Github used to highlight the actual change in a darker color?)

LosD:
Two things:
-We'll also need to allow watching for a single file, not only all of the folder. I'd maybe do it by keeping the general structure where the entire folder is watched, and then just filter the filename before doing the invocation.
- Would this also work with a Hadoop execution?

@ClaudiaPHI is probably the one with most insight into running Hadoop from the monitor, maybe a quick review watching for dangers would be nice.

arjansh:
@LosD 
- I switched from completely implementing the FileAlterationListener to extending the FileAlterationListenerAdaptor.
- I changed it so you can watch single files.
- I did some testing with Hadoop and I can successfully tigger a job to run on a Hadoop cluster using the hot folder. This does however not imply that I can use a folder/file on hdfs as hot folder.

LosD:
No, I wouldn't expect any kind of remote folder watching. It would be super cool, but probably a bit too poll'y (unless they have support in their end, like AWS's long poll etc.)

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -385,6 +419,17 @@ else if (triggerType == TriggerType.ONETIME) {
         }
     }
 
+    private FileAlterationObserver createObserver(String fileName) {
+        File file = new File(fileName);
```
LosD: final! :)

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -369,6 +389,20 @@ else if (triggerType == TriggerType.ONETIME) {
                                 cronExpression);
                         _scheduler.scheduleJob(jobDetail, trigger);
                     }
+                } else if (triggerType == TriggerType.HOTFOLDER) {
+                    String hotFolder = schedule.getHotFolder();
```
LosD: final

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -665,4 +714,40 @@ public String getServerDate() {
         logger.info("Date and TimeStamp for one time schedule: {} | {}", serverDate, timeStampName);
         return serverDateFormat;
     }
+
+    void removeHotFolder(String jobIdentifier) {
+        FileAlterationObserver observer = _registeredHotFolders.get(jobIdentifier);
```
LosD: final

LosD:
Code looks good, except a few finals.

I know that @ClaudiaPHI is doing a lot of testing now, so I'll be lazy and totally skip that :)

ClaudiaPHI:
I have tested and it works fine, but I have some concerns: 

- The job seems to be triggered multiple times. For example: it's triggered once when opening the file, and then after I am saving it. Could we do sometimes to prevent starting jobs every second ? 
- is it ok to assume that if a job is already running do not start again another one ?

arjansh:
@ClaudiaPHI 
- Regarding your first comment that the job is triggered when opening a file, did you monitor a folder or a specific file? Because this doesn't happen for me when I monitor a file. This may be caused by shadow files being created on disk when opening a file for editing.
- Regarding your second comment, I don't know if we can assume that. In most cases I would say yes, but maybe some users want to run a job after each individual change to a file, so also is a file is changed again before the current job triggered by a previous change has finished. Also this could mean the final change to a file didn't trigger a job, because an earlier trigger job was already running when the final change was done.

ClaudiaPHI:
Wait until release is completed. @LosD will decide when to merge it

PR #1337, opened 2016-05-13
Upgrade jackson and spark libraries versions

arjansh:
DataCleaner still uses an older version of the com.fasterxml.jackson libraries than metamodel, which can cause problems when testing custom components for DataCleaner during the development process directly from your IDE. Upgrading these libraries to the latest version remedies these problems. By upgrading to the newest version we can also remove a workaround implemented in the org.datacleaner.restclient.Serializator class.

Because the current version of the spark libraries don't work with the newer com.fasterxml.jackson libraries, the spark version also has to be updated.

### DIFF
--- a/engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
+++ b/engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
@@ -21,17 +21,12 @@ package org.datacleaner.restclient;
 
 import java.io.IOException;
 import java.lang.reflect.Type;
-import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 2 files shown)


LosD:
Damn. My VM is still funky (I imagine it will stay that way until my machine has been reinstalled), so I can't really test the Spark part.

@ClaudiaPHI can I get you to do a quick check that Spark didn't break?

ClaudiaPHI:
yes

ClaudiaPHI:
It works with Spark 1.5.1

LosD:
(Reason @ClaudiaPHI mentions Spark 1.5.1 is that I explicitly asked her to test running it with an old Spark, since I know many distributions will use old Spark versions)

PR #1340, opened 2016-05-17
Feature/1170 replace inlined synonyms

### LINKED ISSUE #1170: New default and label for Synonym lookup property "Look up every token"
The property "Look up every token" in Synonym lookup transformer is not quite as intuitive as we would want it - and we would want it to be turned ON by default. So...

 * Change default value to ```true```
 * Change name (but retain an ```Alias``` for backwards compatibility) to: ```Replace inlined synoyms```
 * Update the description documentation to not mention "tokens" but rather talk about substrings

kaspersorensen:
Fixes #1170 

 * Rename and convention change for strange property "Look up every token"
 * Fixed an unrelated unittests which is only visible if you're on a timezone (and time of day) when your time offset from CET is sufficient to provoke another date when applying a timeless date format.


LosD:
LGTM!

PR #1345, opened 2016-05-19
Specify replacement datastore from commandline

### LINKED ISSUE #565: Specify "replacement" datastore name on command line
Just got the question from a customer on how to run the same job with multiple datastores. Normally we use the "dynamically overriding configuration elements" idea here (http://datacleaner.org/resources/docs/4.0.7/html/ch21s06.html) but one thing that isn't so great about that is that it requires the same type of datastore. I cannot make it replace e.g. a CSV datastore with a Excel datastore. And it also still requires that there is consistency in the naming of datastores in job file and conf.xml.

So rather it would be better to allow using the (existing) "-ds" parameter from the command line while executing a job. That way you could have a job that refers to datastore A, but if you pass "-ds B" then it will use datastore B at runtime. Of course (like always) if the column paths are then not resolveable in datastore B etc., then it will break.

LosD:
This makes it possible to override the datastore to a job with by using -ds on the command line.

Fixes #565

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/cli/CliArguments.java
+++ b/desktop/ui/src/main/java/org/datacleaner/cli/CliArguments.java
@@ -122,7 +122,7 @@ public class CliArguments {
     @Option(name = "-list", usage = "Used to print a list of various elements available in the configuration")
     private CliListType listType;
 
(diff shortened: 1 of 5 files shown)


ClaudiaPHI:
LGTM!

PR #1348, opened 2016-05-19
thread-safe creation of shared instance of ConvertToDateTransformer

jakubneubauer:


LosD:
Shouldn't the block also be synchronized? Otherwise you risk different threads both creating a new instance.

jakubneubauer:
In this case it is not necessary, creating more instances doesn't harm.

LosD:
LGTM, then (can't merge myself, silly Android app doesn't allow it)

jakubneubauer:
no problem :-)

PR #1350, opened 2016-05-19
Upgraded to latest MetaModel version

kaspersorensen:


LosD:
The changes to the column names... Wouldn't those possible kill existing jobs?

Then we need some kind of compatibility fix before we can merge this.

kaspersorensen:
If you think so, be my guest :-) I think it's a really unusual corner case and there are much more likely bugfixes in the MM update :-)

LosD:
While I agree that the MM bugfixes are REALLY important, I don't think it would be okay to break people's jobs with a micro update, corner case or not. 

Luckily we have quite a bit of space in this sprint, so let's leave this PR open then I'll take it over Monday and do some sort of existing job conversion.

LosD:
Wait, it's a job writer test. I thought it was the input colmns names reading if there are any unnamed columns. Wouldn't that also be affected, or does it only kick in if there are multiple?

A single unnamed column is unfortunately a pretty normal situation (multiple unnamed columns are even more normal, but that was broken already, so nothing lost there :))

LosD:
Dang, tried taking a look at it, and it wouldn't be simple at all, some sort of reverse ColumnNamingStrategy would be required (as we can't guarantee that all columns are mapped, so just running the normal `DelegatingIntrinsicSwitchColumnNamingStrategy` on the mapped job columns would be no guarantee).

I don't think I can defend spending that amount of time on it, so we'll have to answer problems as/if they come.

LosD:
Merging! :)

PR #1355, opened 2016-05-23
Feature/1330 ebcdic fixed width files support

### LINKED ISSUE #1330: EBCDIC fixed width files support
Support for input data fixed-width files in EBCDIC encoding is required. In GUI dialog for fixed-width datastore this new encoding (and perhaps other options) should be added. Main changes will be in MetaModel itself.

jhorcicka:
Fixes #1330 and depends on https://github.com/apache/metamodel/pull/103. 

FixedWidthDatastore GUI dialog was updated so that it can support EBCDIC files with a header that should be skipped and no new-line-characters in the content. 

ebcdic-dialog


pom.xml
```diff
@@ -17,7 +17,8 @@
 		<httpcomponents.version>4.4.1</httpcomponents.version>
 		<guava.version>16.0.1</guava.version>
 		<curator.version>2.6.0</curator.version>
-		<metamodel.version>4.5.2</metamodel.version>
+		<!--<metamodel.version>4.5.3</metamodel.version>-->
+		<metamodel.version>4.5.4-SNAPSHOT</metamodel.version>
```
jhorcicka: New release of MM is required here.

desktop/ui/src/main/java/org/datacleaner/windows/FixedWidthDatastoreDialog.java
```diff
@@ -116,31 +121,18 @@ protected void onChange(DocumentEvent event) {
...
-			}
-		});
-		_headerLineComboBox.addListener(new Listener<Integer>() {
-			@Override
-			public void onItemSelected(Integer item) {
-				onSettingsUpdated(false);
-			}
-		});
+	private JCheckBox createCheckBox(String label, boolean selected) {
+		JCheckBox checkBox = new JCheckBox(label, selected);
```
LosD: final

desktop/ui/src/main/java/org/datacleaner/windows/FixedWidthDatastoreDialog.java
```diff
@@ -197,8 +189,60 @@ private void onSettingsUpdated(boolean autoDetectEncoding) {
...
+			if (bytesRead != -1 && bytesRead <= bufferSize) {
+				bytes = Arrays.copyOf(bytes, bytesRead);
+			}
+
+			return bytes;
+		} catch (IOException e) {
+			logger.error("IOException occurred while reading sample buffer", e);
+			return new byte[0];
+		} finally {
+			FileHelper.safeClose(fileInputStream);
```
LosD: You should be able to skip the "safeClose" and use try-with-resources directly.

In fact, there's nothing safe about safeClose in my mind. It swallows exceptions and logs them at debug level.

desktop/ui/src/main/java/org/datacleaner/windows/FixedWidthDatastoreDialog.java
```diff
@@ -197,8 +189,60 @@ private void onSettingsUpdated(boolean autoDetectEncoding) {
...
+	private int getRecordDataLength() {
+		int length = 0;
+
+		if (_valueWidthTextFields != null && _valueWidthTextFields.size() > 0) {
+			for (JXTextField textField : _valueWidthTextFields) {
+				try {
+					final int columnWidth = Integer.parseInt(textField.getText());
+					length += columnWidth;
+				} catch (NumberFormatException e) {
+					// silently ignored
```
LosD: Don't we want to pop up a "you screwed up"-kinda dialog? Or would it prevent entry at all?

LosD:
Code mostly LGTM, a few issues, though.

Another issue is that there's generally a lot of "1 real change, whole file reformatted". That really makes PRs hard to read, because the actual change drowns in a ton of reformatted lines. Please never do whole-file reformats.

LosD:
An exceptions is if most of the file was changed anyway.

I usually reformat the method I changed (and if it is a huge method, I might just mark e.g. 5 lines on each side)

jhorcicka:
Checks are failing because of SNAPSHOT MetaModel dependency.

LosD:
Can we merge this with 4.5.4? Even with the diacritics bug this is an improvement for 5.1

PR #1358, opened 2016-05-24
Specify replacement properties in job invocation web service call.

### LINKED ISSUE #1351: Specify replacement properties in job invocation web service call

arjansh:
Fixes #1351.


arjansh:
Cannot be merged yet. I haven't implemented the necessary unit tests yet (which I'm working on right now).

ClaudiaPHI:
LGTM!

monitor/services/src/main/java/org/datacleaner/monitor/configuration/ConfigurationCache.java
```diff
@@ -106,8 +107,12 @@ protected DataCleanerConfiguration decorateConfiguration(DataCleanerConfiguratio
     }
 
     protected DataCleanerConfiguration readConfiguration() {
+        return readConfiguration(null);
+    }
+    
+    DataCleanerConfiguration readConfiguration(Map<String, String> overrideProperties) {
         final ConfigurationReaderInterceptor interceptor = new MonitorConfigurationReaderInterceptor(_repository,
-                _tenantContext, _injectionManagerFactory);
+                _tenantContext, overrideProperties, _injectionManagerFactory);
```
LosD: Isn't it dangerous to involve the overrides in the cache?

arjansh: When the readConfiguration method is invoked from the ConfigurationCache itself, the value for the overrideProperties is always null. It only contains an object when this method is invoked directly from the TenantContextImpl (and the result is not cached). An other option would be to create a copy of this method on the TenantContextImpl class, but that would involve some more refactoring on that class so I have access to the InjectionManagerFactory.

LosD: Hmmm... I don't like it being part of the cache, but I don't really like the alternative either.

The best would be to get the configuration first (cached), and the apply overrides after, but the interceptor design pretty much precludes that... :(

Oh well, leave it. :)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobTriggeringController.java
```diff
@@ -49,19 +55,37 @@
...
+    }
 
+    @RequestMapping(method = RequestMethod.POST, produces = "application/json", 
+            consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
+    @ResponseBody
+    @RolesAllowed(SecurityRoles.SCHEDULE_EDITOR)
+    public Map<String, String> invokeJob(@PathVariable("tenant") final String tenant,
+            @PathVariable("job") String jobName, @RequestParam(value = "block", required = false) Boolean block,
+            @RequestParam(value = "timeoutMillis", required = false) Integer timeoutMillis,
+            @RequestParam(value = "overrideProperties") final MultipartFile overrideProperties) throws Throwable {
```
LosD: Isn't a MultipartFile a weird standard when it's not usually browser forms that would invoke this?

I'd much prefer simple text file or JSON. But let's maybe make an issue for that.

arjansh: Do you mean the body of the request should contain plain text which is can then be parsed as properties?

LosD: Yeah, that should make for a simpler request from most sources.

LosD:
Code LGTM, but would prefer text/plain as well as json/application as content type for requests. Preferably as alternative(/s) to the already existing multipart (no need to nuke a perfectly functioning way of submitting)

But I also thing we should do that as its own issue.

@arjansh I want to test, what method did you use to do manual testing?

arjansh:
@LosD I created a test.html which I open in a browser where I have an active monitor session, which looks like this:
```
<form action="http://localhost:8080/DataCleaner-monitor/repository/demo/jobs/incomes.trigger" method="post" enctype="multipart/form-data">
  <p>
  Timeout millis: <input type="text" name="timeoutMillis" /><br />
  Block: <input type="checkbox" name="block" /><br />
  Override properties: <input type="file" name="overrideProperties"><br />
  <button type="submit">Submit</button>
  </p>
</form>
```

This triggers a job named "incomes" and I browse to a properties file which contains the property I want to overrule, like this:
`datastoreCatalog.customers.filename=C:\\Temp\\customers2.csv`

LosD:
Ah, that should work. I'll try using Postman if I can get it to login properly, but this should be a nice fallback.

arjansh:
@losd You want both the json/application and text/plain? In that case, for the json/application, do you want the timeoutMillis and block parameter's values inside the json or as request parameters?

arjansh:
Using an actual form on an HTML page is the easiest way not to bugger up the multipart/form-data post.

LosD:
That's the exact reason I try to avoid it, I want to see how it handles non-forms, which I think would be the standard client for this :)

LosD:
I'd only keep the job properties inside the POST data in all types, block and timeout is about the request.

LosD:
multipart test worked perfectly

LosD:
LGTM and works with all 3 submission forms now. Let's merge! :)

PR #1360, opened 2016-05-25
Bug/1328 ordering of the components alphabetically

### LINKED ISSUE #1328: (No) alphabetic ordering of components
I just noticed that the components seem to no longer be alphabetically ordered in the menu/tree. This makes them a bit harder to find I think... 

image

I see this happening at least in the latest 5.0.2 release of DC.

ClaudiaPHI:
Fixes #1328 

image

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/tree/SchemaTree.java
@@ -28,6 +28,8 @@ import java.beans.PropertyChangeListener;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


LosD:
Did you find out what caused the original break of this?

LosD:
(It used to be sorted, but all of a sudden it wasn't... But this looks like an entirely new implementation)

ClaudiaPHI:
It never used to be sorted. I saw the issue long before Kasper saw it. Only the categories were sorted.

LosD:
That is... not true:
image

(4.5.3)

ClaudiaPHI:
I believe this is a simple coincidence of how the descriptors are loaded, but no specific ordering of descriptors was made in that class.

LosD:
Pretty crazy coincidence

jhorcicka:
While working on "remote components" I remember we had an issue with the order of components. It was different on various operating systems.

LosD:
It may very well be a change in how the descriptors are presented (sorted vs unsorted), but it's not something random. _All_ of the components are sorted alphabetically, at least in 4.5.3.

That certainly doesn't make this fix incorrect in any way, I just don't like not knowing what happened.

LosD:
We're both running Windows 7... Could it be Java version?

LosD:
D'oh, now I'm being stupid, we're also both seeing it unordered in DC 5 :)

LosD:
Wait, @ClaudiaPHI could it be that you saw it unordered in 4.5 when you were using the Mac?

ClaudiaPHI:
It may be so. I remember I saw it long time a go, tried to find the cause, did not find it.

LosD:
This is going to bother me forever :)

Oh well... This works, let's merge!

PR #1363, opened 2016-05-26
Read replacement properties file in hot folder

### LINKED ISSUE #1362: Specify replacement properties in hot folder trigger file

arjansh:
Fixes #1362.

Made it possible to add a properties file in the hot folder (or the folder in which the hot file is located), which is used to override properties in the default configuration to run the job with. By convention the properties file has to have the same name as the job and the .properties extension. So if you have a job named "Analyze customers" the properties files has to be named "Analyze customers.properties".

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
@@ -20,6 +20,7 @@
 package org.datacleaner.monitor.server;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


LosD:
Code looks good. but I think the pattern is too limited. I think we must be able to support multiple files, so they can have a sort of history (like making an entry call Analyze-1.properties, Analyze-2.properties, Analyze-3.properties, Analyze-4.properties, etc.).

I'd go for any .properties file, or maybe .trigger.properties.

By the way, we ONLY want to react to the changed file. If a file changes, we do not read any other file for the properties.

LosD:
We agreed to merge as-is and let the sprint review decide if we want to change it.

Test went perfect, merging!

PR #1365, opened 2016-05-26
Bug/1364 inform the user why cannot run on hadoop

### LINKED ISSUE #1364: Inform the user why the datastore in not valid for running on Hadoop
- The button for running a job on Hadoop greyed out even if the job contains a hdfs resource. 
Inform the user why the CSV datastore is not valid. 

- [ ]  Multilines  are allowed
- [ ] The encoding is not UTF-8

ClaudiaPHI:
Fixes #1364

### DIFF
--- a/engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopJobExecutionUtils.java
+++ b/engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopJobExecutionUtils.java
@@ -53,26 +53,28 @@ import org.datacleaner.util.xml.XmlUtils;
 import com.google.common.collect.Iterators;
 
 /**
- * Utility class for creating a job that runs on
- * Hadoop with Spark.
+ * Utility class for creating a job that runs on Hadoop with Spark.
  *
  */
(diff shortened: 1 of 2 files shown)


engine/env/spark/src/main/java/org/datacleaner/spark/utils/HadoopJobExecutionUtils.java
```diff
@@ -53,26 +53,28 @@
...
+                final CsvDatastore csvDatastore = (CsvDatastore) datastore;
+                if (!isValidMultilines(csvDatastore) || !isValidEnconding(csvDatastore)){
+                    return false; 
+                }
+            }
+            return true; 
+        }else{
+            return false; 
+        }
+    }
```
LosD: There seems to be some code style issues here (spaces around scopes etc.)

LosD:
I don't see the message anywhere in the code? Did that already exist?

ClaudiaPHI:
The messages are in commercial edition.

LosD:
Ah, super.

Except for the formatting in a single method, code LGTM.

Didn't test yet, though.

PR #1368, opened 2016-05-27
MDM-7040 added second api that accepts different input body

JoosjeBoon:


monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobInvocationController.java
```diff
@@ -184,6 +153,90 @@ public String eval(String columnPath) {
         return output;
     }
 
+    /**
+     * Takes a JSON request body in this form (2 rows with 1 int and 2 strings
+     * each):
+     *
```
HettyK: This comment is not so clear: Should be something like: takes a JSon request body containing an array of key value pairs. and then the example.

LosD:
Looks good to me! Might want to add final to most of the new variables though, but we not exactly consistent in the existing code, so not really a problem.

Merge when/if Travis is happy.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobInvocationController.java
JoosjeBoon: Agree, done.

PR #1374, opened 2016-05-31
If trigger file for monitor job is a properties file use it for replacement properties

### LINKED ISSUE #1369: If trigger file for monitor job is a properties file use it for replacement properties
Right now, by convention, DataCleaner monitor uses a properties file with the same name as the job it runs to get its replacement properties to run the job with from, when a hot folder trigger is defined.

If the configured hot folder trigger is a properties file however it makes more sense to use that file to get the replacement properties from. This makes it possible to trigger multiple jobs at once with the same replacement properties.

This is a follow up to/refinement of #1362.

arjansh:
Fixes #1369.

If the configured hot folder trigger is a properties file it makes sense to use that file to get the replacement properties from. This makes it possible to trigger multiple jobs at once with the same replacement properties.


ClaudiaPHI:
LGTM!

LosD:
LGTMtoo!

PR #1380, opened 2016-06-07
Bug/1339 remove output col from input col when the compon deleted

### LINKED ISSUE #1339: Analyzer output stream columns are not always updated properly
When columns are removed from a job by deleting a transformer somewhere, then the input and output columns of analyzers (at least the dup detection analyzer is affected) are not properly updated, leading to inconsistent or incorrect job configurations.

Reproduction steps:
* Use the quick analysis wizard for duplicate detection to generate a dup detection for the Customers database.
* Add a CSV and an XLSX writer to the job, one writing the duplicates output stream, the other writing the uniques output stream.
* Remove the email transformer from the job
* Now open the duplicate detection analyzer set up. You will see the email (standardized) and domain (standardized) column are still there. That's incorrect.
* Remove the domain(standardized) columns then close the dup detection analyzer
* Open one of the writer components. You'll notice that the email (standardized) column is still there on both sides, but the domain (standardized) column is gone from the left-hand side. Also notice that all columns below the domain (standardized) have shifted place with one column. 

Expected behavior is that when I remove columns, they are removed throughout the entire following flow automatically, and that this happens without affecting the mapping of other columns.

ClaudiaPHI:
Fixes #1339  

Removed the transformers output columns from the flow and all its
occurrences.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleInputColumnsPropertyWidget.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleInputColumnsPropertyWidget.java
@@ -48,6 +48,7 @@ import org.datacleaner.actions.ReorderColumnsActionListener;
 import org.datacleaner.api.ExpressionBasedInputColumn;
 import org.datacleaner.api.InputColumn;
 import org.datacleaner.data.MutableInputColumn;
+import org.datacleaner.data.TransformedInputColumn;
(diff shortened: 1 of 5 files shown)


LosD:
We should probably mention that this is currently blocked due to issues, so merges it in its current state :)

arjansh:
No longer needed as #1482 fixes the issue this PR aims to fix.

PR #1381, opened 2016-06-07
Feature/1379 news channel

### LINKED ISSUE #1379: News Channel for DataCloud
Showing the News Channel in DataCleaner

rposkocil:
fixes #1379


desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
```diff
@@ -188,6 +188,10 @@
     public static final String DICTIONARY_TEXTFILE_IMAGEPATH = "images/model/dictionary_textfile.png";
     public static final String DICTIONARY_DATASTORE_IMAGEPATH = "images/model/dictionary_datastore.png";
 
+    public static final String NEWS_CHANNEL = "images/news/news_channel_16.png";
+    public static final String NEWS_CHANNEL_ORANGE = "images/news/news_channel_28.png";
```
LosD: Maybe not the best name: It's more important what it is used for (or better, what it represents), than what it looks like. Then we can change the image later without the constant name not making sense anymore, or having to change names.

api/src/main/java/org/datacleaner/api/ShortNews.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public static class Item implements java.io.Serializable {
+
+        private String name;
+        private String title;
+        private Date dateCreated;
+        private String message;
+
+        public Item() {
```
LosD: Any reason this isn't a immutable class? Do we need to mutability?

desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
LosD: (I know we do something similar in other places, but we're trying to move away from that)

desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
```diff
@@ -188,6 +188,10 @@
...
+    public static final String NEWS_CHANNEL_HIGHLIGHTED = "images/news/news_channel_20.png";
```
LosD: Also a bit of a problem with the filenames as well: From the name alone, I would think that they represented exactly the same image in 3 different sizes, but that is not the case.

That reminds me, remember to put the high-res/scalable originals in the NCIM/graphics project so we can reuse them for other things or if we need other sizes.

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -19,10 +19,7 @@
  */
 package org.datacleaner.windows;
 
-import java.awt.BorderLayout;
-import java.awt.CardLayout;
-import java.awt.Cursor;
-import java.awt.Image;
+import java.awt.*;
```
LosD: This should have been avoided by the common Eclise/IDEA formatter setup?

engine/rest-client/src/main/java/org/datacleaner/restclient/NewsChannelRESTClient.java
```diff
@@ -0,0 +1,47 @@
...
+        String response = call(count);
+        ShortNews news = Serializator.shortNewsList(response);
+        return news.getNewsItems();
+    }
+
+    private String call(int count) throws RestrictedFunctionalityException {
+        String response = restClient.getResponse(RESTClient.HttpMethod.GET, url + "?count=" + count, "");
+        return response;
+    }
+}
```
LosD: Missing end-of-line

LosD:
LGTM except for at few issues, mostly naming.

desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
rposkocil: I agree with the bad filenames and I will put the images to graphic project. But a graphic expert explained to me that a resizable icons reduce readability icons by blurring. In the end she have created the icons in a selected size and in better quality than I can do. :-)

LosD: Okay, that's fine, we'll just have to make new ones, then. I guess they should be reasonably easy to copy anyway.

Just a clarification of the resizable icons, there's two main things to icon sizes:
- Simple scalability; Are they using a scalable format like SVG or using something sufficiently high resolution that scaling down shouldn't afffect the quality too much. First one is usually preferred (but we haven't been good at using them anyway).
- The complexity of the image: You can't just scale down something with a lot of small details. Usually what I see is icon designers doing 3 (or more) different versions in varying degrees of complexity. Those would then be scalable separately. I'm guessing that's what your graphics guru was referring to.

Making this much more complex is the issue of DPI: Even though your phone (probably) has an insanely high resolution, you can't just use the complex icon, as it would still look smudgy in small sizes. Currently that isn't an issue for DataCleaner, though, as we don't support DPI scaling in any case (which is actually becoming a problem on new laptops, as they also have insane resolutions, so everything looks _super_ small.)

Whoops, that was a bit of a sidestep... :)

api/src/main/java/org/datacleaner/api/ShortNews.java
rposkocil: Fixed with private setters.

LosD: Oh, that wasn't really what I meant, I meant final fields and no setters. But I guess that might be an issue for the JSON serialization? (I believe that Jackson can handle it, though)

LosD: We discussed this a bit offline, and while it's fixable, it's probably worse than just keeping it mutable.

LosD:
LGTM! Merge when Travis is happy

PR #1387, opened 2016-06-09
userinfo now returns also user ID

jakubneubauer:
user ID is sometimes useful (some ws/... API require user ID in parameters) => DataCloud will return it in userinfo object.


ClaudiaPHI:
LGTM!

jakubneubauer:
Closed this PR because it makes the /.../userinfo REST service backward uncompatible.

PR #1390, opened 2016-06-13
#1389 - datacloud json - ignore unknown properties

### LINKED ISSUE #1389: DataCleaner must ignore unknown properties in json
DataCloud api can contain more properties in json, than DataCleaner (old version) expects.
Old version of DataCleaner must work after change of datacloud api.

khouzvicka:
fixes #1389


jakubneubauer:
yes, that is what we want.

jakubneubauer:
lgtm

PR #1394, opened 2016-06-15
Feature/1388 disable dc transformers mark

### LINKED ISSUE #1388: Disable datacloud transformer - different color and info
Disable datacloud transformers will have gray color and olso information in configuration dialog

khouzvicka:
fixes #1388 
Disabled transformers will be gray.

bestmethod

In configuration of transformer will be information about it.
disableinfo
Also is there backward compatibility. Rest client is sending header with dataCleaner version. 

Newest versions of datacleaner are able to ignore new fields in json, but old no.


LosD:
How come you are making your own desaturation function instead of using one of the filters?

LosD:
ColorConvertOp may match what you want better: http://stackoverflow.com/questions/14358499/how-to-turn-imageicon-to-gray-on-swing

khouzvicka:
@LosD  My method does much better results. You can see both results in this pull request.

LosD:
I don't somewhat really agree that it's prettier, but it doesn't look disabled.

LosD:
Anyway, if the performance is good, I'm okay with it. But the dropdown is already becoming laggy, so it needs to be performant (or only run once, not on every change)

LosD:
I'd try ColorConvertOp first, it seems to look much like your. If we can avoid home-brewed code, it's certainly preferable.

LosD:
(In general, making own algorithms should only be done if all else fails)

khouzvicka:
@LosD 
ColorConvertOp version of icons.
colorconvertop

LosD:
If it's okay for you, let's keep that. Otherwise let's use yours (if it doesn't introduce lag)

khouzvicka:
@LosD Icon is transform max two times.
1) When user click to category in tree.
2) When user adds transformer to workflow, but only one time in app life.

@manuelvandenberg @jakubneubauer 
What is the right version is in your hands. Please choose one.

LosD:
Okay, as long as it isn't recalculated e.g. while searching or every time the branch is reopened, my only objection is the homebrewed filtering, which isn't valid if none of the exitsting ones gives us what we want :)

engine/core/src/main/java/org/datacleaner/VersionComparator.java
```diff
@@ -36,32 +36,37 @@ public int compare(String o1, String o2) {
...
         for (int i = 0; i < Math.min(o1Split.length, o2Split.length); i++) {
-            int snapshotCounter = 0;
+            int specialEndCounter = 0;
             Integer o1Part;
-            if (o1Split[i].endsWith("-SNAPSHOT")) {
-                snapshotCounter++;
-                o1Part = Integer.parseInt(o1Split[i].substring(0, o1Split[i].lastIndexOf("-SNAPSHOT")));
+            if (o1Split[i].contains("-")) {
+                specialEndCounter++;
+                o1Part = Integer.parseInt(o1Split[i].substring(0, o1Split[i].lastIndexOf("-")));
```
LosD: Is anything else depending on the old behavior?

engine/core/src/test/java/org/datacleaner/VersionComparatorTest.java
```diff
@@ -65,4 +66,21 @@ public void testSnapshotAndRelease() {
...
 
+    @Test
+    public void testRC() {
+        VersionComparator comparator = new VersionComparator();
+
+        List<String> versions1 = Arrays.asList("5.0-RC1", "5.0", "5.0.1.5","4.0", "5.0-SNAPSHOT", "4.0.1-SNAPSHOT","5.0.1-beta","5.0.1-alfa", "5.0.1-RC2", "5.0.0-SNAPSHOT", "5.0.1-SNAPSHOT", "5.0.1-RC1");
+        List<String> versions2 = new ArrayList<>(versions1);
+
+        Collections.sort(versions1, comparator);
+        assertEquals("[4.0, 4.0.1-SNAPSHOT, 5.0-RC1, 5.0-SNAPSHOT, 5.0, 5.0.0-SNAPSHOT, 5.0.1-alfa, 5.0.1-beta, 5.0.1-RC1, 5.0.1-RC2, 5.0.1-SNAPSHOT, 5.0.1.5]", versions1.toString());
```
LosD: This looks perfect :)

khouzvicka:
@LosD @manuelvandenberg @jakubneubauer 
And other question is about the text is configuration.
Is it ok with email contact?

LosD:
Good question. I'd think that a simple link to the website would provide them the information they need. Maybe even to the DataCloud information page.

desktop/api/src/main/java/org/datacleaner/panels/AbstractComponentBuilderPanel.java
```diff
@@ -163,6 +165,18 @@ private final void init() {
...
 
+    private void addInformationPanelAboutDisable() {
+        if (_descriptor instanceof Allowable && !((Allowable) _descriptor).isAllowed()) {
+            final ImageIcon icon = imageManager.getImageIcon(IconUtils.STATUS_INFO, IconUtils.ICON_SIZE_TASK_PANE);
+            DCPanel panel = new DCPanel();
+            DCHtmlBox dcHtmlBox = new DCHtmlBox(
+                    "<p>This service is not currently available for your DataCloud account.</p>"
+                            + "<p> Please contact 'support@datacleaner.org' to obtain access.</p>");
+            panel.add(dcHtmlBox);
+            addTaskPane(icon, "Component is disabled", panel, true);
```
LosD: We have a standard for these panels. Please don't make everything to HTML boxes.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -443,6 +481,15 @@ static private String getURLForCreation(String tenant, ComponentDescriptor<?> de
...
     }
 
+    private boolean isComponentEnabled(String componentName){
+        try {
+            canCall(componentName);
+        }catch (RuntimeException e){
+            return false;
+        }
+        return true;
+    }
```
LosD: It probably doesn't matter in this case, but exceptions as part of normal flow is usual very slow (or, at least it used to be, not sure if they optimized it), and in general frowned upon. Isn't there a more proper way to interrogate the current enabled state?

LosD:
Mostly LGTM, but see comments

khouzvicka:
I changed the info in configuration to:
newdisableinforamtion

LosD:
Makes sense, and (somewhat) justifies the use of HTML

engine/core/src/main/java/org/datacleaner/VersionComparator.java
khouzvicka: It is ok.
Comparator was used only for finding newest version between two values. Better code with new suffixes (alfa, RC1) don't change behavior.

LosD: Superb

jakubneubauer:
From Karel's 50 shades of grey I vote for the first one (screenshot from description). The "disabled" is really disabled but in my opinion too much - it is almost unreadable. The simple conversion to greyscale looks nice, but doesn't evoke disabled behavior. The Karel's proposal is in the middle - tries to be a balance between the two.

manuelvandenberg:
to me the ColorConvertOp'd version looks ok, as does the first one.  the 'disabled' is too disabled, i agree with @jakubneubauer on that one.
Generally, less lines of code is better.

LosD:
This should do the same, but with much simpler code... (I thought it would be faster too, but looking at the image filter code, there is absolutely no blitting or anything else to accelerate filters, so the only gain is probably a bit prettier and simple code):
```Java
    public static class DisabledFilter extends RGBImageFilter {
        @Override
        public int filterRGB(final int x, final int y, final int rgb) {
            int red = (rgb >> 16) & 0xff;
            int green = (rgb >> 8) & 0xff;
            int blue = (rgb & 0xff);

            final int max = Math.max(Math.max(red, green), blue);
            final int min = Math.min(Math.min(red, green), blue);
            final int gray = brighter((max + min) / 2, 40);

            return (rgb & 0xff000000) | (gray << 16) | (gray << 8) | gray;
        }
    }

    public static ImageIcon getDesaturationIcon(ImageIcon inputIcon) {
        DisabledFilter filter = new DisabledFilter();
        ImageProducer prod = new FilteredImageSource(inputIcon.getImage().getSource(), filter);
        return new ImageIcon(Toolkit.getDefaultToolkit().createImage(prod));
    }

    private static int brighter(int rgb, int percent) {
        double q = 255 * percent * 0.01;
        double k = (255-q)/255;
        return (int) (k * rgb + q);
    }
```
The original array-based `brighter` can easily be changed to use the new one:
```Java
    private static void brighter(int[] rgb, int percent) {
        rgb[0] = brighter(rgb[0], percent);
        rgb[1] = brighter(rgb[1], percent);
        rgb[2] = brighter(rgb[2], percent);
    }
```

I can't test it though, I get failures in downloadDescriptors, but that's probably just because I don't have the development DataCloud server. :)

khouzvicka:
@LosD  Thanks, your code was very good and it is there.
Could you do merge pull request? Please.

LosD:

PR #1395, opened 2016-06-15
Qualified autowire for remote component config 

jakubneubauer:
Remote component config has now qualified autowire. Needed by DataCloud so that a "composite" config bean can be defined in application context, refering to other component config beans.

BTW, the default "empty" config already has a name "published-components", this is why I picked the qualifier of that name - to be backward compatible with the current bean name.


LosD:
LGTM!

PR #1397, opened 2016-06-17
Fix in swagger json

jakubneubauer:
swagger json generation was not working for controllers that had annotations in ancestor class


LosD:
LGTM! Merge when Travis is happy.

PR #1398, opened 2016-06-17
Remote component API - output columns info in processing response

jakubneubauer:
Optional parameter now allows to obtain column info in the response of transformer result.

This PR solves this issue: the remote components API - the method to process rows - returns processed rows as a "table" of data. If called by DataCleaner engine, no problem since it uses another API method "getOutputColumns" to get info about output columns.

But when calling the API directly, it could be good to obtain the column info directly in the "processing" call response. This is why I made it possible through an optional request boolean flag. (Default is "false").

Example output with columns info: (BTW, the columns info format is reused from the existing "getOutputColumns" response)
```
{
  "rows": [ [ ["hello", "world"] ], [ ["Hi", "Jane"] ] ],
  "columns": [
    {
      "name": "Greeting",
      "type": "java.lang.String",
      "schema": {
        "type": "string"
      }
    },
    {
      "name": "Object name",
      "type": "java.lang.String",
      "schema": {
        "type": "string"
      }
    },
  ]
}
```

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
@@ -241,8 +241,8 @@ public class ComponentControllerV1 {
     public ProcessStatelessOutput processStateless(
             @PathVariable(PARAMETER_NAME_TENANT) final String tenant,
             @PathVariable(PARAMETER_NAME_NAME) final String name,
-            @RequestParam(value = PARAMETER_NAME_OUTPUT_STYLE,required = false, defaultValue = PARAMETER_VALUE_OUTPUT_STYLE_TABULAR) String outputStyle,
-            @RequestParam(value = PARAMETER_NAME_COLUMNS,required = false, defaultValue = "false") boolean outputColumnsInfo,
+            @RequestParam(value = PARAMETER_NAME_OUTPUT_STYLE, required = false, defaultValue = PARAMETER_VALUE_OUTPUT_STYLE_TABULAR) String outputStyle,
+            @RequestParam(value = PARAMETER_NAME_COLUMNS, required = false, defaultValue = "false") boolean outputColumnsInfo,
             @RequestBody final ProcessStatelessInput processStatelessInput) {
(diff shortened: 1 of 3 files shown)


LosD:
Which issue does this apply to?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ComponentControllerV1.java
```diff
@@ -249,9 +238,11 @@ public OutputColumns getOutputColumns(@PathVariable(PARAMETER_NAME_TENANT) final
      */
     @ResponseBody
     @RequestMapping(value = "/{name}", method = RequestMethod.PUT, consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
-    public ProcessStatelessOutput processStateless(@PathVariable(PARAMETER_NAME_TENANT) final String tenant,
+    public ProcessStatelessOutput processStateless(
+            @PathVariable(PARAMETER_NAME_TENANT) final String tenant,
             @PathVariable(PARAMETER_NAME_NAME) final String name,
-            @RequestParam(value = PARAMETER_NAME_OUTPUT_STYLE, required = false, defaultValue = PARAMETER_VALUE_OUTPUT_STYLE_TABULAR) String outputStyle,
+            @RequestParam(value = PARAMETER_NAME_OUTPUT_STYLE,required = false, defaultValue = PARAMETER_VALUE_OUTPUT_STYLE_TABULAR) String outputStyle,
+            @RequestParam(value = PARAMETER_NAME_COLUMNS,required = false, defaultValue = "false") boolean outputColumnsInfo,
```
LosD: Missing spaces before `required` in these two rows.

LosD:
Code LGTM, except a minor formatting issue.

jakubneubauer:
fixed the formatting issue and added more description for you to imagine the value of this PR ;-)

jakubneubauer:
BTW, the idea was already old, mentioned on one SprintReview in HI. Now there is a potential client that wants to use our API directly, so I quickly implemented this enhancement...

LosD:
I appreciate the extra description, but let's make a proper GitHub issue before fixing it next time :)

Merging.

PR #1399, opened 2016-06-20
Feature/1396 additional component metadata annot

### LINKED ISSUE #1396: Additional metadata about remote components
As a REST API user, I want to receive more information about the available services (remote components) so that I use this in my own application.

#### Description
The REST service that returns the available services should be expanded with the following metadata/ info:
- service category: eg. category: enrich
- service metadata which gives information about the entities for which the service can be used. eg.: entityType: person/ organizations/ address and eg. country: NL/ DE etc.

#### Rationale
Systems that need to integrate with the cloud API need to implement a lot of logic to sort these service out. However, since the cloud already has this knowledge about its own services it seems more logical for them to provide it to the consumers of the API.

#### Business value
A more REST based API that could be used not only by HI specific applications but also by customer specific clients. We become more flexible in our applications.

khouzvicka:
fixes #1396

### DIFF
--- a/engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
+++ b/engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
@@ -70,11 +70,6 @@ public class Serializator {
             @Override
             public void setupModule(final SetupContext context) {
                 context.setMixInAnnotations(ComponentScope.ServiceType.class, ComponentScopeServiceTypeMixin.class);
-            }
-        });
-        objectMapper.registerModule(new SimpleModule(){
-            @Override
-            public void setupModule(final SetupContext context) {
                 context.setMixInAnnotations(ComponentScope.EntityType.class, ComponentScopeEntityTypeMixin.class);
             }
         });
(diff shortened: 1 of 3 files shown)


khouzvicka:
@LosD Please for last code review and merge. I think, that we already discussed code in issue.

engine/rest-client/src/main/java/org/datacleaner/restclient/Serializator.java
```diff
@@ -59,6 +64,20 @@
...
+            public void setupModule(final SetupContext context) {
+                context.setMixInAnnotations(ComponentScope.ServiceType.class, ComponentScopeServiceTypeMixin.class);
+            }
+        });
+        objectMapper.registerModule(new SimpleModule(){
+            @Override
+            public void setupModule(final SetupContext context) {
+                context.setMixInAnnotations(ComponentScope.EntityType.class, ComponentScopeEntityTypeMixin.class);
+            }
+        });
```
LosD: I think you should be able to do this using a single module with two calls to `context.setMixInAnnotations`. Not sure about that, and not really an issue, just an alternative :)

LosD:
LGTM! Merge if you prefer the two-module approach

khouzvicka:
Ok. Now it is only one modul. I will wait for travis and after that I will merge it.

LosD:
:+1: :)

LosD:
Travis error seems unrelated.

khouzvicka:
Travis is ok.

PR #1402, opened 2016-06-20
Fix most of run n records cases (but there can still be slip-ups)

### LINKED ISSUE #1371: When using Execute with option 'Run first N records' all filters in the job are ignored
Originally submitted by @jacques-hi:

If a job contains one ore more filters, these are ignored when the job is executed with option 'Run first N records'.

LosD:
Unfiltered Analyzers might mean that too much data will be loaded.
Unfiltered transformers with results will _not_ be filtered.

Fixes #1371

I'm thinking I'd like to add a few test cases before merging.


ClaudiaPHI:
LGTM!

LosD:
Fixed existing unit tests.

ClaudiaPHI:
Ready to merge ?

LosD:
I'd say so. You want the honors?

PR #1407, opened 2016-06-22
#1406 category classes in API.

### LINKED ISSUE #1406: Category classes are in datacleaner core
A lot of Datacloud components must have dependency to datacleaner-core only for the category class.
I would like to move this classes to API. They will be in same package.

khouzvicka:
fixes #1406 
A lot of Datacloud components must have dependency to datacleaner-core only for the category class.
I would like to move this classes to API. They will be in same package.


LosD:
LGTM. I don't think there should be any API breakage when using the package name is unchanged, since a user would depend on the api module to use the engine module anyway.

Merge when Travis is happy.

PR #1412, opened 2016-06-24
Feature/1411 right panel refactor

### LINKED ISSUE #1411: Right information panel refactoring
The right information panel behaves wierd now. Other panels opens behind the actual opened one. It should be refactored to a new component.

There will be better to create one component with card layout instead of 3 components which can overlap each other.

rposkocil:
fixes #1411


jakubneubauer:
Very nice code reduction and duplicity removal. Small idea: what about new RightInformationPanel method like 'toggleWindow(String name)' which closes window if it is shown, otherwise opens it?

jakubneubauer:
LGTM

LosD:
Yeah, that was a bad merge, it broke the display of the which edition the user is running.

PR #1422, opened 2016-07-06
Add a description for GenerateId offset

### LINKED ISSUE #1361: Generate ID transformer, offset is confusing for user, let's use "start with" instead
I was really confused with this. Please fix it for me. :-)

LosD:
The offset field could be confusing, this describes it a little.

Fixes #1361


wvholland:
+1

PR #1426, opened 2016-07-07
Bug/1424 null check filter considers empty strings as null

### LINKED ISSUE #1424: NullCheckFilter: no rows processed in large oracle database
In a simple job consisting of "oracle datastore", "null check filter" and "csv writer (or value distribution)" where the table contains "many" rows (cca 200000), the results are only produced when both branches of NullCheckFilter are used. 

When I use only "NOT_NULL" then there is no result. 

I tried that on the same remote oracle database but on a different table with only a few records (cca 50) and it worked fine. I also tried that on my/local postgres datastore with 300000 rows and it worked fine as well. So, it seems the problem is related to oracle database table with many records. 

null-filter-issue-01

null-filter-issue-02

null-filter-issue-04

jhorcicka:
Fixes #1424. 

Problem with "optimized sql condition" in NullCheckFilter with "Consider empty strings as null" checked. 

Condition ```<Operator.DIFFERENT_FROM, "">``` was replaced by ```<Operator.LIKE, "%">```.

### DIFF
--- a/components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
+++ b/components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
@@ -132,7 +132,7 @@ public class NullCheckFilter implements QueryOptimizedFilter<NullCheckFilter.Nul
                 }
                 q.where(column, OperatorType.DIFFERENT_FROM, null);
                 if (considerEmptyStringAsNull && col.getDataType() == String.class) {
-                    q.where(column, OperatorType.LIKE, "%");
+                    q.where(column, OperatorType.DIFFERENT_FROM, "");
                 }
             }
         } else {
--- a/components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
+++ b/components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
@@ -25,7 +25,6 @@ import java.util.List;
 import javax.inject.Named;
 
 import org.apache.metamodel.query.FilterItem;
-import org.apache.metamodel.query.FunctionType;
 import org.apache.metamodel.query.OperatorType;
 import org.apache.metamodel.query.Query;
 import org.apache.metamodel.query.SelectItem;
@@ -133,7 +132,7 @@ public class NullCheckFilter implements QueryOptimizedFilter<NullCheckFilter.Nul
                 }
                 q.where(column, OperatorType.DIFFERENT_FROM, null);
                 if (considerEmptyStringAsNull && col.getDataType() == String.class) {
-                    q.having(FunctionType.TO_BOOLEAN, column, OperatorType.EQUALS_TO, "TRUE");
+                    q.where(column, OperatorType.LIKE, "%");
                 }
             }
         } else {
--- a/components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
+++ b/components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
@@ -48,21 +48,21 @@ import org.datacleaner.components.categories.FilterCategory;
 @Distributed(true)
 public class NullCheckFilter implements QueryOptimizedFilter<NullCheckFilter.NullCheckCategory>, HasLabelAdvice {
 
-    public static enum NullCheckCategory {
+    public enum NullCheckCategory {
         @Alias("INVALID")
         NULL,
 
         @Alias("VALID")
-        NOT_NULL;
+        NOT_NULL
     }
 
-    public static enum EvaluationMode implements HasName {
-        ALL_FIELDS("When all fields are NULL, the record is considered NULL"), ANY_FIELD(
-                "When any field is NULL, the record is considered NULL");
+    public enum EvaluationMode implements HasName {
+        ALL_FIELDS("When all fields are NULL, the record is considered NULL"),
+        ANY_FIELD("When any field is NULL, the record is considered NULL");
 
         private final String _name;
 
-        private EvaluationMode(String name) {
+        EvaluationMode(String name) {
             _name = name;
         }
 
@@ -82,7 +82,7 @@ public class NullCheckFilter implements QueryOptimizedFilter<NullCheckFilter.Nul
     boolean considerEmptyStringAsNull = false;
 
     @Configured("Evaluation mode")
-    EvaluationMode evaluationMode = EvaluationMode.ANY_FIELD;;
+    EvaluationMode evaluationMode = EvaluationMode.ANY_FIELD;
 
     public NullCheckFilter() {
     }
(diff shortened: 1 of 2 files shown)


components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
```diff
@@ -132,12 +132,12 @@ public Query optimizeQuery(Query q, NullCheckCategory category) {
                 }
                 q.where(column, OperatorType.DIFFERENT_FROM, null);
                 if (considerEmptyStringAsNull && col.getDataType() == String.class) {
-                    q.where(column, OperatorType.DIFFERENT_FROM, "");
+                    q.where(column, OperatorType.LIKE, "%");
```
kaspersorensen: As far as I know (I might be wrong) then '%' will also match the empty string, which would make this wrong?

jhorcicka: Ah, you are actually right. In Oracle when I set column to empty string ('') then null seems to be assigned. In this case my updated works but generally it does not. :-(

LosD: Wait, what? That sounds extremely weird. While the Oracle database maybe isn't the most standards compliant of the bunch, I doubt it would make empty strings NULL unless something specific was put in the column definition.

Could that be a bug in DataCleaner/MetaModel?

LosD: We had an offline investigation, and the reason was that the column type was VARCHAR2 (note the *2*). The main feature of that column is that null and empty strings are treated equally.

components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
```diff
@@ -132,12 +133,12 @@ public Query optimizeQuery(Query q, NullCheckCategory category) {
                 }
                 q.where(column, OperatorType.DIFFERENT_FROM, null);
                 if (considerEmptyStringAsNull && col.getDataType() == String.class) {
-                    q.where(column, OperatorType.DIFFERENT_FROM, "");
+                    q.having(FunctionType.TO_BOOLEAN, column, OperatorType.EQUALS_TO, "TRUE");
```
jhorcicka: This seems to be working. What do you think about it (@LosD, @kaspersorensen)? I think something like "strlen(column) > 0" would be much more readable but I did not find that function anywhere.

kaspersorensen: I doubt this will work. The `HAVING` clause is meant to be a "where clause for aggregations" and that's not what you have here. I think it's a mistake that MetaModel even allows it, but that has to do with the common superclass "FunctionType" of both "AggregateFunction" and "ScalarFunction". It should only allow using a AggregateFunction here, but you're passing in a ScalarFunction (TO_BOOLEAN).

jhorcicka: You are right. It does not work. Now it even gives me an exception when run against oracle db. I don't know what I tested yesterday that it worked.

components/basic-filters/src/main/java/org/datacleaner/beans/filter/NullCheckFilter.java
jhorcicka: I've just tried CHAR instead of VARCHAR2(...) and the problem remains.

LosD: Isn't CHAR always padded? So assigning '' to fx. CHAR(5) would actually be '     '?

jhorcicka: When I try CLOB then variant with both, "NOT_NULL" and "NULL", works fine and variant only with "NOT_NULL" gives me 
```
--- DataCleaner progress information user-log ---
11:30:01 INFO: Job begin
11:30:01 INFO: Starting component 'MESSAGE is null?'
11:30:01 INFO: Starting component 'Value distribution (ID) (MESSAGE is null?=NOT_NULL)'
11:30:01 INFO: Starting component 'Value distribution (MESSAGE) (MESSAGE is null?=NOT_NULL)'
11:30:01ERROR: An error occurred in the analysis job!

Could not execute query: ORA-00932: inconsistent datatypes: expected - got CLOB


org.apache.metamodel.MetaModelException: Could not execute query: ORA-00932: inconsistent datatypes: expected - got CLOB

	at org.apache.metamodel.jdbc.JdbcUtils.wrapException(JdbcUtils.java:70)
	at org.apache.metamodel.jdbc.JdbcDataContext.executeQuery(JdbcDataContext.java:511)
	at org.datacleaner.job.runner.SourceTableRowProcessingPublisher.processRowsInternal(SourceTableRowProcessingPublisher.java:225)
	at org.datacleaner.job.runner.AbstractRowProcessingPublisher.processRows(AbstractRowProcessingPublisher.java:208)
	at org.datacleaner.job.tasks.RunRowProcessingPublisherTask.execute(RunRowProcessingPublisherTask.java:43)
	at org.datacleaner.job.concurrent.TaskRunnable.run(TaskRunnable.java:61)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

jhorcicka: @LosD & CHAR(5) '' => '_____': I would think so too. It might be database/editor/server configuration issue but it behaves as I described (I've just rechecked).

LosD: Looks like Oracle need quite a bit of special treatment.

jhorcicka: Another observation. With table like this:  
messages_varchar2

```
SELECT "MESSAGES_VARCHAR2"."ID", "MESSAGES_VARCHAR2"."MESSAGE" 
FROM DATACLEANER."MESSAGES_VARCHAR2" 
WHERE "MESSAGES_VARCHAR2"."ID" IS NOT NULL AND "MESSAGES_VARCHAR2"."MESSAGE" IS NOT NULL
```

returns expected 1. row. 

However, when AND "MESSAGES_VARCHAR2"."MESSAGE" <> ''' is added:
```
SELECT "MESSAGES_VARCHAR2"."ID", "MESSAGES_VARCHAR2"."MESSAGE" 
FROM DATACLEANER."MESSAGES_VARCHAR2" 
WHERE "MESSAGES_VARCHAR2"."ID" IS NOT NULL AND "MESSAGES_VARCHAR2"."MESSAGE" IS NOT NULL AND "MESSAGES_VARCHAR2"."MESSAGE" <> ''
```
It returns nothing.

jhorcicka: Not confirmed hypothesis is that '' is in Oracle always treated as null, which means that COLUMN != '' is actually COLUMN != NULL and all comparisons against NULL are false. So, the way is to remove empty string from the query in case of Oracle and leave only IS/NOT NULL.

LosD:
I guess we can close this, or is there still an issue besides Oracle being weird?

jhorcicka:
I think we can. The problem is (and the solution will be) in MM.

PR #1427, opened 2016-07-07
bug/1423 ValueDistributionAnalyzer, unconfigured output, execution enabled

### LINKED ISSUE #1423: NullCheckFilter: out of bounds exception
In a simple job with "orderdb", "null check filter" and "value distribution" I get "java.lang.IndexOutOfBoundsException: Index: 0, Size: 0". 

null-filter-oobe

jhorcicka:
Fixes #1423  

Original reason for this fix was an exception while running job with unconfigured input columns in ValueDistributionAnalyzer. The fix is general, though. Every exception in AnalysisJobBuilderWindowImpl.updateStatusLabel() makes the job un-executable now.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
@@ -370,7 +370,7 @@ public final class AnalysisJobBuilderWindowImpl extends AbstractWindow implement
     public void open() {
         super.open();
     }
-
+    
     @Override
     public void changePanel(AnalysisWindowPanelType panel) {
         if (_datastore == null) {
@@ -391,7 +391,7 @@ public final class AnalysisJobBuilderWindowImpl extends AbstractWindow implement
     /**
      * Gets whether or not the datastore has been set in this window (ie. if the
      * tree is showing a datastore).
-     *
+     * 
      * @return true if a datastore is set.
      */
     @Override
@@ -401,7 +401,7 @@ public final class AnalysisJobBuilderWindowImpl extends AbstractWindow implement
 
     /**
      * Initializes the window to use a particular datastore in the schema tree.
-     *
+     * 
      * @param datastore
      */
     @Override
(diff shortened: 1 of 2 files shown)


desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -505,10 +512,10 @@ public void updateStatusLabel() {
                     setStatusLabelWarning();
                 }
             } catch (Exception ex) {
+                executeable = false;
```
jhorcicka: This setting to false was in "if" and "else if" branches but not in the "else" branch. I looked into GIT history and it has never been there. But @LosD said that job execution was probably blocked in the past. It means it was achieved in a different way which I don't know. 

And secondly, is there any reason why this last "else" branch did not set "executable" to false? Or is that just something that was forgotten?

LosD: I think it was, but I can't really back it up. Maybe try using an older version (4.5.x or something) and see if I'm right, and if I was, see what has changed.

That executable flag almost _must_ be on purpose. Otherwise why would it be there at all? But I'm not sure why.

LosD: Ah, too early in the morning. Didn't notice it was also being used for other purposes. But it still partially holds: If all branches are supposed to set the executable flag to false, why do in inside the if construct at all?

ClaudiaPHI:
I belive we can fix this instead in ValueDistributionAnalyzer, with a validate method. 

````
    @Validate
    public void validate() {
      if () {
throw new IllegalStateException("No input Column select")
}

}

LosD:
@ClaudiaPHI The issue is that validation _does_ trigger, but the job is still executable.

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
jhorcicka: I tried it in 4.0.11 and in 4.5 and without any configured column in ValueDistributionAnalyzer the job can still be executed. However, there is no error/exception mentioned (just nothing happens / no rows processed).

LosD:
Still not a fan of the executable flag, but that's not anything now, so not really a required fix for this PR. Otherwise LGTM.

There's merge issues, though.

jhorcicka:
Updated from master.

arjansh:
LGTM!

Issue #1433, opened 2016-07-13
Multiple file upload in Monitor (job files)


jhorcicka:
In Monitor it would be nice to have support for multiple files upload. 

In our current project we create something like a batch of jobs that need to be executed consequently in a row and upload of this batch (10+ files) can be really tedious.

PR #1434, opened 2016-07-13
#1433: Multiple job files upload in monitor.

jhorcicka:
Solves #1433. 

In monitor, job file upload now supports multiple selection. 

file_upload

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobsFolderController.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobsFolderController.java
@@ -64,6 +64,7 @@ import com.google.common.net.UrlEscapers;
 public class JobsFolderController {
 
     private static final Logger logger = LoggerFactory.getLogger(JobFileController.class);
+    private static final String STATUS_SUCCESS = "Success";
 
     @Autowired
     TenantContextFactory _contextFactory;
@@ -73,7 +74,7 @@ public class JobsFolderController {
     public List<Map<String, String>> resultsFolderJson(@PathVariable("tenant") String tenant) {
         final TenantContext context = _contextFactory.getContext(tenant);
 
-        final List<Map<String, String>> result = new ArrayList<Map<String, String>>();
+        final List<Map<String, String>> result = new ArrayList<>();
 
         {
             final List<JobIdentifier> jobs = context.getJobs();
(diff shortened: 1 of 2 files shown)


monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobsFolderController.java
```diff
@@ -160,6 +161,29 @@
     }
 
     @RolesAllowed(SecurityRoles.JOB_EDITOR)
+    @RequestMapping(value = "/multiple_upload", method = RequestMethod.POST, produces = MediaType.TEXT_HTML_VALUE, consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
+    public String uploadMultipleAnalysisJobsToFolderHtml(@PathVariable("tenant") final String tenant,
```
jhorcicka: I created a new method with unique URL suffix "multiple_upload" not to change public API.

monitor/ui/src/main/webapp/scheduling.xhtml
```diff
@@ -57,7 +55,7 @@
 						<span class="glyphicon glyphicon-plus" aria-hidden="true" /> <span>Build
 							job</span>
 					</button>
-					<form action="repository/#{user.tenant}/jobs/#{file.name}"
+					<form action="repository/#{user.tenant}/jobs/multiple_upload"
```
jhorcicka: What does "#{file.name}" refer to? The value was (at least in my case) an empty string and I did not find any "File"Bean.java. Does it have any purpose or is/was that just some leftover?

LosD: Good question. If it works without,  it probably was :-)

LosD:
LGTM, but I can't test it right now. If you did a thorough test, merge when Travis is happy, otherwise I can do it later.

jhorcicka:
I tested it on "/scheduling" (job file upload) but I am not sure if the xhtml template code is not used anywhere else. "conf.xml" file upload works fine (single selection). 

I am already using the feature on my localhost so my needs were satisfied. Now it can wait some time before it is merged into the master and another testing is always good. So, if you are willing to test, please do, @LosD. :-) Thanks.

LosD:
Will do! :-)

ClaudiaPHI:
LGTM!

LosD:
Testing seems to work, but the URL looks weird, with a raw Java array reference in there: `http://localhost:8080/DataCleaner-monitor/scheduling?job_upload=Success&job_filename=[Ljava.lang.String;@fc18766`

We might want to get rid of that.

LosD:
image
Picture version :)

LosD:
Since you stopped using it for feedback, maybe just nuke it from the redirect in the multiple_upload endpoint (then you could maybe also check if it exists, and reinstate the filename uploaded if it does).

jhorcicka:
Both, URL parameter job_filename and final success message now contain the list of files. 

file-upload

LosD:
Uh, nice! Let's merge when Travis is happy

PR #1442, opened 2016-07-14
#1441 More flexible DateRangeFilter

### LINKED ISSUE #1441: More flexible DateRangeFilter
Within our current project we need "now - 18 years" date to filter on. This is not currently possible to specify in DateRangeFilter.

jhorcicka:
Resolves #1441. 

**DO NOT MERGE YET**

In DateRangeFilter GUI user can specify "now() plus 1d 2m 3y". Negative values are also acceptable: "0d 0m -18y". 

date-range-filter


ClaudiaPHI:
I have tested it. When I reopen the configuration dialog, the previous setting is deleted.

jhorcicka:
It happens when you edit the text field and then click "Close". If you click (after setting the value) on its (or another) radio button then it works fine, in my case. 

I'll try to fix it.

jhorcicka:
I added "on property change" listener, it works for me. Can you please re-test, @ClaudiaPHI? Thanks.

ClaudiaPHI:
The name (title) of the filter is not updated if the configuration has been changed. 
image

jhorcicka:
I replaced PropertyChangeListener by FocusListener which seems to fix some problems. 

The title bar of the configuration dialog is not changed, though (after closing and reopening it is correctly updated). But this is something that was already there before these changes and it is not something that can be influenced from SingleDatePropertyWidget right now. So, I assume it is OK. If not, let's create a separate issue for this. 

date-range-filter-change

LosD:
Hmmm... Love the functionality, but I'm not a great fan of the UI... I just don't really have a good alternative.

Anyone has any good ideas or opinions? @kaspersorensen @arjansh

kaspersorensen:
I also think the UI is becoming a bit clumsy. I can live with it, because after all it's a pretty rare thing. But if you wanted to make it better, then I could suggest to have the current value presented with a label and then just have a button next to it to change it. The button would then open up a dialog, popup or a overlay-panel or such. In that panel you could have all the bells and whistles, but they would dissappear after setting it to the value you want.

arjansh:
As @kaspersorensen proposes, a button which opens up a dialog seems like a good option here. I would put all options except for the datepicker in such a dialog. Also in the dialog, some extra explanation regarding what today and yesterday exactly mean can be added (when I select "today", is it the beginning of today of the end of today or does that differ depending on whether I choose it for lowest or highest value?).

ClaudiaPHI:
What happened with PR ?

jhorcicka:
This is still in progress. The separate dialog is not there yet.

jhorcicka:
I added a separate dialog where user can specify a new value (with some description). I did not separate datepicker and other elements. In my opinion it is better to have all "changing options" bound together. 

filter-dialog
setting-dialog

engine/core/src/test/java/org/datacleaner/util/convert/ShiftedTodayTest.java
```diff
@@ -0,0 +1,82 @@
...
+        calendar.set(Calendar.MINUTE, 0);
+        calendar.set(Calendar.SECOND, 0);
+        calendar.set(Calendar.MILLISECOND, 0);
+        calendar.add(Calendar.DAY_OF_MONTH, days);
+        calendar.add(Calendar.MONTH, months);
+        calendar.add(Calendar.YEAR, years);
+        
+        return calendar.getTime();
+    }
+}
```
arjansh: No newline at end of file.

desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleDatePropertySettingDialog.java
```diff
@@ -0,0 +1,207 @@
...
+import org.datacleaner.util.convert.NowDate;
+import org.datacleaner.util.convert.ShiftedToday;
+import org.datacleaner.util.convert.TodayDate;
+import org.datacleaner.util.convert.YesterdayDate;
+import org.datacleaner.widgets.Alignment;
+import org.datacleaner.widgets.DescriptionLabel;
+import org.datacleaner.windows.AbstractDialog;
+import org.jdesktop.swingx.JXDatePicker;
+
+public class SingleDatePropertySettingDialog extends AbstractDialog implements ActionListener {
```
arjansh: The serializable class SingleDatePropertySettingDialog does not declare a static final serialVersionUID field of type long.

engine/core/src/main/java/org/datacleaner/util/convert/ShiftedToday.java
```diff
@@ -0,0 +1,89 @@
...
+ * Special date value, created by shifting "today" according to the String input in format: "+1d +2m -3y".
+ */
+public class ShiftedToday extends Date implements ExpressionDate {
+
+    private static final long serialVersionUID = 1L;
+    private final String _input;
+
+    public ShiftedToday(String input) {
+        _input = validateInput(input);
+        int[] counts = parseCounts();
```
arjansh: Final.

engine/core/src/main/java/org/datacleaner/util/convert/ShiftedToday.java
```diff
@@ -0,0 +1,89 @@
...
+
+        Calendar calendar = Calendar.getInstance();
```
arjansh: Also final.

jhorcicka:
Thanks, @arjansh. Your tips were filled.

jhorcicka:
I added "not set" text when the value is uninitialized. 

not-set

arjansh:
I just tested it and I have some reservations:
- In the "Set date value", when I select a date for the "Select: " option, the radio button isn't check automatically, which I would expect. The same goes for the text field next to "Now plus", when I change its value, I would expect the Now plus option to be checked.
- When I check the "Now plus" option, the time shown in the field (in the Date Range dialog) always shows "00:00:00", so when I fill in the value "+1d +0m +0y", it can actually mean plus one and a half day, depending on the current time of the day.
- When I select another value then "Select:", a date is filled in the field (in the Date Range dialog), but it is not determinable whether that date has been explicitly selected using the "Select:" option or the result of one of the other options, maybe, when selecting "Now" for example it should show "Now (%current date%), where %current date% show the exact current date.
- When I set a date value, then close the "Set date value" dialog, then close the "Date range" filter dialog, then open the "Date range" filter dialog again and the open the "Set date value" dialog again, no radio button is checked.
- When I set "Now" in the "Set date value" dialog, the exact current date (including time) is filled in in the field (in the Date Range dialog), when I now close both dialogs and then reopen the "Date range" filter dialog again (at least one second later), the value in the "Date range" filter dialog is before "Now" as what is now now.
- I'm not sure about the name of the "Change" button. Maybe "Select" (maybe ask someone else's opinion on this too).
- I personally don't like the text "Not set! Please, select some date.". Again, maybe ask someone else's opinion on this.

jhorcicka:
After a discussion with @LosD, @arjansh and @Jan-Arie, we decided to:
1) separate "static" and "dynamic" options in GUI. 
2) not show real date value in case of "now", "today", "yesterday" or "today plus". 
3) set the default values to "today" instead of "now". 

static-dynamic

all-labels

LosD:
I'm no fan of "Static" and "Dynamic". OTOH, I don't know if I have a better alternative

jhorcicka:
I believe it was you, @LosD, who wanted the groups. About the labels, I can imagine that this is not the best wording for English speaking users so I am more than open for suggestions. 

Would "Active" or "Relative" be better alternative for "Dynamic"? The first group can be "Particular date" and the label can disappear.

LosD:
I agree with the groups, just not the labels. The reason I don't like static and dynamic is that it's very... techie.

How about "Fixed date", and something like "Resolved at execution" (though I don't like that particularly well either).

components/basic-filters/src/main/java/org/datacleaner/beans/filter/DateRangeFilter.java
```diff
@@ -71,4 +75,50 @@ public Date getLowestValue() {
...
+    @Override
+    public String getSuggestedLabel() {
+        final Date highestValue = getHighestValue();
+        final Date lowestValue = getLowestValue();
+        
+        if (highestValue == null || lowestValue == null) {
+            return null;
+        }
+        
+        final InputColumn column = getColumn();
```
arjansh: InputColumn is a raw type. References to generic type InputColumn<E> should be parameterized.

desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleDatePropertySettingDialog.java
```diff
@@ -0,0 +1,293 @@
...
+
+        _datePicker = createDatePicker();
+        _todayPlusTextField = createTodayPlusTextField();
+    }
+    
+    private JButton createCloseButton() {
+        final JButton closeButton = WidgetFactory.createPrimaryButton("Close", IconUtils.ACTION_SAVE_BRIGHT);
+        closeButton .addActionListener(e -> {
+            dispose();
+        });
```
arjansh: Why not `closeButton.addActionListener(e -> dispose()); ` instead?

jhorcicka:
Corrected, @arjansh.

desktop/ui/src/main/java/org/datacleaner/widgets/properties/SingleDatePropertySettingDialog.java
LosD: Also some weirdness with the space after `closeButton`

arjansh:
LGTM!

PR #1447, opened 2016-07-15
Allow POST for jobs without specific content type

### LINKED ISSUE #1425: Monitor job trigger doesn't accept POST for normal jobs.
It seems that we did some change that made POST special, so that that trigger doesn't work if it is not POST, at least not when using the old, unconfigured way (it seems that POST _requires_ JSON now, looking at the code). This doesn't really fly, we need POST to behave in the old way. Both because it is expected to work as always, and because it is the only really allowed method according to the HTTP spec; most of the other HTTP methods are idempotent, and starting a new job is not an idempotent request as new executions will start every time the server receives the request.

LosD:
This makes sure that jobs without a content type and/or body can still use POST to trigger a job execution.

Fixes #1425

We'd probably like this to be tested pretty thoroughly, since it has already slipped through the cracks once.

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobTriggeringController.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobTriggeringController.java
@@ -58,8 +58,7 @@ public class JobTriggeringController {
     @Autowired
     SchedulingService _schedulingService;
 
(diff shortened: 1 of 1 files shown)


ClaudiaPHI:
I will take a look at it.

arjansh:
LGTM!

PR #1449, opened 2016-07-15
Fix a race condition in equals filter

### LINKED ISSUE #1448: Equals filter may fail when comparing columns
We have a race in the equals filter for comparing columns, so that 5-10% of the compares gets it wrong.

Related to updating the fields in the filter method, which means another thread might update the field before re-read.

LosD:
Do not update the _operand field in the filter method.

Fixes #1448


ClaudiaPHI:
LGTM!

PR #1452, opened 2016-07-19
Hide the output columns of a transformer

### LINKED ISSUE #1451: Hide(DeSelect) the Output Columns if the Transformer has Output Streams
If a transformer implements HasOutputStreams, we want to keep the normal output columns for compatibility with the old jobs. 
However, for the normal Output Columns should be hidden.  

image

ClaudiaPHI:
Fixes #1451 
Hide the output columns of a transformer

### DIFF
--- a/api/src/main/java/org/datacleaner/api/HideOutputColumns.java
+++ b/api/src/main/java/org/datacleaner/api/HideOutputColumns.java
@@ -1,0 +1,42 @@
+/**
+ * DataCleaner (community edition)
+ * Copyright (C) 2014 Neopost - Customer Information Management
+ *
+ * This copyrighted material is made available to anyone wishing to use, modify,
+ * copy, or redistribute it subject to the terms and conditions of the GNU
+ * Lesser General Public License, as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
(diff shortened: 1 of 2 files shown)


LosD:
This seems to have broken

engine/core/src/main/java/org/datacleaner/job/builder/TransformerComponentBuilder.java
```diff
@@ -127,7 +128,17 @@ public TransformerComponentBuilder(AnalysisJobBuilder analysisJobBuilder, Transf
...
                     final String id = _id + "-" + _idGenerator.nextId();
-                    _outputColumns.add(new TransformedInputColumn<>(name, id));
+                    final TransformedInputColumn<Object> column = new TransformedInputColumn<>(name, id);
+                    /*
+                     * If the transformer has the annotation
+                     * "HideOutputColumns", the output columns are hidden in
+                     * other components as input columns. If the user wants to
+                     * use them, he can select them in the transformer.
+                     */
+                    if (this.getDescriptor().getAnnotation(HideOutputColumns.class).isHidden()) {
```
LosD: Break is here: If the annotation is not there, a NPE is guaranteed.

api/src/main/java/org/datacleaner/api/HideOutputColumns.java
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Determines if the output columns {@link Transformer#getOutputColumns()} of
+ * transformer should hidden in other components
```
arjansh: ... should be hidden ... instead of ... should hidden ... ?

arjansh:
One tiny comment regarding a missing word in the javadoc. Outside of that it looks good to me!

api/src/main/java/org/datacleaner/api/HideOutputColumns.java
LosD: Yeah. Can you make a quick fix, then merge?

PR #1453, opened 2016-07-20
Check if the job has been saved been closing the window

### LINKED ISSUE #1450: Save-on-close closes job even though save failed or canceled
If a user closes an unsaved job, opts to save it and then the save fails because of an error, then job will _still_ be closed, meaning loss of data.

## Steps to reproduce
1. Create new job
2. Pull in a table (if not already added) and a component that will fail if not configured properly, like _Create staging table_. Do NOT wire or configure it.
3. Close job.
4. Choose save
5. Close error message

## Expected result
Job is not closed. User does not lose changes made.

## Actual result
Job closed, changes lost.

ClaudiaPHI:
Fixes #1450


LosD:
LGTM!

PR #1460, opened 2016-08-08
Bump MM to 4.5.4

### LINKED ISSUE #1459: Bump MetaModel to 4.5.4
New MM is out with both fixes of bugs affecting us, as well as new features that we need.

Bump.

LosD:
Fixes #1459


jhorcicka:
LGTM. 

However, this MM version does not contain fix for Fixed-width reader with diacritics, we'll have to wait for the next one.

PR #1462, opened 2016-08-09
#1456 Tokenizer disregards number of tokens when target is ROWS

### LINKED ISSUE #1456: Tokenizer should disregard "Number of tokens" when Token target is "rows"
It seems that the "Number of tokens" property is in effect on the Tokenizer component, even if "Token target" is set to rows. This shouldn't be the case I think.

jhorcicka:
Fixes #1456 

In Tokenizer dialog, when "token target" is ROWS then "number of tokens" field should be disabled. This worked fine while changing "token target" (onChange event) but after the dialog window was reopened it was enabled again. This is now fixed. 

tokenizer-cols
tokenizer-rows

### DIFF
--- a/components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TokenizerTransformer.java
+++ b/components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TokenizerTransformer.java
@@ -19,11 +19,14 @@
  */
 package org.datacleaner.beans.transform;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.StringTokenizer;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import org.apache.metamodel.util.HasName;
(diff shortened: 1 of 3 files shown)


kaspersorensen:
I dont think this is the same issue as what I reported, but maybe I should have provided a bit more context/example material.

Take a look at this screenshot:

image

In here I just did a tokenizer on the orderdb's `customer.customername` field. I switched it to target "rows", not touching the "number of tokens" property. It seems that the number 2 (default for the "number of tokens") is still used though! Because e.g. "Amica Models and Co." is only split into two tokens!

jhorcicka:
Ah, thanks for clarification.

jhorcicka:
I hope this is the desired behaviour then. 

tokenizer-cols
tokenizer-rows

kaspersorensen:
Yes, looks good.

LosD:
LGTM!

PR #1464, opened 2016-08-11
Reinstates edition display

### LINKED ISSUE #1463: Edition display broken
A recent change broke the display of edition, so that it always just says: "License and Edition":
image

We need it to read the edition from the Version class again.

LosD:
This reinstates the edition display that was broken by a refactoring of the right-side panel.

Also makes a (very simple) test to ensure it doesn't get broken again by mistake.

Fixes #1463


arjansh:
LGTM!

PR #1468, opened 2016-08-15
Lower verbosity of log message for displaying text using other font

arjansh:
When DataCleaner desktop is unable to display a text using the default font, it falls back to a custom font and logs a message at warning level, which can flood the logs, but isn't really a problem, because it is handled nicely. Therefore lowering verbosity of log message.


ClaudiaPHI:
:+1:

PR #1469, opened 2016-08-18
Feature/1352 enable configuration reference data monitor

### LINKED ISSUE #1352: Enable configuration of reference data in the monitor
The monitor has no way of working with the reference data, making it necessary to upload a new configuration file. This is a pretty error-prone process, both because of the ease and tough consequences of corrupting the file, but also simply because there's a good chance the user will be juggling several configuration files, making it easy to use the wrong one.

jhorcicka:
Resolves #1352 

In Monitor there is a new main page "Reference Data" with lists of dictionaries, synonyms and string patterns. Users can upload new data (by providing conf.xml -- only reference data subsection will be taken) and remove any item one by one. 

reference-data


monitor/services/src/main/java/org/datacleaner/monitor/configuration/WriteUpdatedConfigurationFileAction.java
```diff
@@ -76,9 +92,92 @@ public void run(OutputStream out) throws Exception {
...
+        } else if (referenceDataObject instanceof DatastoreDictionaryType) {
+            return ((DatastoreDictionaryType) referenceDataObject).getName();
+        } else if (referenceDataObject instanceof CustomElementType) {
+            return ((CustomElementType) referenceDataObject).getClassName();
+        } else if (referenceDataObject instanceof TextFileSynonymCatalogType) {
+            return ((TextFileSynonymCatalogType) referenceDataObject).getName();
+        } else if (referenceDataObject instanceof DatastoreSynonymCatalogType) {
+            return ((DatastoreSynonymCatalogType) referenceDataObject).getName();
+        }
+
```
jhorcicka: Is there a better way how to achieve this? I tried: 
```
    private String getComparableName(Object referenceDataObject) {
        if (referenceDataObject instanceof HasName) {
            return ((HasName)referenceDataObject).getName();
        } else if (referenceDataObject instanceof CustomElementType) {
            return ((CustomElementType)referenceDataObject).getClassName();
        }
        
        return "";
    }
```
but it doesn't work. All those specific types implement some interface that extends HasName (except for CustomElementType).

monitor/services/src/main/java/org/datacleaner/monitor/referencedata/ReferenceDataItem.java
```diff
@@ -0,0 +1,57 @@
...
+        SYNONYM_CATALOG,
+        STRING_PATTERN,
+    }
+
+    private String _name;
+    private Type _type;
+
+    public ReferenceDataItem() {
+    }
+
```
jhorcicka: This is, unfortunately, needed by GWT (serialization).

monitor/ui/pom.xml
```diff
@@ -26,6 +27,9 @@
 						<goals>
 							<goal>compile</goal>
 						</goals>
+						<configuration>
+							<sourceLevel>1.7</sourceLevel> <!-- 1.6 or 1.7 -->
```
jhorcicka: 1.8 does not seem to be allowed.

monitor/services/src/main/java/org/datacleaner/monitor/referencedata/ReferenceDataItem.java
LosD: That's weird. The default empty constructor should always be available when no other constructor is created.

Maybe it is some weird GWT parsing thing.

monitor/ui/src/main/webapp/referencedata.xhtml
```diff
@@ -0,0 +1,71 @@
...
+							</span> 
+							<input type="text" class="form-control" readonly="readonly"/>
+						</div>
+						<button type="submit" class="btn btn-primary"> 
+							<span class="glyphicon glyphicon-upload" aria-hidden="true"/> <span>Upload</span>
+						</button>
+					</div>
+				</div>
+			</div>
+
```
jhorcicka: To be able to style it properly, I had to use xhtml/css instead of GWT's FormUpload widget.

monitor/services/src/main/java/org/datacleaner/monitor/configuration/WriteUpdatedConfigurationFileAction.java
LosD: How doesn't it work? `instanceof` should work fine with interfaces?

monitor/ui/pom.xml
LosD: Nope, we're using GWT 2.7 which only allows 1.7

LosD:
It seems to be missing Datastore-based reference data?

LosD:
(as well as simple direct-entry dictionaries, but that is less important)

monitor/services/src/main/java/org/datacleaner/monitor/referencedata/ReferenceDataItem.java
jhorcicka: Sure, but there is another constructor with values.

LosD: Okay, I'm officially completely blind now. Totally missed it :)

monitor/services/src/main/java/org/datacleaner/monitor/configuration/WriteUpdatedConfigurationFileAction.java
jhorcicka: Ah, my bad, those classes do not implement any interfaces (they have "name" property, though). Sorry for misleading info.

LosD: Okay, then there's overloading or reflection, and both probably smells worse than this :)

jhorcicka:
Ah, yes, thanks for noticing. I'll try to fill it.

jhorcicka:
I am closing this PR and branch because the issue was resolved by #1479.

PR #1482, opened 2016-08-30
Bug/1470 removing linked component breaks job

### LINKED ISSUE #1339: Analyzer output stream columns are not always updated properly

### LINKED ISSUE #1470: Removing already linked component can break job
In some cases, when deleting a component whose output column has already been mapped to another column, then re-adding the same component again and mapping it to the same columns, _something_ gets confused.

### Steps to reproduce
1. Link an input column to a component
2. Link component to other component
3. Delete first component
4. Re-add component, and link it to the other component (using "Link to...") again.
5. Save and reload job.

### Expected outcome
Links fine, saves and reloads without issue

### Actual outcome
Either doesn't look like it is linked in the graph UI, or links the same column twice.
When opening job, it fails.

### Notes
The issue does not seem to be in "Link to...", but that the original component column links aren't deleted from the other component, which means the new component can/will be confused with the original. This makes it possibly related to #1339.

When the job is saved, it looks like the input column ids get a "2" attached, which is what breaks the loading. Part of the symptom might be from the fix to #1030 (I don't think that fix is wrong, though, it just exposes this issue).

We probably want to make sure that JaxbJobWriter refuses to save a job broken like this, instead of inventing ids for columns that doesn't exist.

### Workarounds
To avoid the issue, either delete the other component with the first, or make sure to manually remove the dangling references.

If you realize the job was broken when loading it, open it with an editor and look for input column ids that has a "2" attached in the end, without a corresponding output column, then remove the "2".

arjansh:
Fixes #1470.
Also fixes #1339.

When you remove a transformer or remove a field for a transformer's configuration, other components using fields provided by the transformer in an input column need to be updated in that such fields are removed, because they no longer are available.

This is achieved by registering a listener on a transformer when assigned such a field to a component. The listener removes such fields from components when they are removed from a transformer (or the transformer is removed).


arjansh:
This pull request also fixes a large part of #1339. It fixes the part that deleting transformers is not propagated through the entire flow. It doesn't fix the part that mapping of other columns becomes corrupted.

LosD:
Hmmm... Does it fix #1339 or does it only fix it partially? Your description makes it sound like a full fix, but the comment makes it sound partial :)

If only partial, you'll probably need to edit the description, otherwise it will close the issue when merged.

arjansh:
It fixes #1339 completely. I updated my outdated comment.

LosD:
That's totally what I wanted to hear. :-)

engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
```diff
@@ -322,11 +324,50 @@ public B setConfiguredProperty(String configuredName, Object value) {
     public B setConfiguredProperty(ConfiguredPropertyDescriptor configuredProperty, Object value) {
         final boolean changed = setConfiguredPropertyIfChanged(configuredProperty, value);
         if (changed) {
+            if (configuredProperty.isInputColumn()) {
+                for (TransformedInputColumn<?> transformedInputColumn : getTransformedInputColumns(value)) {
```
LosD: Seems a little weird having non-functional for loop with a functional expression inside.

jhorcicka:
Can anyone check and confirm/reject following?

Simple job with "customers", "concatenator", "text-case-transformer" and "value distribution": 
job-01
When I load and execute the job, everything goes fine. When I remove "concatenator", I get error as expected.
job-02
When I recreate the job to its original state: 
job-03
job-04
job-05
It can be executed again. But when I want to close the job I get the error again (I tried both saving and not-saving the job before closing it). 
job-06

engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
```diff
@@ -322,11 +324,50 @@ public B setConfiguredProperty(String configuredName, Object value) {
...
+                                            new ComponentBuilderTransformerChangeListener(this, configuredProperty)));
+                }
+            }
+
             onConfigurationChanged();
         }
         return (B) this;
     }
 
+    protected boolean isProvidingColumn(final TransformedInputColumn<?> transformedInputColumn,
```
LosD: Not sure if it makes sense, but if this was a method on TransformedInputColumn, you could use a method reference above.

engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
```diff
@@ -383,10 +424,62 @@ protected boolean setConfiguredPropertyIfChanged(final ConfiguredPropertyDescrip
...
+                                            dependentProperty.setValue(_configurableBean, null);
+                                        }
+                                    }
+                                }
+                            });
+        }
+    }
+
+    private static <E> E[] getArray(Class<E> clazz, List<?> baseList) {
+        E[] result = (E[]) Array.newInstance(clazz, baseList.size());
```
LosD: This seems expensive. Is it really needed?

LosD: Oh, and "final" (goes for a lot of the new code in this file)  :)

engine/core/src/main/java/org/datacleaner/job/builder/ComponentBuilderTransformerChangeListener.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public ComponentBuilderTransformerChangeListener(ComponentBuilder componentBuilder,
+            ConfiguredPropertyDescriptor propertyDescriptor) {
+        _componentBuilder = componentBuilder;
+        _propertyDescriptor = propertyDescriptor;
+    }
+
+    @Override
+    public void onAdd(TransformerComponentBuilder<?> builder) {
+        // Do nothing.
```
LosD: I've seen a lot of this in other implementations of TransformerChangeListener... Should we make an adapter at some point (not now)?

engine/core/src/main/java/org/datacleaner/job/builder/TransformerComponentBuilder.java
```diff
@@ -334,7 +334,9 @@ protected void onRemovedInternal() {
      *            a new change listener
      */
     public void addChangeListener(TransformerChangeListener listener) {
-        _localChangeListeners.add(listener);
+        if (!_localChangeListeners.contains(listener)) {
```
LosD: This scares me. I think we ping-ponged on `localChangeListeners.contains` in various places quite a few times. Maybe check if this was one of those places, and if it was, why it was removed. I'm afraid of re-opening new bugs.

engine/core/src/test/java/org/datacleaner/job/builder/BaseMockAnalyzer.java
```diff
@@ -0,0 +1,38 @@
...
+    public void run(InputRow row, int distinctCount) {
+        // Do nothing.
+    }
+
+    @Override
+    public AnalyzerResult getResult() {
+        return null;
+    }
+
+}
```
LosD: Missing EOL

engine/core/src/test/java/org/datacleaner/job/builder/MockAnalyzer.java
```diff
@@ -0,0 +1,40 @@
...
+    @MappedProperty("Columns")
+    public String[] _columnNames;
+
+    @Configured
+    public InputColumn<?> _column;
+
+    @Configured
+    @MappedProperty("Column")
+    public String _columnName;
+}
```
LosD: Missing EOL

engine/core/src/test/java/org/datacleaner/job/builder/MultipleJobsAnalyzer.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.datacleaner.api.ColumnProperty;
+import org.datacleaner.api.Configured;
+import org.datacleaner.api.InputColumn;
+
+public class MultipleJobsAnalyzer extends BaseMockAnalyzer {
+    @Configured
+    @ColumnProperty(escalateToMultipleJobs = true)
+    public InputColumn<?> _column;
+}
```
LosD: Missing EOL

engine/core/src/main/java/org/datacleaner/job/builder/ComponentBuilderTransformerChangeListener.java
```diff
@@ -0,0 +1,100 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    @Override
+    public boolean equals(Object object) {
```
LosD: `hashcode()` seems to be missing (`equals()` and `hashcode()` should always be overridden in pairs).

You can use `Objects.hash(_componentBuilder, _propertyDescriptor)` for a quick implementation.

LosD:
Mostly looks good to me, but check comments. Will do my own testing when I'm home.

LosD:
BTW, anyone are more than welcome to test. We reaaaalllly don't want this to keep breaking.

engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
arjansh: I will add the final :), this method is really needed, because I have a list of objects and I need to pass a typed array to the ConfiguredPropertyDescriptor#setValue(Object, Object) method, or otherwise I'll get some exception that the value can't be set because it is of the wrong type. I could figure out no easier way to get a typed array from my list of objects (believe me, I tried).

engine/core/src/main/java/org/datacleaner/job/builder/ComponentBuilderTransformerChangeListener.java
arjansh: Sounds reasonable. (not now however :))

engine/core/src/main/java/org/datacleaner/job/builder/TransformerComponentBuilder.java
arjansh: I checked and I wasn't ping-ponged here. It is logical that it's scary, because it's a poor man's solution to the problem that the addChangeListener is in fact invoked way too often (as a result of this pull request) and this is the only way too ensure the number of listeners doesn't completely explode. With the current setup I could figure out no way to neatly remove listeners when they're no longer needed without going through a number of extremely complex setup hoops.

LosD: That's good. And yeah, we need all that fixed.

jhorcicka:
I re-tested the issue I posted (multiple times) and now it works fine.

LosD:
LGTM!

PR #1485, opened 2016-08-31
Changes rendering and JS importing.

### LINKED ISSUE #1484: Html renderers using graphs rather broken
The HTML renderers has many very breaking problems: The homegrown importJS often fails more often than not and all data is written into the same global "data" variable (which might have worked once, but now it is read using an anonymous inner function called later, where the next graph's data has overwritten it).

Lets fix using a common JS import library and element-prefixed data names.

LosD:
Since it is using a new, incompatible helper library (http://eobjects.org/resources/datacleaner-html-rendering/analysis-result-v2.js), this might break extensions.

Currently has a bug where Flot graphs aren't perfectly sized. Still seems better than what was.

Fixes #1484


components/html-rendering/src/main/scala/org/datacleaner/result/html/FlotChartLocator.scala
```diff
@@ -17,42 +17,42 @@ object FlotChartLocator {
...
    * Gets the URL for the flot plugin for stacked charts, typically named as "jquery.flot.stack.min.js"
    */
-  def getFlotStackUrl: String = getFlotHome() + "/jquery.flot.stack.min.js";
+  def getFlotStackUrl: String = getFlotHome() + "/jquery.flot.stack.js";
 
   /**
    * Gets the URL for the flot plugin for threshold effect, typically named as "jquery.flot.threshold.min.js"
    */
-  def getFlotThresholdUrl: String = getFlotHome() + "/jquery.flot.threshold.min.js";
+  def getFlotThresholdUrl: String = getFlotHome() + "/jquery.flot.threshold.js";
```
jhorcicka: Why don't we have/use minified versions anymore?

components/html-rendering/src/test/resources/benchmark-renderings/HtmlAnalysisResultWriterTest-testEmptyResult.html
```diff
@@ -1,13 +1,35 @@
...
 <head>
-  <title>Analysis result</title>  <script type="text/javascript" src="http://eobjects.org/resources/datacleaner-html-rendering/analysis-result.js"></script>
-<link rel="shortcut icon" href="http://eobjects.org/resources/datacleaner-html-rendering/analysis-result-icon.png" />
+  <title>Analysis result</title>  <link rel="shortcut icon" href="http://eobjects.org/resources/datacleaner-html-rendering/analysis-result-icon.png" />
+<link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.0/themes/base/jquery-ui.css" />
+<link rel="stylesheet" type="text/css" href="http://eobjects.org/resources/datacleaner-html-rendering/analysis-result.css?load=script" />
 <script type="text/javascript">//<![CDATA[
-  var analysisResult = {};
+var analysisResult = {};
+requirejs = {
```
jhorcicka: I see it's "only" in tests but can the same "requirejs configuration" be shared somehow among BaseHeadElement.java, HtmlAnalysisResultWriterTest-testEmptyResult.html and HtmlAnalysisResultWriterTest-testSingleResultElement.html?

jhorcicka:
LGTM (just a few questions)

components/html-rendering/src/main/scala/org/datacleaner/result/html/FlotChartLocator.scala
LosD: Damn, I think I reinstated those. Anyway I think we actually don't use this anymore, but maybe we should (I originally moved the configuration to the eobjects server, but that failed since the configuration gets loaded too late).

components/html-rendering/src/test/resources/benchmark-renderings/HtmlAnalysisResultWriterTest-testEmptyResult.html
LosD: I didn't worry too much about it, since we'll hopefully rework this as part of the monitor rewrite, but there's a good chance we'll skip this when we do that, so let's do it now.

jhorcicka:
Are these "numbers in image" related to that known/reported bug? 

This one is OK. 
graph-1

Numbers on y-axis are in the graph. 
graph-2

Numbers on y-axis are in the graph. 
graph-3

LosD:
I didn't notice that (only the scroll bar) but probably yes. 

Try zooming in and out from the graph (not the browser zoom, draw a rectangle with your mouse in the graph), then it should fix itself if it is the same. 

I think the reason the first one doesn't fail, is that it is visible when rendered.

jhorcicka:
Yes, when I zoom then it's fine.

LosD:
Okay, then it's the same thing. Let's make a bug for that, and merge this once I handled your comments

components/html-rendering/src/main/scala/org/datacleaner/result/html/FlotChartLocator.scala
LosD: Nope, it makes no sense anymore. Flot charts are configured using the requirejs configuration

components/html-rendering/src/test/resources/benchmark-renderings/HtmlAnalysisResultWriterTest-testEmptyResult.html
LosD: Ah, no. These are "benchmark" files. Part of the standard testing way, so they are just copied from an inspected result. Honestly I don't like these at all (I believe in testing actual results, not blocks of text), but at least they are easy to copy from the output (after a careful inspection, of course).

LosD:
IE11 and Edge seems happy

arjansh:
LGTM!

PR #1515, opened 2016-09-14
Disable "Save" and "Save as" when datastore is null

### LINKED ISSUE #1498: Disable "save" actions when no job is opened
On the main screen of DC desktop you can click "Save" or "Save as" even if you're not building a job right now. 

no-job-save

I think these actions should be disabled when there is nothing to save. 

EDIT: They are disabled after you start DC. But when you open a job and close it again, these actions stay enabled and that is the issue here.

ClaudiaPHI:
fixes #1498

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
@@ -908,7 +908,7 @@ public final class AnalysisJobBuilderWindowImpl extends AbstractWindow implement
     }
 
     private void onSourceColumnsChanged() {
-        boolean everythingEnabled = !_analysisJobBuilder.getSourceColumns().isEmpty();
+        boolean everythingEnabled = !_analysisJobBuilder.getSourceColumns().isEmpty() && _datastore != null;
 
         _saveButton.setEnabled(everythingEnabled);
         _saveAsButton.setEnabled(everythingEnabled);


desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -908,7 +908,7 @@ public void applyPropertyValues() {
     }
 
     private void onSourceColumnsChanged() {
-        boolean everythingEnabled = !_analysisJobBuilder.getSourceColumns().isEmpty();
+        boolean everythingEnabled = _datastore != null;
```
LosD: This seems wrong. Shouldn't it be _both_?

LosD:
Travis error is seemingly unrelated (stalled build)

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
ClaudiaPHI: There is no reason to be both.  It works perfectly when I tested.

jhorcicka:
The first difference is just after starting a new empty job. In earlier version save buttons were allowed but clicking on them caused an exception ("fix job problems to save it"). Now the buttons are disabled until a datastore is added to the graph.

LosD:
Well spotted! So we should keep both the original check and the null check, then.

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -908,7 +908,7 @@ public void applyPropertyValues() {
     }
 
     private void onSourceColumnsChanged() {
-        boolean everythingEnabled = !_analysisJobBuilder.getSourceColumns().isEmpty();
+        boolean everythingEnabled = !_analysisJobBuilder.getSourceColumns().isEmpty() &&  _datastore != null;
```
LosD: Now I'm totally being a pain in the behind, but there's a double space between `&&` and `_datastore`

ClaudiaPHI: done

LosD:

PR #1516, opened 2016-09-15
#552 JavaScriptAdvancedTransformer -- default code was unrunnable at Monitor

### LINKED ISSUE #552: JS job fails on monitor server
Reporting this bug to investigate on behalf of another user. When deploying an advanced JS transformer job on DC monitor, this issue occurs:

```
11:06:05 - Job execution FAILURE
- Can't find method ch.qos.logback.classic.Logger.debug(string,java.lang.String[],java.lang.Object[],org.eobjects.analyzer.job.concurrent.ThreadLocalOutputRowCollector). (JavaScriptAdvancedTransformer#8) (EvaluatorException)
- Failure component: ImmutableTransformerJob[name=Q_Invalid_Columns,transformer=JavaScript transformer (advanced)]
```

jhorcicka:
Fixes #552 

In JavaScriptAdvancedTransformer default code there was ```logger.debug(String, String[], String[], String)``` used. The last item was outputCollector which actually provided it's Object.toString() output (not so useful for users anyway). So, it was removed from the debug output and now ```logger.debug(String, String[], String[])``` is used which works fine.

### DIFF
--- a/components/data-structures/src/main/java/org/datacleaner/beans/datastructures/BuildListTransformer.java
+++ b/components/data-structures/src/main/java/org/datacleaner/beans/datastructures/BuildListTransformer.java
@@ -45,7 +45,7 @@ import org.slf4j.LoggerFactory;
 @Categorized(DataStructuresCategory.class)
 public class BuildListTransformer implements Transformer {
 
-    private static final Logger logger = LoggerFactory.getLogger(BuildMapTransformer.class);
+    private static final Logger logger = LoggerFactory.getLogger(BuildListTransformer.class);
 
     @Inject
     @Configured
--- a/components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptAdvancedTransformer.java
+++ b/components/javascript/src/main/java/org/datacleaner/beans/script/JavaScriptAdvancedTransformer.java
@@ -66,7 +66,7 @@ public class JavaScriptAdvancedTransformer implements Transformer {
     @StringProperty(multiline = true, mimeType = { "text/javascript", "application/x-javascript" })
     String sourceCode = "var transformerObj = {\n"
             + "\tinitialize: function() {\n\t\tlogger.info('Initializing advanced JavaScript transformer...');\n\t},\n\n"
-            + "\ttransform: function(columns,values,outputCollector) {\n\t\tlogger.debug('transform({},{}) invoked', columns, values);\n\t\tfor (var i=0;i<columns.length;i++) {\n\t\t\toutputCollector.putValues(columns[i],values[i])\n\t\t}\n\t},\n\n"
+            + "\ttransform: function(columns,values,outputCollector) {\n\t\tlogger.debug('transform({},{},{}) invoked', columns, values, outputCollector);\n\t\tfor (var i=0;i<columns.length;i++) {\n\t\t\toutputCollector.putValues(columns[i],values[i])\n\t\t}\n\t},\n\n"
             + "\tclose: function() {\n\t\tlogger.info('Closing advanced JavaScript transformer...');\n\t}\n}";
 
     @Inject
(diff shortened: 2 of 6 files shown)


LosD:
Hmmmm... This fixes the symptom, but not the underlying issue. Multiple args logging should work the same both in desktop and monitor.

jhorcicka:
I thought it did not work in desktop as well but I've just tried that and it does. So, I will continue...

jhorcicka:
I removed logback-classic dependency from Monitor and it fixed this issue. I am not aware of potential consequences, though. Can you comment this change, @LosD?

LosD:
Not sure why that is... Maybe try to find out what the difference is between desktop and monitor since it matters.

jhorcicka:
Diff command of jar and war "lib" directories gives the following: 
```
Only in jar-lib: log4j-1.2.17.jar
Only in jar-lib: slf4j-log4j12-1.7.10.jar
Only in war-lib: log4j-over-slf4j-1.7.10.jar
```

jhorcicka:
To simulate the problem, I tried to run the following piece of code: 
```
import org.mozilla.javascript.Context;
import org.mozilla.javascript.Scriptable;
import org.mozilla.javascript.ScriptableObject;
import org.slf4j.Logger;

public class Main {
    private static final Logger logger = LoggerFactory.getLogger(Main.class);
    
    public static void main(String args[]) {
        //logger.debug("java: a={},b={}", 42, 'B');
        logger.debug("java: a={},b={},c={}", 42, 'B', new String[] { "1", "2" });
        final Context cx = Context.enter();
        
        try {
            final Scriptable scope = cx.initStandardObjects();
            ScriptableObject.putProperty(scope, "logger", logger);
            
            //final String js = "logger.debug('javascript: a={},b={}', 42, 'B')";
            final String js = "logger.debug('javascript: a={},b={},c={}', 42, 'B', ['1','2'])";
            
            final Object result = cx.evaluateString(scope, js, "<cmd>", 1, null);
            System.err.println(Context.toString(result));
        } finally {
            Context.exit();
        }
    }
}

	<dependencies>
		<!--<dependency>-->
			<!--<groupId>ch.qos.logback</groupId>-->
			<!--<artifactId>logback-classic</artifactId>-->
			<!--<version>1.0.13</version>-->
		<!--</dependency>-->
		<dependency>
			<groupId>log4j</groupId>
			<artifactId>log4j</artifactId>
			<version>1.2.17</version>
		</dependency>
		
		<dependency>
			<groupId>org.slf4j</groupId>
			<artifactId>slf4j-log4j12</artifactId>
			<version>1.7.10</version>
		</dependency>
		<dependency>
			<groupId>org.mozilla</groupId>
			<artifactId>rhino</artifactId>
			<version>1.7R4</version>
		</dependency>
	</dependencies>
```

When log4j is used, everything goes fine. When logback is used, then java output is there but javascript output fails because of "Can't find method ch.qos.logback.classic.Logger.debug(string,number,string,object)". Basically our issue.

LosD:
That's weird. The whole idea with SLF4J is to hide the actual logger implementation.

Even weirder is that SLF4J and logback are related, so if anything you'd expect it to work better, not worse :O

I guess Rhino is doing something weird?

jhorcicka:
Perhaps, but I also tried to update rhino's version to the newest and it did not have any effect.

LosD:
Hmmm... It seems like it is us actually mapping it to the logger.

LosD:
Might be something to do with it being called using reflection.

LosD:
@kaspersorensen Do you remember if there is anything tying us (or gives us an advantage) to use log4j in the desktop, and logback in the monitor?

kaspersorensen:
Only that log4j is documented in the reference docs and such so it would be a (minor) breaking change to go for logback since users may have created their own log4j config files.

LosD:
I was actually more thinking of going the other way, since it seems to be logback that is giving us trouble. Biggest drawback is that I would really like to know why in the world it is failing instead of just chickening out :)

@jhorcicka: Another alternative is reviving my old Rhino->Nashorn branch and see if the logging works better: https://github.com/datacleaner/DataCleaner/tree/feature/1105-javax-script-based-JS-transformers

The reason it stalled was that we're doing some funny business with the user data both acting as an array and as a map, and something else came up while I was reviewing how to replicate that in Nashorn.

jhorcicka:
When I updated the simple code provided above to use "nashorn", then javascript logging did not work for both logging frameworks (logback, log4j). Same problem, no method accepting these argument types was found.

jhorcicka:
Own JavaScriptLogger replaced the classical Logger which fixes the problem at Monitor without removing logback dependency. 

@LosD: please review and merge at will.

LosD:
LGTM!

PR #1517, opened 2016-09-15
Use proper parsing instead of non-working homegrown solution

### LINKED ISSUE #1490: URL parser fails with many links
The URL parser is very fragile. Take this CSV: https://gist.github.com/LosD/32132d6909fd989bd2d9580d6828eb5c

All except one link fails. It seems to have something to do with links without a path and/or unexpected characters.

### Steps to reproduce
1. Download and open DC job on linked CSV file
1. Add URL Parser
1. Preview

### Actual result
<null> in every output field, except for one line

### Expected result
Parsable fields, like Domain should be parsed

LosD:
This many issues with the standardizer, but also breaks compatibility a little bit, as it does not add username to host, or fragment to query. I did keep the a field with no path at all as null, as it is not technically filled. I had a bit of a hard time deciding here, so please comment if you disagree.

For the sake of not breaking jobs, the port number is kept a string even though that makes no sense.

Please note that I did a full-file reformat as it is only the same file in name and (intended) function.

 Fixes #1490

### DIFF
--- a/components/standardizers/src/main/java/org/datacleaner/beans/standardize/UrlStandardizerTransformer.java
+++ b/components/standardizers/src/main/java/org/datacleaner/beans/standardize/UrlStandardizerTransformer.java
@@ -19,100 +19,71 @@
  */
 package org.datacleaner.beans.standardize;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.net.URI;
+import java.net.URISyntaxException;
 
 import javax.inject.Named;
 
 import org.datacleaner.api.Categorized;
 import org.datacleaner.api.Configured;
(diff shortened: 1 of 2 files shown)


kaspersorensen:
:+1: from me. I would even change the port to number and add the username and such. But that's just me being a bit aggressive wrt. breaking backwards compatibility maybe :)

LosD:
Hmmm, I won't break peoples jobs without providing them a way to convert them. Especially not unless it is a major version.

ClaudiaPHI:
LGTM!

arjansh:
LGTM!

PR #1518, opened 2016-09-19
fixes #1077. Improved the upload zip folder controller

### LINKED ISSUE #1077: Avoid RepositoryZipController.uploadRepository(...) from deleting old repository if input file is empty
I was testing the restore functionality of ```RepositoryZipController.uploadRepository(...)``` the other day and found that it is quite dangerously just replacing the contents of the repo - even if the new is completely blank.

I would say that we should _at least_:

 * Check that the input file is valid - that it at least has basic repo characteristics like being a zip file, containing a conf.xml file etc.
 * Place the old repo in tmp files for possible cancellation.

ClaudiaPHI:
fixes #1077. 
Improved the upload zip folder controller

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/RepositoryZipController.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/RepositoryZipController.java
@@ -19,12 +19,13 @@
  */
 package org.datacleaner.monitor.server.controllers;
 
+import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.HashMap;
+import java.util.Calendar;
 import java.util.List;
-import java.util.Map;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import java.util.zip.ZipOutputStream;
@@ -52,6 +53,7 @@ import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.multipart.MultipartFile;
 
+
 @Controller
 @RequestMapping("/{tenant}/zip")
 public class RepositoryZipController {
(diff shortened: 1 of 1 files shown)


ClaudiaPHI:
Sample response: 

image

LosD:
I've been involved in decisions here, so I'm excusing myself from reviewing, but I'd like to add a remark to the change from JSON to plain text: Since we're presenting the path to the user so that s/he will know where to find it, and JSON will escape backslashes, it was deemed better to go with simple, but readable text/plain for now.

This will probably be reverted when we're going away from just showing JSON as a response to user actions (we decided not to do that now, since we're ripping the monitor UI completely apart soon).

@JoosjeBoon, just to be sure: You're not using this API, right?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/RepositoryZipController.java
```diff
@@ -78,28 +80,58 @@ public void downloadRepository(@PathVariable("tenant") final String tenant, Http
...
+        if (file.getSize() == 0){
+            return "Failure. The file is empty"; 
+        } 
+        
+        final String contentType = file.getContentType().trim();
+        if ((!contentType.equals("application/zip")) && (!contentType.equals("application/octet-stream"))) {
+            return "Failure. The file isn't a .zip archive";
+        }
+        
+        final File oldRepositoryZipFile = createZipfolder(rootFolder);
```
arjansh: Why do a preemptive copy of the old repository, before first checking if the uploaded zip is a valid repository? Why not only do this if the uploaded zip is a valid repository? In that manner, there is no overhead of creating it and it doesn't have to be deleted in case the uploaded zip is an invalid repository.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/RepositoryZipController.java
```diff
@@ -52,6 +52,8 @@
 import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.multipart.MultipartFile;
 
+import com.ibm.icu.util.Calendar;
```
arjansh: Why use com.ibm.icu.util.Calendar instead of java.util.Calendar?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/RepositoryZipController.java
```diff
@@ -78,28 +80,58 @@ public void downloadRepository(@PathVariable("tenant") final String tenant, Http
...
+        final long timeInMillis = Calendar.getInstance().getTimeInMillis();
+        final File tempfile = File.createTempFile("repository_" + rootFolder.getName() + "_" + timeInMillis, ".zip");
+        final FileOutputStream fos = new FileOutputStream(tempfile);
+        final ZipOutputStream zipOutput = new ZipOutputStream(fos);
+        compress(rootFolder, zipOutput);
+        logger.info("The old repository has been compressed and uploaded to" + tempfile.getAbsolutePath());
+        FileHelper.safeClose(zipOutput);
+        return tempfile;
+    }
+    /** the repository is valid if it contains a conf.xml file**/
```
arjansh: Why not use javadoc style for the comment? Seems like a better fit for a comment describing what a method does.

arjansh:
@ClaudiaPHI I added some inline comment. can you look at those?

LosD:
Travis failure not related. Another stalled build (seems like Tattletale stalled)

LosD:
Restarted build

JoosjeBoon:
These improvements are not breaking for DataHub. We still use the iframes instead of api's.

arjansh:
LGTM!

PR #1520, opened 2016-09-20
Fix possible deadlock in ClasspathScanDescriptorProvider

### LINKED ISSUE #1467: ClasspathScanDescriptorProvider can cause deadlock
When you add a DescriptorProviderListener to a ClasspathScanDescriptorProvider which (in)directly invokes ClasspathScanDescriptorProvider#scanPackage(String, boolean, ClassLoader, boolean, File[]) deadlock can occur because this method executes a Task and invokes AbstractDescriptorProvider#notifyListeners() before it is fully finished, that is, it hasn't invoked its taskDone() method yet, which decrements its _tasksPending number, which causes additional invocations of to ClasspathScanDescriptorProvider#scanPackage(String, boolean, ClassLoader, boolean, File[]) to wait.

arjansh:
Fixes #1467.

Make sure listeners are notified when task is really done, so no deadlock occurs because _tasksPending is greater than 0 and blocks running of tasks.


LosD:
Seems like a good and correct change... But also a dangerous one. Are we sure that listeners are not incorrectly expecting to get notified earlier?

I'm a little worried about the disrepancy between your description and what actually happens: `notifyListeners()` also seems to be called if `_tasksPending` is greater than 0 (call is outside the check).

arjansh:
Effectively this change only re-orders the manner in which the value of the _tasksPending is decremented and the moment at which the listeners are notified. Nothing more. The invocation of the notifyListeners() method was the last thing inside the execute() method of the executed task, after which the onComplete() is invoked which directly invokes taskDone(). The listeners are expected to be notified when the task is done, but can become deadlocked when the value of the _tasksPending isn't decremented before, because then the situation might occur that a task finishes, invokes its notifyListeners() method and the invoked listener (in)directly invokes a method on the ClasspathScanDescriptorProvider, which lets it wait, because the _tasksPending hasn't been decremented and as such it interprets it as the task which just finished is still running, causing the deadlock.

LosD:
Ah, yes. I'm happy! :)

PR #1524, opened 2016-09-23
Refresh schemas on datastore when retrieving them

### LINKED ISSUE #1461: Jobs with ExcelWriter may fail in monitor with a NPE in MetaModel
When running a job containing an ExcelWriter, there's a chance that the job will fail with a NullPointException. Unfortunately, it is pretty elusive (I had the issue, changed the monitor to run with a debugger, and the problem dissappeared, even after disabling the debugger again).

I looks a bit like the MetaModel ExcelContext race conditions we have seen before.

```
java.lang.NullPointerException: null
	at org.apache.metamodel.excel.ExcelUpdateCallback.getLastRowNum(ExcelUpdateCallback.java:131) ~[MetaModel-excel-4.5.2.jar:4.5.2]
	at org.apache.metamodel.excel.ExcelUpdateCallback.createRow(ExcelUpdateCallback.java:125) ~[MetaModel-excel-4.5.2.jar:4.5.2]
	at org.apache.metamodel.excel.ExcelInsertBuilder.execute(ExcelInsertBuilder.java:52) ~[MetaModel-excel-4.5.2.jar:4.5.2]
	at org.datacleaner.output.AbstractMetaModelOutputWriter$1.run(AbstractMetaModelOutputWriter.java:94) ~[DataCleaner-engine-core-5.0.3.jar:na]
	at org.apache.metamodel.excel.ExcelDataContext.executeUpdate(ExcelDataContext.java:232) ~[MetaModel-excel-4.5.2.jar:4.5.2]
	at org.datacleaner.output.AbstractMetaModelOutputWriter.flushBuffer(AbstractMetaModelOutputWriter.java:85) ~[DataCleaner-engine-core-5.0.3.jar:na]
	at org.datacleaner.output.AbstractMetaModelOutputWriter.close(AbstractMetaModelOutputWriter.java:105) ~[DataCleaner-engine-core-5.0.3.jar:na]
	at org.datacleaner.extension.output.AbstractOutputWriterAnalyzer.getResult(AbstractOutputWriterAnalyzer.java:75) ~[DataCleaner-writers-5.0.3.jar:na]
	at org.datacleaner.extension.output.AbstractOutputWriterAnalyzer.getResult(AbstractOutputWriterAnalyzer.java:38) ~[DataCleaner-writers-5.0.3.jar:na]
	at org.datacleaner.job.tasks.CollectResultsTask.execute(CollectResultsTask.java:60) ~[DataCleaner-engine-core-5.0.3.jar:na]
	at org.datacleaner.job.concurrent.TaskRunnable.run(TaskRunnable.java:61) ~[DataCleaner-api-5.0.3.jar:na]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [na:1.8.0_101]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [na:1.8.0_101]
	at java.lang.Thread.run(Unknown Source) [na:1.8.0_101]
```

arjansh:
Fixes #1461.

Make sure schemas on a datastore are refreshed when retrieving them from the datastores map on the ExcelOutputWriterFactory class, because they may contain stale references. This can particularly be the case when accessing multiple sheet within an Excel sheet within one job where existing sheets are overwritten, which in practice means a sheet is deleted and should be recreated, which obviously may fail if it is still being referenced even though it has been deleted.


LosD:
I'd like to see a little bit of JavaDoc on the `MultiStreamTransformer`. Just to tell its purpose (which is only testing, I guess)

LosD:
Otherwise LGTM

arjansh:
@LosD:
I added a javadoc to the `MultiStreamTransfomer`. Maybe it would also make sense to rename it to something like `MultiStreamTestTransformer`.

Next to that, the `testConcurrentWriting` test which I added to the CreateExcelSpreadsheetAnalyzerTest class is really an integration test and not a unit test. The fix for the test is even in another maven module. Maybe it should be moved and run as an integration test?

LosD:
Makes sense!

components/writers/src/test/java/org/datacleaner/extension/output/MultiStreamTestTransformer.java
```diff
@@ -0,0 +1,82 @@
...
+        if (_even && _evenRowCollector != null) {
+                _evenRowCollector.putValues(row.getValues(_valueColumns).toArray());
+        } else {
+            if (!_even && _unevenRowCollector != null) {
+                _unevenRowCollector.putValues(row.getValues(_valueColumns).toArray());
+            }
+        }
+        _even = !_even;
+    }
+}
```
LosD: Missing EOL

LosD:
Single EOL line missing, otherwise LGTM.

Might be easiest just to let Travis complete now, add the EOL and merge straight away. I'm more than confident that the newline wont break the build :)

Issue #1201, opened 2016-03-03
Multithreaded inserting of rows into SQL Server fails


LosD:
Jobs that insert large amounts of rows (500000+) in an SQL Server table fails when running in normal (=multithreaded) mode, even if run with multiple connections disabled.

LosD:
This may be an issue with the jTDS driver, at least uPortal has [had some issues](https://wiki.jasig.org/display/UPM40/MSSQL+with+the+jTDS+JDBC+Driver), requiring special settings to avoid problems. They are recommending MS' own driver.

At least it is probably a good idea to compare pros and cons.

ClaudiaPHI:
The multi-threading insertion into SQL server fails if there is 1 connection to database. 

- Jtds driver gives the following exception: 

```
Could not rollback transaction: rollback() should not be called while in auto-commit mode.

org.apache.metamodel.MetaModelException: Could not rollback transaction: rollback() should not be called while in auto-commit mode.
	at org.apache.metamodel.jdbc.JdbcUtils.wrapException(JdbcUtils.java:61)
	at org.apache.metamodel.jdbc.JdbcUpdateCallback.commitOrRollback(JdbcUpdateCallback.java:112)
	at org.apache.metamodel.jdbc.JdbcUpdateCallback.close(JdbcUpdateCallback.java:85)
	at org.apache.metamodel.jdbc.JdbcDataContext.executeUpdate(JdbcDataContext.java:850)
	at org.datacleaner.beans.writers.InsertIntoTableAnalyzer.run(InsertIntoTableAnalyzer.java:414)
	at org.datacleaner.beans.writers.InsertIntoTableAnalyzer.run(InsertIntoTableAnalyzer.java:76)
	at org.datacleaner.util.WriteBuffer.flushBuffer(WriteBuffer.java:84)
	at org.datacleaner.util.WriteBuffer.addToBuffer(WriteBuffer.java:60)
	at org.datacleaner.beans.writers.InsertIntoTableAnalyzer.run(InsertIntoTableAnalyzer.java:377)
	at org.datacleaner.job.runner.AnalyzerConsumer.consumeInternal(AnalyzerConsumer.java:71)
	at org.datacleaner.job.runner.AbstractRowProcessingConsumer.consume(AbstractRowProcessingConsumer.java:159)
	at org.datacleaner.job.runner.ConsumeRowHandlerDelegate.consume(ConsumeRowHandlerDelegate.java:64)
[...]

PR #1525, opened 2016-09-26
partial fix for #1201

ClaudiaPHI:
This PR fixes the exception caused by JDTS driver, when trying to run DC in multi-threaded mode, but with a single connection to the SQL server. 
 See #1201  

Regarding the exception caused by Microsoft SQL server driver, when trying to run DC in multi-threaded, but with a single connection to the database, @LosD  and I, think it's best to do nothing as it might affect the performance of the system when we are trying to run in multi-threaded mode and with multiple connections to the database.  I will add a section in the documentation's troubleshooting page, on how to avoid the exception.


LosD:
You cannot just dump a jar file in DataCleaner. It's not using a compatible license. I am going to have to delete this branch now, as we're currently breaking MS' rights under their license. Please re-add without the JAR.

PR #1532, opened 2016-09-27
Many jobs at Monitor -- performance

### LINKED ISSUE #1501: Monitor performance issues with many jobs
The performance of the monitor seems to go down quite a bit when many jobs are registered. Especially just showing the scheduled page takes a long time, and adding an alert also seems to be very slow.

jhorcicka:
Fixes #1501 

For Scheduling page (listing of jobs) I used lighter version of "getSchedules" where I omitted properties loading. It does not seem to be necessary just for listing and then simple actions (run, delete, show history, ...). 

Time of listing 1000 copies of a simple job was decreased from 26s to 9s.

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/scheduling/SchedulingService.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/scheduling/SchedulingService.java
@@ -42,7 +42,8 @@ import com.google.gwt.user.client.rpc.RemoteServiceRelativePath;
 public interface SchedulingService extends RemoteService {
 
     @RolesAllowed({ SecurityRoles.VIEWER, SecurityRoles.SCHEDULE_EDITOR })
-    public List<ScheduleDefinition> getSchedules(TenantIdentifier tenant) throws DCSecurityException;
+    public List<ScheduleDefinition> getSchedules(TenantIdentifier tenant, boolean loadProperties) 
+            throws DCSecurityException;
 
     @RolesAllowed(SecurityRoles.SCHEDULE_EDITOR)
     public ScheduleDefinition updateSchedule(TenantIdentifier tenant, ScheduleDefinition scheduleDefinition)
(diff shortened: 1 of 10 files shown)


monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -322,7 +328,28 @@ public ScheduleDefinition eval(InputStream inputStream) {
 
         return schedule;
     }
+    
+    private ScheduleDefinition getScheduleWithoutProperties(final TenantIdentifier tenant, 
+            final JobIdentifier jobIdentifier) {
+        final JobContext jobContext = getJobContext(tenant, jobIdentifier);
+        final String groupName = jobContext.getGroupName();
+        final ScheduleDefinition schedule = new ScheduleDefinition(tenant, jobIdentifier, groupName);
```
jhorcicka: The old functionality is still there with the second argument set to "true". But I have to check hot folder.

jhorcicka: Hot folder seems to be working properly.

monitor/services/src/main/java/org/datacleaner/monitor/scheduling/SchedulingService.java
```diff
@@ -42,7 +42,8 @@
 public interface SchedulingService extends RemoteService {
 
     @RolesAllowed({ SecurityRoles.VIEWER, SecurityRoles.SCHEDULE_EDITOR })
-    public List<ScheduleDefinition> getSchedules(TenantIdentifier tenant) throws DCSecurityException;
+    public List<ScheduleDefinition> getSchedules(TenantIdentifier tenant, boolean loadProperties) 
+            throws DCSecurityException;
```
ClaudiaPHI: Can we keep the old method as well and just redirected it to this one with "getSchedules(tenant, true)" ? Maybe someone else uses the method.

LosD: Not really a concern. It's a GWT class, so it not a public API (yet, hopefully at the end of the monitor rewrite, all interfaces to the web app are public)

LosD:
LGTM!

PR #1533, opened 2016-09-27
Find metrics from analyzers within outputdatastreams as well

### LINKED ISSUE #1437: In DC Monitor components producing output data streams break timeline metrics
If I have a job with a component which produces an output data streams (e.g. a Completeness Analyzer) and that job has some additional components (e.g. a Value Distribution analyzer) in the scope of such an output data stream and you run that job through DataCleaner monitor and then try to create a timeline for it, you can't select columns from the additional components for metrics in the timeline.

Illustrated by the screenshots below, where I have a job with a Completeness Analyzer and a Value Distribution analyzer and I can only select fields from the Completeness Analyzer when selecting metrics to monitor in DataCleaner monitor.

2016-07-14 10_54_27-customer_completeness analysis xml _ customers _ analysis job _ datacleaner
2016-07-14 10_55_19-datacleaner monitor

LosD:
This will make it possible to add timeline metrics for analyzers from all scopes, not just those in the root scope.

Fixes #1437

### DIFF
--- a/api/src/main/java/org/datacleaner/job/AnalysisJob.java
+++ b/api/src/main/java/org/datacleaner/job/AnalysisJob.java
@@ -19,11 +19,13 @@
  */
 package org.datacleaner.job;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Stream;
 
+import org.apache.metamodel.schema.Column;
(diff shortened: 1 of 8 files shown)


jhorcicka:
Tested and LGTM.

PR #1536, opened 2016-09-29
Feature/add plugable status labels

ClaudiaPHI:
This classes can be used to add any sort of panel to the right side of DC as info Panels. See DataCleanerCloud panel or News Items

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
@@ -55,6 +55,7 @@ import org.apache.commons.vfs2.FileSystemException;
 import org.apache.metamodel.schema.Column;
 import org.apache.metamodel.schema.Schema;
 import org.apache.metamodel.util.FileHelper;
+import org.apache.metamodel.util.Func;
 import org.datacleaner.Version;
 import org.datacleaner.actions.NewAnalysisJobActionListener;
 import org.datacleaner.actions.OpenAnalysisJobActionListener;
@@ -102,12 +103,14 @@ import org.datacleaner.util.WidgetFactory;
 import org.datacleaner.util.WidgetUtils;
 import org.datacleaner.util.WindowSizePreferences;
 import org.datacleaner.widgets.CollapsibleTreePanel;
+import org.datacleaner.widgets.CommunityEditionStatusLabel;
 import org.datacleaner.widgets.DCLabel;
 import org.datacleaner.widgets.DCPersistentSizedPanel;
 import org.datacleaner.widgets.DataCloudStatusLabel;
 import org.datacleaner.widgets.ExecuteButtonBuilder;
-import org.datacleaner.widgets.LicenceAndEditionStatusLabel;
 import org.datacleaner.widgets.NewsChannelStatusLabel;
+import org.datacleaner.widgets.PlugabblePanel;
+import org.datacleaner.widgets.PlugableRightPanelLabel;
 import org.datacleaner.widgets.PopupButton;
 import org.datacleaner.widgets.visualization.JobGraph;
 import org.jdesktop.swingx.JXStatusBar;
@@ -255,6 +258,8 @@ public final class AnalysisJobBuilderWindowImpl extends AbstractWindow implement
 
     private static final Logger logger = LoggerFactory.getLogger(AnalysisJobBuilderWindow.class);
     private static final ImageManager imageManager = ImageManager.get();
+    
+    public static final List<Func<AnalysisJobBuilderWindowImpl, JComponent>> PLUGGABLE_LABELS_COMPONENTS = new ArrayList<>(0);
 
     private static final int DEFAULT_WINDOW_WIDTH = 1000;
     private static final int DEFAULT_WINDOW_HEIGHT = 710;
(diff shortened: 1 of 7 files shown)


desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -255,6 +260,8 @@ public void onRemove(final InputColumn<?> sourceColumn) {
 
     private static final Logger logger = LoggerFactory.getLogger(AnalysisJobBuilderWindow.class);
     private static final ImageManager imageManager = ImageManager.get();
+    
+    public static final List<Func<AnalysisJobBuilderWindowImpl, JComponent>> PLUGGABLE_LABELS_BANNER_COMPONENTS = new ArrayList<>(0);
```
LosD: I know this is how we've done it in other places, but I'm really no fan of exposing what should be internal structures like this. Can't we use the ServiceLoader interface, or some other proper solution instead?

At least hide it behind an `addRightPanelComponent()` or something like that.

The name also seems a bit weird. It's not a banner?

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -765,6 +772,17 @@ private JComponent getWindowPanelContent() {
         final JXStatusBar statusBar = WidgetFactory.createStatusBar(_statusLabel);
         RightInformationPanel rightInformationPanel = new RightInformationPanel(_glassPane);
 
+        for (Func<AnalysisJobBuilderWindowImpl, JComponent> pluggableComponent : PLUGGABLE_LABELS_BANNER_COMPONENTS) {
+            final JComponent component = pluggableComponent.eval(this);
+            if (component != null) {
+                if (component instanceof JPanel) {
+                    final PlugabblePanel panel = (PlugabblePanel) component;
```
LosD: Checking for an instance of a JPanel, but casting to a PluggablePanel seems wrong to me.

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -765,6 +772,17 @@ private JComponent getWindowPanelContent() {
         final JXStatusBar statusBar = WidgetFactory.createStatusBar(_statusLabel);
         RightInformationPanel rightInformationPanel = new RightInformationPanel(_glassPane);
 
+        for (Func<AnalysisJobBuilderWindowImpl, JComponent> pluggableComponent : PLUGGABLE_LABELS_BANNER_COMPONENTS) {
+            final JComponent component = pluggableComponent.eval(this);
+            if (component != null) {
+                if (component instanceof JPanel) {
```
LosD: `null` wouldn't be an instance of JPanel, so the null check is a bit redundant

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
LosD: Oh, and you should never pass the Impl classes. Use the interface, that's what it is there for.

ClaudiaPHI: I could use the ServiceLoader, but wouldn't that be a mix of a class in the enterprise edition with the community edition ?

LosD: Not sure I understand? Using the ServiceLoader, you would provide an interface that any JAR could then implement. Then DataCleaner would just load all available implementations, and add them by itself.

For inspiration, see how we provide an implementation of SparkJobLifeCycleListener in the commercial spark package, and add it in `org.datacleaner.spark.Main` line 67.

Of course, this being a Swing class, the actual service needs to be some kind of factory that provides the `PluggablePanel` in `getPluggablePanel()` or similar method.

ClaudiaPHI:
Ready for review again
2016-09-30 15_32_35-welcome _ datacleaner
2016-09-30 15_22_03-welcome _ datacleaner

desktop/ui/src/main/java/org/datacleaner/windows/AnalysisJobBuilderWindowImpl.java
```diff
@@ -773,9 +776,20 @@ private JComponent getWindowPanelContent() {
...
-
-        final LicenceAndEditionStatusLabel statusLabel = new LicenceAndEditionStatusLabel(rightInformationPanel);
-        statusBar.add(statusLabel);
+        
+        if (Version.isCommunityEdition()) {
+            final CommunityEditionStatusLabel statusLabel = new CommunityEditionStatusLabel(rightInformationPanel);
+            statusBar.add(statusLabel);
+            statusBar.add(Box.createHorizontalStrut(20));
+        } else {
+            final ServiceLoader<PlugabblePanel> pluggablePanelsLoaders = ServiceLoader.load(PlugabblePanel.class);
```
LosD: Hmmmm... Something seems a bit off here. You should probably make a factory/descriptor for the icon and panel in the factory, like:
```
interface InformationPanelDescriptor {
    JComponent createInformationComponent(AnalysisJobBuilderWindow analysisJobBuilderWindow);
    String getLabelImagePath();
    String getLabelText();
}
```

... And load that using the service loader instead. That also removes the dependency on DCPanel, any JComponent is (and should be) usable with the RightInformationPanel.

I'd maybe also drop the PluggableRightPanelLabel name... The label itself is not pluggable, and the position can change (yes, I know RightInformationPanel is already misnamed in that regard, but let's not make it worse :)). How about just InformationPanelLabel?

LosD:
LGTM!

PR #1539, opened 2016-09-30
fixed the exception for jtds driver in case of single connections

### LINKED ISSUE #1201: Multithreaded inserting of rows into SQL Server fails

ClaudiaPHI:
fixed the exception for jtds driver in case of single connections

fixes #1201.


arjansh:
LGTM!

PR #1541, opened 2016-10-05
Improved dependent properties synchronization

### LINKED ISSUE #1531: Job with concatenator fails using run first N records
I have a job which contains a Plain search/replace, a Concatenator and an output write, which runs successfully when executed normally, but fails when run using the run first N records option. It results in a stacktrace which in itself doesn't explain very much:

```
13:48:16ERROR: An error occurred in the analysis job!

org.datacleaner.job.ComponentValidationException
	at org.datacleaner.descriptors.ValidateMethodDescriptorImpl.convertThrownException(ValidateMethodDescriptorImpl.java:42)
	at org.datacleaner.descriptors.AbstractMethodDescriptor.invoke(AbstractMethodDescriptor.java:83)
	at org.datacleaner.descriptors.ValidateMethodDescriptorImpl.validate(ValidateMethodDescriptorImpl.java:37)
	at org.datacleaner.lifecycle.LifeCycleHelper.validate(LifeCycleHelper.java:160)
	at org.datacleaner.job.tasks.InitializeTask.executeInternal(InitializeTask.java:71)
	at org.datacleaner.job.tasks.InitializeTask.execute(InitializeTask.java:55)
	at org.datacleaner.job.concurrent.TaskRunnable.run(TaskRunnable.java:61)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
```

arjansh:
Fixes #1531.

Made the algorithm which maps an old input column to a new input column a bit less strict to prevent inserting null values when synchronizing dependent values when a transformer's properties are updated.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
+++ b/engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
@@ -445,19 +445,19 @@ public abstract class AbstractComponentBuilder<D extends ComponentDescriptor<E>,
 
                                 if (dependentValue != null) {
                                     // First build a list containing value and references tuples.
-                                    final Map<Integer, Object> originalMappings = new HashMap<>();
+                                    final Map<String, Object> originalMappings = new HashMap<>();
 
                                     final List<Object> synchronizedDependents = new ArrayList<>();
 
                                     if (currentValue.getClass().isArray()) {
                                         for (int i = 0; i < Array.getLength(currentValue); i++) {
-                                            originalMappings.put(Array.get(currentValue, i).hashCode(), Array.get(
+                                            originalMappings.put(getKey(Array.get(currentValue, i)), Array.get(
                                                     dependentValue, i));
                                         }
 
                                         for (int i = 0; i < Array.getLength(newValue); i++) {
-                                            synchronizedDependents.add(originalMappings.get(Array.get(newValue, i)
-                                                    .hashCode()));
+                                            synchronizedDependents.add(originalMappings.get(getKey(Array.get(newValue,
+                                                    i))));
                                         }
 
                                         dependentProperty.setValue(_configurableBean, getArray(dependentProperty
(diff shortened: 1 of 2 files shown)


engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
```diff
@@ -472,6 +472,18 @@ private void synchronizeDependentProperties(final ConfiguredPropertyDescriptor p
         }
     }
 
+    private static String getKey(final Object object) {
+        if (object instanceof InputColumn<?>) {
+            final InputColumn<?> inputColumn = (InputColumn<?>) object;
+
+            if (inputColumn.isVirtualColumn()) {
+                return inputColumn.getName();
```
LosD: A potential danger here is that we don't disallow the same column name from different components, so it's not guaranteed to be unique. OTOH, I can think of lots of other things that can go wrong in that case, so might not really be an issue.

LosD:
LGTM, but please consider if my reservation in code comment is a real danger. Merge if it isn't :)

engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
jhorcicka: Is it the following case? I used the same name for both concatenators' outputs and there is a message "not unique columns". 
same-input-column-names

jhorcicka:
Functionality works at least for the provided job (see #1531), which was not the case in the previous version.

engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
LosD: Ah, but that's the CSV file writer that won't produce a file with non-unique columns. Nothing to do with the job. If the job had disallowed it, you wouldn't have had two columns with the same name in the first place.

jhorcicka: OK, then let's wait for @arjansh's expert opinion.

arjansh:
@losd: regarding the potential issue: the essential use case for the  synchronizeDependentProperties methods is to remove mappings to columns which have just been removed from the model, so I see no danger here.

LosD:
Check. Let's merge. @jhorcicka, agree?

PR #1550, opened 2016-10-10
Allow all non-file datastore in isValidSourceDatastore()

### LINKED ISSUE #1549: Network-based datastore ruled out
In #1261 we made an (overall good) change to help users not start a long-running job on a cluster if the job wasn't configured properly.

Unfortunately, we were a little too aggressive, and ruled out _all_ datastores not Csv, Json or fixed-width. That means that network-based datastores, and especially Hadoop centric datastores (Hive, HBase, Impala etc) cannot be run on Spark using the GUI.

LosD:
This should allow most datastores we can't rule out to work. Will also make the running from GUI a little less fool-proof, but that's the lesser of two evils

Fixes #1549


LosD:
HALT! Do NOT merge! This will never work without changes to SparkRunner

LosD:
Closing, this is not a small thing. We probably need to restructure a lot to make this work.

PR #1551, opened 2016-10-12
#1499 DCPopupPanel dialog's width was extended

### LINKED ISSUE #1499: DC Monitor: inspect profiling result -- overflowing text
DC Monitor, Dashboard, timeline created. When I click the rounded end of any of the lines in the graph (Inspect profiling result), a dialog appears. It contains 3 buttons and the text of the middle one overflows. 

narrow-button

jhorcicka:
Fixes #1499 

DCPopupPanel's width was extended from 800px to 950px because of the overflowing text in "Inspect profiling result". 

950px


LosD:
I actually hate this... But it's better than all the alternatives, and we want simple changes now, so LGTM :)

Let's merge when Travis is happy

PR #1553, opened 2016-10-13
Add docker integration test

### LINKED ISSUE #1542: Setup real job run test on DataCleaner Monitor using Docker
Add an integration test to datacleaner-monitor-ui, which deploys the DataCleaner monitor war to a Docker image, then deploy a job on it through a REST endpoint and then runs that job through a REST endpoint and validates the result.

It seems most logical to have this integration test be part of a profile, so it typically only runs on Continuous integration servers like Jenkins and Travis. You should also be able to run it locally on your own machine however.

arjansh:
Fixes #1542.

Use the docker-maven-plugin to build and start a docker image to which DataCleaner-monitor is deployed and test it in an integration test by doing some REST(like) calls to it and validating the results.

Run the integration test using the integration-test profile, e.g. `mvn clean verify -Pintegration-test`.

Furthermore the following environment variables has to be set:

- DOCKER_HOST_NAME - contains the host name of your Docker Daemon, e.g. localhost or 192.168.99.100.

And optionally:

- DOCKER_HOST - contains the URL of the Docker Daemon, e.g. http://localhost:2375 or tcp://192.168.99.1000:2376.
- DOCKER_CERT_PATH - contains the path to the SSL certificate when SSL is used for communicating with the Docker daemon, e.g. ~/.docker/ or C:\Users\[username]\.docker\machine\certs.

You can also override the default tomcat port (8080), using the dockerTomcatPort maven property, e.g. `mvn clean verify -Pintegration-test -DdockerTomcatPort=8081`.

Also did a little change to DataCleaner-env-spark to make sure the unit tests run successfully regardless of the order in which they are run.


LosD:
Description could use an update, but PR itself LGTM. Merging!

PR #1554, opened 2016-10-14
Update requirements button visibility and text on scope change

### LINKED ISSUE #826: Requirements button in configuration dialog does not update on scope change
The requirements button in the configuration dialog does not reflect a change of scope.
- A button that is visible/invisible because there is/isn't any filters in the scope, keeps that state.
- A button that is showing an active filter, does not change to _(no requirement)_ when changing scope (as it should, as the filter is not valid in the new scope).

LosD:
The requirements button relevancy is no longer only check on initial creation of the top banner.

 Fixes #826


jhorcicka:
Tested and LGTM.

PR #1555, opened 2016-10-14
Set a minimum size for dialogs

### LINKED ISSUE #519: OutputDataStreams UI graph support

### LINKED ISSUE #1519: Some desktop windows can be shrinked to almost invisible size
In many cases dialog and other windows can be shrinked by the user to almost invisible size. I think it would be nice to introduce a "minimal size" that would prevent this. 

This was firstly mentioned at #234 but it is not tightly related to that issue.

LosD:
Currently uses the result of getDialogWidth, which might be too wide in some instances, though I haven't seen any yet.

Fixes #1519

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/windows/AboutDialog.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/AboutDialog.java
@@ -107,7 +107,7 @@ public class AboutDialog extends AbstractDialog {
 
     @Override
     protected boolean isWindowResizable() {
-        return true;
+        return false;
     }
 
     @Override
--- a/desktop/ui/src/main/java/org/datacleaner/windows/AboutDialog.java
+++ b/desktop/ui/src/main/java/org/datacleaner/windows/AboutDialog.java
@@ -102,12 +102,12 @@ public class AboutDialog extends AbstractDialog {
 
     @Override
     protected int getDialogWidth() {
-        return 600;
+        return 650;
     }
 
     @Override
     protected boolean isWindowResizable() {
-        return false;
+        return true;
     }
 
     @Override
--- a/desktop/api/src/main/java/org/datacleaner/windows/AbstractDialog.java
+++ b/desktop/api/src/main/java/org/datacleaner/windows/AbstractDialog.java
@@ -22,6 +22,7 @@ package org.datacleaner.windows;
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Component;
+import java.awt.Dimension;
 import java.awt.Image;
 import java.awt.event.ActionEvent;
 import java.awt.event.KeyEvent;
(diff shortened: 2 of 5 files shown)


LosD:
Whoops, branch name and commit comment references the wrong issue (519, seems I missed a 1).

PR description was also wrong, but that is fixed now

desktop/ui/src/main/java/org/datacleaner/windows/AboutDialog.java
```diff
@@ -107,7 +107,7 @@ protected int getDialogWidth() {
 
     @Override
     protected boolean isWindowResizable() {
-        return true;
+        return false;
```
arjansh: Why should the About dialog no longer be resizable?

jhorcicka:
I think the change influenced "only" component configuration windows and reference-data dialogs. 

Browse file:
browse
Configuration dialog (OK):
configuration
Job window: 
job-window
Main window (probably the same as job-window):
main-window
Open file (probably the same as browse-file):
open
Options dialog:
options-dialog
Reference data (may be okay, but 400 or 350 instead of 300 might even be better): 
reference-data

desktop/ui/src/main/java/org/datacleaner/windows/AboutDialog.java
LosD: Because it is stupid. It's not something you work with. You read it, you close it.

LosD: Made me take another look, though. We should probably have a little bit bigger window for the sake of the licenses.

LosD: Hmmm... Okay, okay. The license reading pretty much makes it something you might want to resize. Dangit. I hate even having chrome on about dialogs.

Oh well.

LosD:
Okay, minimums added.

I'm not entirely sure what you mean with the ref data window size. It currently has a minimum of 600x200?, so I'm not sure where 300 fits in? Anyway, let's leave it alone for now. At least you can't make it invisible :)

jhorcicka:
```
setMinimumSize(new Dimension(getDialogWidth(), 200));
```
Here I wanted 
```
setMinimumSize(new Dimension(getDialogWidth(), 300));
```
so that user can see a bit more than just the yellow description box (see ref-data screenshot above).

LosD:
Ah, yeah, I can do that. I was just confused because you wrote _instead of 300_ :)

LosD:
@jhorcicka I think you can re-review. File chooser should also respect min size now :)

jhorcicka:
@LosD: Yes, sorry for that 300->200. 

Tested and LGTM.

PR #1556, opened 2016-10-14
Added caches to SourceColumnFinder to prevent unneccesary recursion

### LINKED ISSUE #1552: Component count in DC has drastic effect on execution startup time and overhead
A customer has pretty complex jobs, and the number of components seems to affect execution drastically:

Time is for the whole job, but since there is no data in the input, I've kept it, since it is still a significant overhead.

#### 14 attributes, 2-3 seconds 
14attr-new
----
#### 15 attributes, 4-5 seconds
15attr-new
----
#### 16 attributes, 8-9 seconds
16attr-new
----

#### 17 attributes, 8-9 seconds
This one is the only one where execution time doesn't seem to change. Not sure why... If you look at startup time, it seems faster, it's just that the overhead in the actual execution makes the total time roughly the same
17attr-new
----
#### 18 attributes, 21-22 seconds
18attr-new

arjansh:
Fixes #1552.

When a job is executed the RowProcessingPublishers class partitions the job into RowProcessingPublishers, part of this is registering the job on a RowProcessingPublishers object. When doing this it uses a SourceColumnFinder to register the relevant columns to a RowProcessingPublishers object. On the SourceColumnFinder the findOriginatingColumns is invoked which goes into a recursion to determine which columns in a job are the "origin" of a given column. The more components the job has the more often this is caused and the deeper the recursion gets.

This pull request fixes this by putting objects for which the originating Columns have been found into an internal cache on the SourceColumnFinder so they don't have to be retrieved more than once. 

I also refactored the signature of some of the internal methods a bit to make this possible. A set was passed on to store the results in, but that made it hard for me to add a result to a cache at some moment.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/SourceColumnFinder.java
+++ b/engine/core/src/main/java/org/datacleaner/util/SourceColumnFinder.java
@@ -47,8 +47,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Helper class for traversing dependencies between virtual and physical
- * columns.
+ * Helper class for traversing dependencies between virtual and physical columns.
+ *
+ * For performance reasons this class stores found sources in an internal cache. As there is no mechanism to
+ * invalidate or refresh this cache, instances of this class should not be assigned to fields of other
+ * classes.
  */
 public class SourceColumnFinder {
 
--- a/engine/core/src/main/java/org/datacleaner/util/SourceColumnFinder.java
+++ b/engine/core/src/main/java/org/datacleaner/util/SourceColumnFinder.java
@@ -61,8 +61,8 @@ public class SourceColumnFinder {
     private Set<HasFilterOutcomes> _outcomeSources = new HashSet<HasFilterOutcomes>();
     private Set<HasComponentRequirement> _outcomeSinks = new HashSet<HasComponentRequirement>();
 
-    private Map<InputColumn<?>, Set<Column>> originatingColumnsOfInputColumnCache = new HashMap<>();
-    private Map<Object, Set<Column>> originatingColumnsOfSourceCache = new HashMap<>();
+    private final Map<InputColumn<?>, Set<Column>> originatingColumnsOfInputColumnCache = new HashMap<>();
+    private final Map<Object, Set<Column>> originatingColumnsOfSourceCache = new HashMap<>();
 
     private void addSources(Object... sources) {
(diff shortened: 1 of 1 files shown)


arjansh:
One possible risk I see, which is currently non-existent in the code base however is that a SourceColumnFinder is stored in a field member of a class and its caches become stale. Currently SourceColumnFinder objects are never stored in field members, so this poses no problem right now.

engine/core/src/main/java/org/datacleaner/util/SourceColumnFinder.java
```diff
@@ -59,6 +61,9 @@
     private Set<HasFilterOutcomes> _outcomeSources = new HashSet<HasFilterOutcomes>();
     private Set<HasComponentRequirement> _outcomeSinks = new HashSet<HasComponentRequirement>();
 
+    private Map<InputColumn<?>, Set<Column>> originatingColumnsOfInputColumnCache = new HashMap<>();
+    private Map<Object, Set<Column>> originatingColumnsOfSourceCache = new HashMap<>();
+
```
ClaudiaPHI: finals

engine/core/src/main/java/org/datacleaner/util/SourceColumnFinder.java
```diff
@@ -288,43 +291,65 @@ private Table findOriginatingTableOfSource(Object source, Set<Object> resolvedSe
...
+        if (cachedOriginatingColumns != null) {
+            return cachedOriginatingColumns;
         }
-        if (source instanceof InputColumnSinkJob) {
-            InputColumn<?>[] input = ((InputColumnSinkJob) source).getInput();
-            if (input != null) {
-                for (InputColumn<?> inputColumn : input) {
-                    findOriginatingColumnsOfInputColumn(inputColumn, result);
+
+        Set<Column> originatingColumns = new HashSet<>();
```
ClaudiaPHI: final

ClaudiaPHI:
Is there any way we can test this PR ?

arjansh:
@ClaudiaPHI Nothing functionally changed. Besides some monkey testing by running different kind of jobs, we probably just have to trust the current suite of unit tests.

LosD:
@arjansh As we discussed offline, there's probably no simple and performant way around your worry. Let's JavaDoc our way out of it, something like "This uses caching of columns to improve performance, please don't store instance for later reuse".

LosD:
LGTM!

PR #1558, opened 2016-10-17
Bug/153 multiple mapped input columns property

jhorcicka:
Fixes https://github.com/NCIM/datacleaner-commercial-editions/issues/153

MultipleMappedInputColumnsPropertyWidget was not updated correctly when a component (which columns were used in the widget) was removed from the graph.

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleMappedStringsPropertyWidget.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleMappedStringsPropertyWidget.java
@@ -210,15 +210,12 @@ public class MultipleMappedStringsPropertyWidget extends MultipleInputColumnsPro
             return;
         }
         
-        SwingUtilities.invokeLater(new Runnable() {
-            @Override
-            public void run() {
-                _alreadyUpdating = true;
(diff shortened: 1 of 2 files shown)


ClaudiaPHI:
LGTM! I have tested it and it seems to work fine.

desktop/ui/src/main/java/org/datacleaner/widgets/properties/MultipleMappedStringsPropertyWidget.java
```diff
@@ -215,6 +204,23 @@ public MappedStringsPropertyWidget getMappedStringsPropertyWidget() {
         }
         return result.toArray(new InputColumn[result.size()]);
     }
+    
+    public void updateMappedStrings() {
+        if (_alreadyUpdating) {
+            return;
+        }
+        
+        SwingUtilities.invokeLater(new Runnable() {
```
LosD: Could be replaced with lambda

LosD:
LGTMtoo, but you could replace a single variable with a lambda.

jhorcicka:
@LosD, updated.

LosD:
LGTM. Travis was happy before, and probably still is, but let's wait for it to be happy again, then merge :)

PR #1565, opened 2016-10-20
No longer normalize columnType's path attribute when reading a job

### LINKED ISSUE #1564: Unable to load job with Excel datastore with multiline headers
When you register an Excel sheet with line feeds in the first row as a datastore, build a job using that datastore, save the job and then reload the job, you get an unexpected error.

Steps to reproduce:
- Register a datastore using the attached double.xlsx Excel sheet.

[double.xlsx](https://github.com/datacleaner/DataCleaner/files/541247/double.xlsx)

- Create a job using that datastore.
- Add Sheet1 from the double.xlsx datastore to the job builder.
- Add a Pattern finder and select ColumnA and Column B as input columns.
- Save the job.
- Close the job.
- Try to open the job again. Now you get an unexpected error and the job can't be opened.
2016-10-20 11_03_51-unexpected error

arjansh:
Fixes #1564.

The job XML Schema Definition defines the path of a columnType to be a "normalizedString", as a result of which during the unmarshalling of jobs column names which contain line feeds are normalized and line feeds are removed/replaced with spaces. As a result of this the column names can be different when a job is read and the column names can't be matched against the datastore's column names.

The fix is to remove the "normalizedString" type from the columnType's path attribute, so the column name's are the same when reading a job from a file as they were when it was written to file.


LosD:
I think this is another case of "I hate it, and it scares me, but I think the alternatives are worse".

(The alternatives here being that either we start doing mapping to and from the real column names, or MetaModel does)

LGTM

PR #1566, opened 2016-10-20
Copy example repo to temp dir in unit tests

### LINKED ISSUE #1562: Errors when setting up monitor-services unit tests on Windows
When running the unit tests of the monitor-services module on Windows, unit tests sometimes (randomly) fail with an IOException like this one:

```
testPostWithOverrideProperties(org.datacleaner.monitor.server.controllers.JobTriggeringControllerIntegrationTest)  Time elapsed: 0.07 sec  <<< ERROR!
java.io.IOException: Unable to delete directory target\example_repo\tenant4.
        at org.apache.commons.io.FileUtils.deleteDirectory(FileUtils.java:1541)
        at org.apache.commons.io.FileUtils.forceDelete(FileUtils.java:2270)
        at org.apache.commons.io.FileUtils.cleanDirectory(FileUtils.java:1653)
        at org.apache.commons.io.FileUtils.deleteDirectory(FileUtils.java:1535)
        at org.datacleaner.monitor.server.controllers.JobTriggeringControllerIntegrationTest.testPostWithOverrideProperties(JobTriggeringControllerIntegrationTest.java:59)
```

For some reason the directory can't be deleted. The unit test should be set up in different manner, so it doesn't fail because it can't set up the folder structure it uses during the test.

arjansh:
Fixes #1562.

Use a FileRepository which copies the example repo to a real temp dir to prevent IOExceptions when running tests on Windows.

### DIFF
--- a/monitor/services/src/test/java/org/datacleaner/monitor/scheduling/quartz/ExecuteJobTest.java
+++ b/monitor/services/src/test/java/org/datacleaner/monitor/scheduling/quartz/ExecuteJobTest.java
@@ -47,6 +47,8 @@ import org.quartz.JobDetail;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.support.ClassPathXmlApplicationContext;
 
+import com.google.common.io.Files;
+
 import junit.framework.TestCase;
 
 public class ExecuteJobTest extends TestCase {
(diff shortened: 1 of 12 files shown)


ClaudiaPHI:
LGTM!

PR #1570, opened 2016-10-21
#1569 Docker-integration tests: AuthenticationIT problem

### LINKED ISSUE #1569: Docker testing: authentication problem in AuthenticationIT test
When I try to run current docker-based-integration-tests, it fails with "401 expected, 200 received"-like error. 

The problem seems to be in the order of other tests. The first one successfully authenticates and the other one expects to be logged-out which is not true.

jhorcicka:
Fixes #1569 

AuthenticationIT test now uses default authentication in "before" setup which assures that nobody is logged in before the actual testing requests.


arjansh:
LGTM!

PR #1573, opened 2016-10-21
 Hide column names panel everywhere

### LINKED ISSUE #1571: Hide column naming from Excel, CSV and FW
There still isn't a MetaModel release out that has the column naming in place. Since we're releasing DC shortly, we'll need to hide the changes that puts that in place until MM 4.5.5 release.

ClaudiaPHI:
fixes #1571


LosD:

PR #1575, opened 2016-10-21
Simplified REST API - automatic column mapping by map key

jakubneubauer:
REST component API - if input is a map and no input columns config is given, try to map the values it to input columns, taking key as a input column property name.

Example:

When doing a REST query for Phone correction, now it is possible send this payload:
```
{
  "data": [
    {
      "Phone number column": "234234234"
    }
  ]
}
```
Previously, it was necessary provide definition of columns, provide configuration for Input column property and then the data, like this:
```
{
    "configuration":{
        "properties":{
            "Phone number column":"Phone"
        },
        "columns":[
            {"name":"Phone"}
        ]
    },
    "data": [
        ["+420739583601"]
    ]
}
```


monitor/services/src/main/java/org/datacleaner/monitor/server/components/InputColumnAndMappedPropertyRewriter.java
```diff
@@ -45,58 +47,138 @@ public boolean rewriteInput(TransformerDescriptor<?> transformer, ProcessStatele
...
+                values.add(value);
+                continue;
+            }
+            String[] aliases = inputColPropDesc.getAliases();
+            if(aliases != null) {
+                for (String alias : aliases) {
+                    value = item.get(alias);
+                    if(value != null) {
+                        values.add(value);
+                        continue ITER;
```
LosD: I'm sorry, but that's going to be a "nope", unless you have a very good reason to used a labeled continue.

Why not just have the `values.add(json.nullNode());` in an else and use break? Might also be a sign that something like a findAlias method could be a good idea.

monitor/services/src/main/java/org/datacleaner/monitor/server/components/InputColumnAndMappedPropertyRewriter.java
```diff
@@ -45,58 +47,138 @@ public boolean rewriteInput(TransformerDescriptor<?> transformer, ProcessStatele
...
+            // OK, we have one object
+        } else {
+            return false;
+        }
+
+        JsonNodeFactory json = Serializator.getJacksonObjectMapper().getNodeFactory();
+        if(input.data.isArray()) {
+            int i = 0;
+            for(JsonNode item: input.data) {
+                ((ArrayNode)input.data).set(i++, mapDataObjectToInputColumnsByName(item, inputColProperties));
```
LosD: Just making sure: Is it safe to mutate an ArrayNode you're traversing? I guess so, since you're not adding or deleting elements.

BTW: If you need an indexer, why not use a traditional for loop? This is like the worst of both worlds.

monitor/services/src/main/java/org/datacleaner/monitor/server/components/InputColumnAndMappedPropertyRewriter.java
```diff
@@ -45,58 +47,138 @@ public boolean rewriteInput(TransformerDescriptor<?> transformer, ProcessStatele
...
+            input.data = json.arrayNode().add(mapDataObjectToInputColumnsByName(input.data, inputColProperties));
+        }
+
+        if(input.configuration == null) {
+            input.configuration = new ComponentConfiguration();
+        }
+
+        int i = 1;
+        for(ConfiguredPropertyDescriptor inputCol: inputColProperties) {
+            String colName = "c" + i++;
```
LosD: Same thing with a traditional for loop here.

jakubneubauer:
@LosD , can you check and eventually merge this pretty old branch, please?

LosD:
LGTM!

PR #1576, opened 2016-10-21
#1563 docker-testing: jobs from Monitor were added.

### LINKED ISSUE #1563: Docker testing: add tests for all sample jobs in DC Monitor
Run tests of sample jobs using our new Docker testing, except duplicate detection and name correction.

jhorcicka:
Resolves #1563

Jobs that are already available in Monitor were added to integration testing.


ClaudiaPHI:
I am not sure I like having copies of jobs. It make easy to make mistakes. Forget to make the changes in both files. Isn't it possible to give the normal path ?

LosD:
I agree with @ClaudiaPHI. we really want to test the same jobs the users see, so we're also making sure any edits to those would continue working.

jhorcicka:
In the proof of concept @arjansh used a file upload as well, that's why there were copies of the jobs. 

I removed the file upload part and added "real" names of the jobs so that now the available (uploaded) jobs are tested, @LosD and @ClaudiaPHI.

ClaudiaPHI:
LGTM!

PR #1584, opened 2016-10-25
#1579 AbstractWindow: minimal size.

### LINKED ISSUE #1579: Preview window does not have a minimal size
When you "preview" for example datastore content in the job, the dialog window does not seem to have a minimal size set. 

before

after

jhorcicka:
Fixes #1579 

setMinimumSize(...) was added in AbstractWindow. 

Now e. g. "preview" window or "analysis result" window can not be shrinked to nothing anymore.


ClaudiaPHI:
Tested and it works as intended.  LGTM!

PR #1590, opened 2016-10-28
#1500 Component group removal

### LINKED ISSUE #1500: Group removal of components
Duplicate of #312

When I select more components in the job graph and try to remove them by pressing "delete" key then sometimes it works but usually it does not. Selecting components one by one by a single click and then pressing "delete" works fine. 

Sometimes it seems that selecting another set of components followed by pressing "delete" helps but it is not a rule. 

multiple-deletion

jhorcicka:
Fixes #1500  

JobGraphKeyListener was replaced by JobGraphBindingsManager (listener was replaced by key-bindings).


desktop/ui/src/main/java/org/datacleaner/widgets/visualization/JobGraphBindingsManager.java
```diff
@@ -0,0 +1,197 @@
...
+                        analysisJobBuilder.removeSourceTable(table);
+                    } else if (vertex instanceof Column) {
+                        final Column column = (Column) vertex;
+                        analysisJobBuilder.removeSourceColumn(column);
+                    }
+                }
+            }
+        });
+        // register BACKSPACE key to the same action
+        _component.getInputMap(INPUT_MAP_CONDITION).put(KeyStroke.getKeyStroke(KEY_BACKSPACE), KEY_BACKSPACE);
```
LosD: Does this really work? Shouldn't it be KEY_DELETE as the action key?

LosD: Tested, nope it doesn't.

jhorcicka:
@LosD, what do you mean by "At least the BACKSPACE_KEY should be changed"? 

What does not work for you with KEY_BACKSPACE?

ClaudiaPHI:
Tested, works as it should

Issue #1589, opened 2016-10-28
Convert to date transformer gives different results on different systems


arjansh:
Depending on what system (e.g. your computer, another computer, Travis CI) you run a job with a `Convert to date` transformer in it, it's results may in practice vary.

E.g. when I run the sample `Customer age analysis` job, which is provided with DataCleaner, I run the job and I inspect the results of the Month distribution, on my computer it reports 460 occurrences of January, when a colleague runs it, it reports 462 occurrences and when Travis CI runs it through an integration test it reports 464 occurrences. I assume this is caused by the `Convert to date` transformer as that component is responsible for transforming the string represented `birthdate` field's values to dates.

This issue was found when implementing #1577.

PR #1594, opened 2016-10-31
Integration tests for sample jobs

### LINKED ISSUE #1577: Have desktop's jobs tested by integration tests
Jobs currently not available in Monitor by default but available in Desktop should be added into integration testing process. 

Relates to #1563

arjansh:
Fixes #1577.

Adds integration test for the nine sample jobs in DataCleaner which work out of the box. Runs the jobs in the same manner as the Command Line Interface.

### DIFF
--- a/desktop/ui/pom.xml
+++ b/desktop/ui/pom.xml
@@ -85,6 +85,28 @@
 		</plugins>
 	</build>
 
+	<profiles>
+		<profile>
+			<id>integration-test</id>
+			<build>
+				<plugins>
+					<plugin>
+						<artifactId>maven-failsafe-plugin</artifactId>
+						<version>2.19.1</version>
+						<executions>
+							<execution>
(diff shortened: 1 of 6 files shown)


LosD:
LGTM, but a single consideration (that can easily be moved later, though).

ClaudiaPHI:
LGTM!

ClaudiaPHI:
The "Customer  age " job fails with: 
image

ClaudiaPHI:
This test fails on my machine. I do not think we can merge it yet until #1589 is fixed.

arjansh:
@ClaudiaPHI I don't think your test failure is caused by #1589, but by #1596. I made the test a little less strict (and added a TODO) to work around that issue.

ClaudiaPHI:
LGTM! Tests pass

PR #1602, opened 2016-11-08
#1601 ColumnMeaning update

### LINKED ISSUE #1601: ColumnMeaning update
Update of org.datacleaner.metadata.ColumnMeaning: "job title" vs "person title", "landline" vs "mobile phone".

jhorcicka:
Resolves #1601 

"Job title" vs "person title"
"landline" vs "moble"

### DIFF
--- a/api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
+++ b/api/src/main/java/org/datacleaner/metadata/ColumnMeaning.java
@@ -49,6 +49,8 @@ public enum ColumnMeaning implements HasName, HasAliases {
 
     PERSON_NAME_FULL("Full name", "Name", "Person name", "Contact", "Contact name"),
 
+    PERSON_TITLE("Person title", "Title", "Salutation"),
+
     PERSON_GENDER("Gender", "Sex"),
 
     PERSON_AGE("Age", "Customer age", "Contact age"),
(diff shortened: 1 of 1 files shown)


LosD:
Hmmm... The more I think about this, the more unsafe it seems to me. If you move aliases from an existing value to a new one, extensions needs to be updated, or they might stop working with the default . E.g. one that expects `PHONE_PHONENUMBER` for all phone numbers would suddenly not get mobile numbers.

ClaudiaPHI:
I do not know what to say. Did any client requested for a "mobile" phone field. Do they want a division based on that ?

jhorcicka:
Well, I need to automatically map two separate input columns (land line, mobile) and provide enhanced versions to the output (again, separated in two or more columns). I might be able to leverage PhoneCorrection's "land type" output column if this PR is a no-go.

LosD:
Wont really solve the issue. If _any_ phone numbers (including mobile) changes to a new enum value, extensions will break.

LosD:
Wait a minute... Now I'm in doubt. We have several of these in different projects. It might actually only be a case of updating the Wizards for this one, and then it's quite a minor issue.

LosD:
(For the long term, we'll still need better solutions. This is supposed to be the "master" list of meanings, though I don't think it has taken effect yet)

LosD:
Okay, this seems to have been a non-issue, since nothing else is using the field currently. Merging! :)

PR #1606, opened 2016-11-10
Use a temporaryFolder for testing jobs

### LINKED ISSUE #1605: CLI integration tests uses user's conf.xml instead of a local file
Because it uses the conf.xml from the user's datacleaner home, the recently added integration tests are not as independent of the environment as they should be.

Furthermore, since they aren't ran from a repository entirely inside target/ or a temp folder, they will change the source repository.

LosD:
This makes sure that jobs are using a completely fresh copy of the testing repository for every test, which is independent of both developers DC home, and between tests.

Fixes #1605

### DIFF
--- a/desktop/ui/src/test/java/org/datacleaner/cli/JobTestHelper.java
+++ b/desktop/ui/src/test/java/org/datacleaner/cli/JobTestHelper.java
@@ -28,6 +28,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


desktop/ui/src/test/java/org/datacleaner/cli/JobTestHelper.java
```diff
@@ -76,9 +77,13 @@ public static void testJob(final String jobFileName, final Map<String, String[]>
         }
     }
 
-    private static String runJob(final String jobFileName) throws Exception {
+    private static String runJob(final File repository, final String jobName) throws Exception {
+        final String jobFileName =
+                URLDecoder.decode(new File(repository, "jobs/" + jobName + ".analysis.xml").getPath(), "UTF-8");
+        final String confFileName =
+                URLDecoder.decode(new File(repository, "conf.xml").getPath(), "UTF-8");
```
arjansh: The same logic is repeated here twice, maybe extract it to a helper method?

LosD:
Ready for re-review

arjansh:
LGTM! I'll merge when Travis is green.

PR #1613, opened 2016-11-16
#1612 Text fields in e. g. CSVWriter are editable again

### LINKED ISSUE #1612: Unable to edit writer output column names
Fix to non-updating MultipleMappedInputColumnsPropertyWidget introduced a problem with text fields listeners (e. g. in CSVWriter). User can not edit these text fields any more.

jhorcicka:
Fixes #1612 

Changes related to not-updating text fields in MultipleMappedStringsPropertyWidget were reverted because they make text fields not editable.


LosD:
A few differences from a straight revert, but that's only formatting and the addition of a logger, so that's fine. Do you want this actually merged to master, or should we only use it for the release, and then another PR would fix the original code instead?

jhorcicka:
Let's wait a bit and see if I can find something better than this. So, right now use it for the release without merging to the master (which we can do later anyway, I guess).

LosD:
Weird. My local build fails with a JobTestHelper error, which seems unrelated. But Travis is happy, at least for the push.

LosD:
Ah, now it's happy. Not sure what went wrong.

LosD:
LGTM! Won't merge, though (see previous discussion) :)

PR #1625, opened 2016-11-28
Fix code format and automate checking

### LINKED ISSUE #1622: Automate code format checking and reformat entire codebase to follow
DC needs some automat checking of code format.

LosD:
Lots and lots of changes here. Probably impossible to fully review, but checking that it's working, that the CheckStyle rules are valid (and as complete as possible), and that all our wants are covered should be a minimum.

I've tried to avoid changes that could change behavior, but in some cases, like missing default labels, I've had to make decisions.

There's quite a bit of checkstyle rules ignored. Either because of problems with checkstyle, too many issues to fix in a short time, inconsistency between checkstyle, or stuff that needed behavioral changes (finals among these).

Fixes #1622


LosD:
BTW, to whoever does the actual merge. _Please_ make sure to squash, and edit the first line of squash commit message to add something like "IF YOU ARE LOOKING FOR A LOGICAL CHANGE, LOOK BEFORE THIS COMMIT", (or something to that effect) and a line after explaining that this is only changing formatting and similar things.

This ensures that anyone looking for sources of bugs doesn't have to spend time trying to understand what this did.

PR #1628, opened 2016-11-28
Set DATACLEANER_HOME to the temp folder

### LINKED ISSUE #1627: CLI integration testing does not use own DataCleaner home
The CLI integration tests use the developer/testers own DataCleaner home, instead of a temporary folder. That both makes the developers environment affect the test, and creates a risk of tests affecting each other.

LosD:
This will always use the temporary repository folder as DATACLEANER_HOME, so that files cannot conflict with developer home and other tests.

Fixes #1627


ClaudiaPHI:
LGTM!

PR #1629, opened 2016-11-28
update the source column in outputstream job builder

### LINKED ISSUE #1616: Output datastreams does not update component output column name changes
If a component changes the name of its output column after that column has been added to a component that produces an output datastream, the output datastream will not update that name (even though it seems so, looking at the list of output datastream columns).
Output datastream list after column changed name:
image

Available columns according to component attached to that output datastream:
image

Besides the job generally not working properly, the saved job will be too corrupted to be loaded.

#### Steps to reproduce
1. Start job based on Customers datastore
2. Add concatenator to canvas.
3. Concatenate first and last name columns, name output column "complete name".
4. Add completeness analyser to canvas. 
5. Select "complete name" as the column to analyze.
6. Open concatenator and change name of output column to "full name"
7. Add Pattern finder to canvas
8. Change scope of Pattern finder to "Complete rows".
9. Notice name of available input column, select available input column.

#### Actual result
Available input column is named "complete name". If job is run, it patterns finds `<null>`s on a `<null>` column. If job is saved, it cannot be loaded again since "complete name" does not exist.

#### Expected result
Available input column is named "full name". If job is run, a proper distribution of names is generated, if job is saved it can be loaded again.

ClaudiaPHI:
fixes #1616.
 update the source column in outputstream job builder

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
+++ b/engine/core/src/main/java/org/datacleaner/job/builder/AbstractComponentBuilder.java
@@ -50,6 +50,7 @@ import org.datacleaner.api.Renderable;
 import org.datacleaner.configuration.DataCleanerConfiguration;
 import org.datacleaner.configuration.InjectionManager;
 import org.datacleaner.connection.OutputDataStreamDatastore;
+import org.datacleaner.data.MetaModelInputColumn;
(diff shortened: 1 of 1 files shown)


jhorcicka:
Tested and LGTM.

PR #1640, opened 2016-12-08
Feature/1636 hot folder

### LINKED ISSUE #1636: Hot folder (incoming)
The job should be kicked off when input (data) file arrives in the hot folder.

jhorcicka:
Resolves #1636 

Various strategies how to wait for an incomplete file (that is being copied to the hot folder, for example).


monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -131,16 +137,80 @@ private HotFolderAlterationListener(final JobIdentifier job, final ScheduleDefin
...
+                waitStrategy = new WindowsWaitForCompleteFileStrategy();
+            } else if (osName.contains("linux")) {
+                logger.info("Using LinuxWaitForCompleteFileStrategy. ");
+                waitStrategy = new LinuxWaitForCompleteFileStrategy();
+            } else {
+                logger.info("Using GeneralWaitForCompleteFileStrategy. ");
+                waitStrategy = new GeneralWaitForCompleteFileStrategy();
+            }
+
+            waitStrategy.waitForComplete(file);
```
LosD: How will the change-notification framework we're using handle the thread being blocked for a long time, especially if there are multiple changed files?

LosD: (Maybe consider using "completed" notifications instead)

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -131,16 +137,80 @@ private HotFolderAlterationListener(final JobIdentifier job, final ScheduleDefin
...
+            try {
+                final TenantContext tenantContext = _tenantContextFactory.getContext(tenant);
+                final JobContext jobContext = tenantContext.getJob(job);
+                final String jobFileName = jobContext.getJobFile().getName();
+                final RepositoryFile jobRepositoryFile = tenantContext.getTenantRootFolder().getFolder("jobs")
+                        .getFile(jobFileName);
+                final JaxbJobReader reader = new JaxbJobReader(tenantContext.getConfiguration());
+                final File jobFile = new File(jobRepositoryFile.toResource().getQualifiedPath());
+
+                return reader.readMetadata(jobFile).getDatastoreName();
```
LosD: Wow, that's a lot of code for getting the datastore name. Don't we have some sort of utility function that does this?

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/AbstractWaitForCompleteFileStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.monitor.server.filesystem;
+
+import java.io.File;
+
+public abstract class AbstractWaitForCompleteFileStrategy implements WaitForCompleteFileStrategy {
+    protected static final int ATTEMPT_WAIT_INTERVAL_MS = 500;
+    protected static final int WAIT_ATTEMPT_LIMIT = 20 * 60 * (2 * ATTEMPT_WAIT_INTERVAL_MS);
```
LosD: These should probably be configurable as well. For example the `GeneralWaitForCompleteFileStrategy` might need a little (or a lot, actually) more time than half a second, depending on the program, chunk sizes and quality of Internet connection when downloading.

We might want to specify the total limit as time instead of attempts. Attempts make sense when retrying something that failed, but a timeout probably makes more sense in this case.

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/IncompleteFileException.java
```diff
@@ -0,0 +1,28 @@
...
+ * along with this distribution; if not, write to:
+ * Free Software Foundation, Inc.
+ * 51 Franklin Street, Fifth Floor
+ * Boston, MA  02110-1301  USA
+ */
+package org.datacleaner.monitor.server.filesystem;
+
+import java.io.IOException;
+
+public class IncompleteFileException extends IOException {
```
LosD: I can't decide if I really agree that this is an `IOException`. The IO channels might be functioning perfectly, it just took longer than expected (or program was shut down). I don't really mind though, so let's leave as-is unless others also think it's a wrong thing to inherit from.

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/WindowsWaitForCompleteFileStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+
+/**
+ * Windows implementation of {@link WaitForCompleteFileStrategy}.
+ */
+public class WindowsWaitForCompleteFileStrategy extends AbstractWaitForCompleteFileStrategy {
+    /**
+     * It relies on a locking mechanism when a file being copied is locked from other processes.
```
LosD: Hmmm... Won't the lock only be blocked from _writing_ in many cases?

And couldn't you use `File.canWrite()` for that, or does that only check actual permissions?

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/WindowsWaitForCompleteFileStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+     */
+    @Override
+    public boolean isReady(final File file) {
+        try (FileInputStream inStream = new FileInputStream(file)) {
+            inStream.read();
+
+            return true;
+        } catch (final IOException ioe) {
```
LosD: Let's just use `e`, It's not like catch blocks are large enough that anyone would be in doubt about which exception it is :)

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
jhorcicka: I did not find any.

LosD: Oh well

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/LinuxWaitForCompleteFileStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+     * @return true if the file is not opened in 'write' mode, false otherwise
+     */
+    @Override
+    public boolean isReady(final File file) {
+        final Process process;
+
+        try {
+            final String command = FILE_OPENED_FOR_WRITE_COMMAND_TEMPLATE
+                    .replace(FILE_PLACEHOLDER, file.getAbsolutePath());
+            process = new ProcessBuilder("/bin/sh", "-c", command).start();
```
LosD: I think all of them are part of the [File Hierarchy Standard](http://www.linuxfoundation.org/collaborate/workgroups/lsb/fhs), so I wouldn't worry too much about it.

LosD: Oh, reading closer, it seems like FHS doesn't really care which system utilities are there. I though it also specified a minimum set of tools.

Anyway, I can't imagine a Linux system without all of them. Maybe even Unix systems (though they might behave a little different to their Linux counterparts)

LosD: You _could_ also make the template configurable, for maximum system compatibility. Might not be worth it, though.

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/WindowsWaitForCompleteFileStrategy.java
jhorcicka: File.canWrite() returns _true_ even during the copying.

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/AbstractWaitForCompleteFileStrategy.java
jhorcicka: Which kind of configuration do you want? (Where should I configure this?)

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/WindowsWaitForCompleteFileStrategy.java
jhorcicka: This was actually an error, changed to _new FileWriter(...)_ which works.

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
LosD: It's might not be a problem, but we'll need to test it. For example if many files changes at once (might sound far fetched, but I'd not be surprised if they upload a lot of campaigns at the same time)?

Anyway, blocking threads like that seems... Dirty. A single thread picking up work should be able to handle all active checks and notify the system of completed files, up to a pretty large limit (depending on the actual checks, of course).

Anyway, not a blocker, but something to consider.

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/LinuxWaitForCompleteFileStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+
+            try (InputStream inputStream = process.getInputStream();
+                    InputStreamReader inputStreamReader = new InputStreamReader(inputStream);
```
ClaudiaPHI: finals

LosD: o_O

LosD: What was the error description?

LosD: Oh, I guess it was ambiguous, like `public` in an interface method. They are `final` by definition.

jhorcicka: Yes, it was something like "unnecessary 'final' keyword" (I don't remember exact wording).

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/LinuxWaitForCompleteFileStrategy.java
jhorcicka: It is configurable now.

monitor/services/src/main/java/org/datacleaner/monitor/server/filesystem/AbstractWaitForCompleteFileStrategy.java
jhorcicka: I increased the default value to 30 minutes, but it is now configurable so "anyone" can change that.

jhorcicka: Strategy and time limit are now configurable in datacleaner-monitor.properties (CE edition only).

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
jhorcicka: See executeJobWithFileInNewThread(...).

LosD: Still not entirely what I was looking for, better though.

But at least use one of the existing thread pools.

jhorcicka:
New features
* <FILENAME>.trigger file landing in hot-folder executes the job with original datastore. 
* Input columns of the new file are checked against the original datastore (files with different "header" will not trigger anything). 
* Job variables (file name, extension, date and time).

monitor/ui/pom.xml
```diff
@@ -208,13 +208,13 @@
...
-									<goal>stop</goal>
-								</goals>
-							</execution>
+							<!--<execution>-->
+								<!--<id>stop</id>-->
+								<!--<phase>post-integration-test</phase>-->
+								<!--<goals>-->
+									<!--<goal>stop</goal>-->
+								<!--</goals>-->
+							<!--</execution>-->
```
arjansh: Please re-institute the "stop" execution, so no Docker images keep lingering around.

monitor/ui/src/main/webapp/repository/demo/datastores/simple-numbers.csv
```diff
@@ -0,0 +1,6 @@
+VALUE
+1
+2
+3
+4
+5
```
arjansh: Missing newline at end of file.

jhorcicka: Fixed.

jhorcicka:
Failing test on Windows was fixed by adding .gitattributes file containing new-line-character settings. Without that there were problems with bash scripts being copied from Windows environment to the docker container (and using them).

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.sh		eol=lf
+universalJavaApplicationStub		eol=lf
```
LosD: Whoops, found after review approval, but missing EOL at EOF

LosD:
Damn, I think I found an issue: The current time format will not work on windows files

monitor/services/src/main/java/org/datacleaner/monitor/server/SchedulingServiceImpl.java
```diff
@@ -103,9 +118,18 @@
...
 
     private final class HotFolderAlterationListener extends FileAlterationListenerAdaptor {
         private static final String PROPERTIES_FILE_EXTENSION = ".properties";
+        private static final String TRIGGER_FILE_EXTENSION = ".trigger";
+        private static final String VARIABLE_FILE_NAME = "hotfolder.input.filename";
+        private static final String VARIABLE_EXTENSION = "hotfolder.input.extension";
+        private static final String VARIABLE_DATE = "datacleaner.run.date";
+        private static final String VARIABLE_TIME = "datacleaner.run.time";
+        private static final String DATE_FORMAT = "yyyy/MM/dd";
+        private static final String TIME_FORMAT = "HH:mm:ss";
```
LosD: `DATE_FORMAT` and `TIME_FORMAT` needs to use safe characters for filenames. That means anything but ASCII's control characters, null and `\/:*?"<>|`

My suggestion is a simple `yyyyMMdd` for `DATE_FORMAT` and `HHmmss` for `TIME_FORMAT`
If you want separators, you could also use `yyyy-MM-dd` and `HH.mm.ss`, but I don't think it really matters. The most important function is making sure files are separated (so mulitple runs of the same file doesn't overwrite output or fail) and sorted properly.

.gitattributes
jhorcicka: Added.

jhorcicka:
Date&time formats are now without separators.

PR #1646, opened 2016-12-19
Align Eclipse formatter with IntelliJ

arjansh:
Changed eclipse code formatter so it formats the code more like Intellij does using the IDEA formatter.

Also added a few @formatter:off, @formatter:on annotations for some difficult cases.


LosD:
@arjansh Do you want anyone to test it?

I can install an Eclipse and compare with my IntelliJ, but I'm not sure we'd get anything out of it, as the testing would be pretty random lookups, which we already did.

arjansh:
@losd I think it would be nice if someone else using Eclipse starts using it, to double check that there are no corruptions introduced into the Formatter-Eclipse.xml. I don't think it is useful to directly apply it to all the code, because then we'll just encounter the same differences as before.

LosD:
Agreed. I guess @ClaudiaPHI is the only non-you Eclipse user.

ClaudiaPHI:
The build passes. LGTM!

PR #1652, opened 2016-12-23
Retry handler for DataHub connections

### LINKED ISSUE #1651: Retry call on DataHubConnection after timeout
It is not desirable that a job directly fails when it gets a connection timeout in a DataHubConnection, this may be caused by the DataHub connection getting flooded by calls and thus not being able to handle them all within the expected time frame. Therefore it would be nice if we would have some kind of retry mechanism when using DataHub connections.

arjansh:
Fixes #1651.

Adding a HttpRequestRetryHandler to the HttpClient used within the DataHubConnection, which does 3 retries when connecting to DataHub before failing, instead of failing directly when one call times out.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubConnection.java
+++ b/engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubConnection.java
@@ -21,10 +21,16 @@ package org.datacleaner.metamodel.datahub;
 
 import static org.datacleaner.metamodel.datahub.DataHubSecurityMode.CAS;
 
+import java.io.InterruptedIOException;
 import java.net.URISyntaxException;
+import java.net.UnknownHostException;
+import java.util.Arrays;
+
+import javax.net.ssl.SSLException;
 
 import org.apache.http.client.utils.URIBuilder;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.DefaultHttpRequestRetryHandler;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
 import org.datacleaner.util.SecurityUtils;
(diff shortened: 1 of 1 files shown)


engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubConnection.java
```diff
@@ -136,4 +143,11 @@ private URIBuilder appendToPath(final URIBuilder uriBuilder, final String pathSe
 
         return uriBuilder.setPath(pathSegment);
     }
+
+    private class DataHubRequestRetryHandler extends DefaultHttpRequestRetryHandler {
+        DataHubRequestRetryHandler() {
+            super(3, false,
+                    Arrays.asList(InterruptedIOException.class, UnknownHostException.class, SSLException.class));
```
LosD: Any advantage to creating a subclass that just calls the constructor of the super class?

I see a bit in hiding the details, but since it is only used privately, the benefits to that are pretty limited.

OTOH, it's not like it's hard to understand, so nothing lost from this... Mostly just made me curious :)

LosD: Ahhh, get it. It's used multiple times, so it was this or a createRetryHandler method (though I think you might be allowed to reuse a retry handler).

LosD: BTW, any reason to retry connect exception? That's usually connection refused. I don't think timeouts would trigger that.

arjansh: The reason I extend the DefaultHttpRequestRetryHandler using the constructor is that the standard constructors of that class adds ConnectException to the list of "nonRetriable" classes, but in this case we do want to retry in case of a ConnectException, because that's just what the timeout triggers.

LosD: Ah! That constructor is protected! Totally missed that :)

arjansh:
We need to do some more investigation on the root cause of this issue. Is it caused because we are trying to connect using a connection which has been closed by the server because it has been idle for too long, or is the connection still alive, but is the server just taking too long to return a response. If the first thing is the case we just have to make sure we reconnect or introduce some kind of connection pool mechanism for the http connections (using something like http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html?)

LosD:
There's also ways to force HTTP Client to close the connection itself after some time: https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html#d5e425

(However, if that is the issue, I really think the server is misbehaving)

arjansh:
Closing PR, because we need to do some more research on the root cause of the problem first.

PR #1658, opened 2017-01-10
Add dictionaries to TextCaseTransformer

### LINKED ISSUE #1654: Text case transformer improvements
The text case transformer should be a little more compatible with Unicode, and be better at word-break detection.

Furthermore, it should be possible to define dictionaries and patterns to affect the casing.

LosD:
Dictionaries for whole sentence, words and beginning and ends has been added. This should make it possible to serve all needs for companies and people (though not super optimal regarding people. You'll need to add e.g. "O'S", "O'L", "McD" etc. for all possible Irish/Scotch names.).

Fixes #1654


api/src/main/java/org/datacleaner/reference/DictionaryConnection.java
```diff
@@ -29,6 +31,18 @@
...
+        return StreamSupport.stream(iterable.spliterator(), false);
+    }
+
+    default Stream<String> stream() {
+        final Iterable<String> iterable = this::getLengthSortedValues;
+        return StreamSupport.stream(iterable.spliterator(), false);
+
+    }
+
+
```
LosD: Removing

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextCaseTransformer.java
```diff
@@ -19,24 +19,44 @@
...
+import com.ibm.icu.text.BreakIterator;
 
 @Named("Text case transformer")
 @Description("Modifies the text case/capitalization of Strings.")
 @Categorized(TextCategory.class)
 public class TextCaseTransformer implements Transformer {
+    private static final String DEFAULT_START_CASE_DICTIONARY = "TextCaseTransformer start case dictionary";
+    private static final String DEFAULT_END_CASE_DICTIONARY = "TextCaseTransformer end case dictionary";
+    private static final String DEFAULT_WORD_CASE_DICTIONARY = "TextCaseTransformer word case dictionary";
+    private static final String DEFAULT_COMPLETE_CASE_DICTIONARY = "TextCaseTransformer fixed case dictionary";
```
LosD: Yeah, that's a mess. Will find some better wording.

LosD:
The error comes from a bug in the old code: It specifically checks for ".!?:", but it does so on the following character, which is usually a space, instead of first character of the next word.

The new code does not have that issue, but the problem is that this is not only a change in behavior, but for our integration test, it is actually a change for the worse; e.g. the bug meant that it would output "Sr. consultant", but the new code will output "Sr. Consulatant". Since the output is an abbreviation, the old code is actually more correct. OTOH, for real actual sentences, I'm pretty sure the new code is proper.

I could introduce a new mode, like "Capitalize first character of value" that behaves like the old one?

jhorcicka:
What about some dictionary for abbreviations that would be skipped during capitalization (...words after abbreviations)?

LosD:
It's a big change, so I think we'll do that later, if needed.

PR #1660, opened 2017-01-10
Functionality for invoking analyzer jobs on DC monitor

### LINKED ISSUE #1661: JobInvocationController should allow invoking analyzer jobs
Currently, the methods for invoking jobs on the JobInvocationController only execute the transformer from the job provided. For new functionality in DataHub we need it to be able to execute analyzers as well.

MennoB:
Fixes #1661

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/configuration/PlaceholderAnalysisJobWithAnalyzers.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/configuration/PlaceholderAnalysisJobWithAnalyzers.java
@@ -29,7 +29,7 @@ public class PlaceholderAnalysisJobWithAnalyzers extends PlaceholderAnalysisJob 
 
     private final List<AnalyzerJob> analyzerJobs;
 
-    public PlaceholderAnalysisJobWithAnalyzers(Datastore datastore, AnalysisJob delegateJob) {
+    public PlaceholderAnalysisJobWithAnalyzers(final Datastore datastore, final AnalysisJob delegateJob) {
         super(datastore, delegateJob);
         this.analyzerJobs = delegateJob.getAnalyzerJobs();
 
--- a/monitor/services/src/main/java/org/datacleaner/monitor/configuration/PlaceholderAnalysisJobWithAnalyzers.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/configuration/PlaceholderAnalysisJobWithAnalyzers.java
@@ -25,6 +25,10 @@ import org.datacleaner.connection.Datastore;
 import org.datacleaner.job.AnalysisJob;
 import org.datacleaner.job.AnalyzerJob;
 
+/**
+ * {@link AnalysisJob} which replaces the {@link Datastore} of another
+ * {@link AnalysisJob}.
+ */
 public class PlaceholderAnalysisJobWithAnalyzers extends PlaceholderAnalysisJob {
 
     private final List<AnalyzerJob> analyzerJobs;
@@ -35,6 +39,11 @@ public class PlaceholderAnalysisJobWithAnalyzers extends PlaceholderAnalysisJob 
 
     }
 
+    /**
+     * Gets all {@link AnalyzerJob}s contained in this job.
+     *
+     * @return
+     */
     @Override
     public List<AnalyzerJob> getAnalyzerJobs() {
         return analyzerJobs;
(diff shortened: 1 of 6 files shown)


MennoB:
#1661

LosD:
For the automatic issue closing-on-merge, you need to write it as "Fixes #1661", and I think it needs to go into the description, not a comment.

MennoB:
@LosD Ah, thanks!

monitor/services/src/main/java/org/datacleaner/monitor/configuration/PlaceholderAnalysisJobWithAnalyzers.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.datacleaner.connection.Datastore;
+import org.datacleaner.job.AnalysisJob;
+import org.datacleaner.job.AnalyzerJob;
+
+public class PlaceholderAnalysisJobWithAnalyzers extends PlaceholderAnalysisJob {
+
+    private final List<AnalyzerJob> analyzerJobs;
+
+    public PlaceholderAnalysisJobWithAnalyzers(Datastore datastore, AnalysisJob delegateJob) {
```
ClaudiaPHI: all parameters should be final(the team has introduced new checkstyle rules)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobInvocationController.java
```diff
@@ -176,17 +149,139 @@ public JobInvocationPayload invokeJob(@PathVariable("tenant") final String tenan
...
+     * @param input
+     * @throws Throwable
+     */
+    @RequestMapping(value = "/{tenant}/jobs/{job:.+}.invoke.complete/mapped", method = RequestMethod.POST, 
+            produces = "application/json", consumes = "application/json")
+    @ResponseBody
+    @RolesAllowed(SecurityRoles.TASK_ATOMIC_EXECUTOR)
+    public void invokeJobWithAnalyzersMapped(@PathVariable("tenant") final String tenant,
+            @PathVariable("job") String jobName, @RequestBody final JobInvocationPayload input) throws Throwable {
+        logger.info("Request payload: {}", input);
```
ClaudiaPHI: final for jobMane

monitor/services/src/main/java/org/datacleaner/monitor/configuration/PlaceholderAnalysisJobWithAnalyzers.java
MennoB: Done.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/JobInvocationController.java
MennoB: Done.

JoosjeBoon:
Hi guys, can we pull this one if everyone is happy? Thanks!

LosD:
Sorry, I'm going to have to ask for more description and JavaDoc. I see new endpoints, but I don't see a description on why they are there, neither in the JavaDoc nor in the pull request. 

Quickly looking at the code, the only difference seems to be the addition of analyzers. Wouldn't it be better to have an option to enable or disable them in the existing endpoints? That would also reduce the code duplication.

HettyK:
Hi Dennis,
We like the seperate endpoint better because it is more clear than a flag. Is that ok with you. There is hardly any code duplication because shared code was moved to a seperate function. I will add some javadoc to the
PlaceholderAnalysisJobWithAnalyzers will that do then?

LosD:
To me, a new resource doing the same thing but different is a rather weird design, but then of course we're already being weird with the /mapped endpoint, so I guess it's fine. 

But then at least add the differences between the endpoints to the JavaDoc. Right now they are identical, except for the missing return (any reason you decided to pull that? If you're interested in the output _and_ want to write, you now need to run the job twice)

LosD:
@HettyK The PR will probably fail now. For some reason you reformatted the code with some weird formatting rules. I'm not sure, but I think CheckStyle will barf on those.

HettyK:
Ai, I will take a look....

LosD:
Regarding the actual text, I was more thinking something in the line of "this will run like <the other endpoint>, but is meant for writing data, so will also run analyzers, but will not return transformed values". Just so anyone looking has in idea what's going on.

(Sorry about keeping you working on this)

LosD:
Okay, I'm happy. Let's see if Travis also is, then merge :)

HettyK:
ok thanks! :-)

LosD:
Travis happy, merging!

PR #1670, opened 2017-01-23
Monitor rest interface for reference data

### LINKED ISSUE #1668: REST interface for reference data resources
There's still not REST interface for the data resources. This is needed for another project.

Please note that this will be pretty inconsistent with our other web APIs, but they're already inconsistent from each other, and I'm hoping that we can slowly change them to standardize around a more proper RESTful format.

LosD:
This creates a set of endpoints for manipulating reference data.

The interface is not entirely complete, e.g. you can only add discrete values and file-based dictionaries/synonym catalogs, and you can only download discrete values, but it should be enough to support the intended project.

To simplify the controller, `ReferenceDataDao`(and `ReferenceDataDaoImpl`) has been heavily changed to better take advantage of `JaxbJobWriter`.

Fixes #1668 

#### Reviewing notes
The interface can be gleamed from the test cases, except for the file upload (still missing, will add to branch shortly).

The GWT reference data page should probably be tested to make sure that my changes hasn't broken it. I did a cursory test, and things seemed to work as intended.


LosD:
Errr.. @ClaudiaPHI, you assigned yourself to this yesterday, but I assume you're not still reviewing it?

monitor/services/src/main/java/org/datacleaner/monitor/server/dao/ReferenceDataDao.java
```diff
@@ -28,49 +28,71 @@
...
      *
      * @param tenantContext
-     * @param stringPattern
+     * @param synonymCatalog
+     */
+    void addStringPattern(TenantContext tenantContext, StringPattern synonymCatalog);
+
+    /**
+     * Removes a dictionary from a tenant's configuration.
+     *  @param tenantContext
```
arjansh: Indentation is off, one white space too many

monitor/services/src/main/java/org/datacleaner/monitor/wizard/referencedata/AbstractReferenceDataWizardSession.java
```diff
@@ -55,23 +55,17 @@ public Integer getPageCount() {
 
     @Override
     public String finished() {
-        final DocumentBuilder documentBuilder = XmlUtils.createDocumentBuilder();
-        final TenantContext tenantContext = _wizardContext.getTenantContext();
-        final Element updatedReferenceDataSubSection = getUpdatedReferenceDataSubSection(documentBuilder);
-        final ReferenceDataDao referenceDataDao = getReferenceDataDao();
```
arjansh: DocumentBuilder, TentantContext, XmlUtils and Element are no longer used in class, so the imports can also be removed.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,383 @@
...
-            @RequestParam("file") final MultipartFile file) {
-        String status = "success";
-
-        try {
-            processUploadedFile(tenant, file.getInputStream());
-        } catch (final Exception e) {
-            status = "failure";
+    public ResponseEntity putDictionaryFile(@PathVariable("tenant") final String tenant,
+            @PathVariable("name") final String name, @RequestParam("file") final MultipartFile sourceMultipartFile,
+            @RequestParam(value="casesensitive", defaultValue="true") final boolean isCaseSensitive,
```
arjansh: WhitespaceAround: '=' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,383 @@
...
+            @RequestParam(value="encoding", defaultValue = "UTF-8") final String encoding) throws IOException {
```
arjansh: Same here.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,383 @@
...
+        _referenceDataDao.addSynonymCatalog(context, catalog);
+        return ResponseEntity.created(ServletUriComponentsBuilder.fromCurrentRequestUri().build().toUri()).build();
+    }
+
+    @RolesAllowed(SecurityRoles.CONFIGURATION_EDITOR)
+    @RequestMapping(value = "/synonymCatalog/{name}", method = { RequestMethod.PUT },
+            consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
+    public ResponseEntity putSynonymCatalogFile(@PathVariable("tenant") final String tenant,
+            @PathVariable("name") final String name, @RequestParam("file") final MultipartFile sourceMultipartFile,
+            @RequestParam(value="casesensitive", defaultValue="true") final boolean isCaseSensitive,
```
arjansh: WhitespaceAround: '=' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,383 @@
...
+            @RequestParam(value="encoding", defaultValue = "UTF-8") final String encoding) throws IOException {
```
arjansh: Same here.

monitor/services/src/test/java/org/datacleaner/monitor/server/controllers/ReferenceDataControllerTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Test
+    public void testPutDictionaryFile() throws Exception {
+        final File file = new File("src/test/resources/testdictionary");
+        final MockMultipartFile multipartFile =
+                new MockMultipartFile("file", "testDictionary", "text/plain", new FileInputStream(file));
+        final MvcResult mvcResult = _mockMvc.perform(
+                fileUpload("/test/referencedata/dictionary/testDictionary").file(multipartFile).with(withPut())
+                        .contentType(MediaType.MULTIPART_FORM_DATA).param("casesensitive", "false")).andExpect(status().isCreated()).andExpect(
```
arjansh: Line is longer than 130 characters (found 143).

monitor/services/src/test/java/org/datacleaner/monitor/server/controllers/ReferenceDataControllerTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Test
+    public void testPutSynonymCatalogFile() throws Exception {
+        final File file = new File("src/test/resources/testsynonymcatalog");
+        final MockMultipartFile multipartFile =
+                new MockMultipartFile("file", "testsynonymcatalog", "text/plain", new FileInputStream(file));
+        final MvcResult mvcResult = _mockMvc.perform(
+                fileUpload("/test/referencedata/synonymCatalog/testCatalog").file(multipartFile).with(withPut())
+                        .contentType(MediaType.MULTIPART_FORM_DATA).param("casesensitive", "false")).andExpect(status().isCreated()).andExpect(
```
arjansh: Line is longer than 130 characters (found 143).

monitor/services/src/test/resources/testsynonymcatalog
```diff
@@ -0,0 +1 @@
+foo,bar,baz
```
arjansh: No newline at end of file.

engine/xml-config/src/main/java/org/datacleaner/configuration/JaxbConfigurationReader.java
```diff
@@ -683,6 +684,24 @@ private ReferenceDataCatalog createReferenceDataCatalog(final ReferenceDataCatal
                         synonymCatalogList.add(sc);
 
                         removeVariablePath();
+                    } else if (synonymCatalogType instanceof ValueListSynonymCatalogType) {
+                        final ValueListSynonymCatalogType vlsct = (ValueListSynonymCatalogType) synonymCatalogType;
```
LosD: I believe I'm just following the existing code style in this file. I agree it's overly terse, though.

engine/xml-config/src/main/java/org/datacleaner/configuration/DomConfigurationWriter.java
```diff
@@ -527,6 +531,24 @@ private Element toElement(final TextFileSynonymCatalog sc) {
 
         return elem;
     }
+    
+    private Element toElement(final SimpleSynonymCatalog sc) {
```
LosD: Same here. Just trying to blend in :)

monitor/pom.xml
```diff
@@ -64,6 +64,12 @@
 			<version>${project.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.jayway.jsonpath</groupId>
+			<artifactId>json-path</artifactId>
+			<version>2.1.0</version>
```
LosD: Whoopsie. Right.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/referencedata/model/DictionaryBasedModel.java
```diff
@@ -0,0 +1,51 @@
...
+package org.datacleaner.monitor.server.controllers.referencedata.model;
+
+import java.util.Collection;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class DictionaryBasedModel<T> {
+    protected final boolean _caseSensitive;
```
LosD: I think it's because it was initially pulled up from subclasses, then I probably forgot to tighten visibility when I removed the references.

monitor/extensions/referencedata-wizards/src/main/java/org/datacleaner/monitor/server/wizard/shared/file/FileWizardSession.java
```diff
@@ -52,14 +52,6 @@ public Integer getPageCount() {
...
     }
 
-    @Override
-    protected Element getUpdatedReferenceDataSubSection(final DocumentBuilder documentBuilder) {
-        copyUploadedFileToReferenceDataDirectory();
-        return addElementToConfiguration();
-    }
-
-    protected abstract Element addElementToConfiguration();
-
```
LosD: Hmmm... That removal might actually have been a mistake. I'll have to look into that.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,383 @@
...
+
+    @RolesAllowed(SecurityRoles.VIEWER)
+    @RequestMapping(value = "/stringPattern/{name}", method = { RequestMethod.GET, RequestMethod.HEAD },
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    public StringPatternModel getStringPattern(@PathVariable("tenant") final String tenant,
+            @PathVariable("name") final String name, final HttpServletRequest request) {
+        final StringPattern stringPattern = getReferenceDataCatalog(tenant).getStringPattern(name);
+
+        if (stringPattern == null) {
+            logger.warn("Could not find string pattern \"" + name + "\"");
```
LosD: Did I do this? Yuck.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,361 @@
...
+            stringPattern = new SimpleStringPattern(name, stringPatternModel.getPattern());
+            break;
+        case REGEX:
+            stringPattern = new RegexStringPattern(name, stringPatternModel.getPattern(), false);
+            break;
+        case REGEX_MATCH_ENTIRE_STRING:
+            stringPattern = new RegexStringPattern(name, stringPatternModel.getPattern(), true);
+            break;
+        default:
+            throw new IllegalArgumentException("No such string pattern type");
```
LosD: Not really a problem, it's an enum so it should never happen, mostly here because Java doesn't realize that all branches are already covered, making it complain about possibly uninitialized stringPattern below. OTOH, it's not really a tough task :)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,361 @@
...
+        final TenantContext context = _contextFactory.getContext(tenant);
+        _referenceDataDao.removeStringPattern(context, name);
+        _referenceDataDao.addStringPattern(context, stringPattern);
+
+        return ResponseEntity.created(ServletUriComponentsBuilder.fromCurrentRequestUri().build().toUri()).build();
+    }
+
+    @RolesAllowed(SecurityRoles.CONFIGURATION_EDITOR)
+    @RequestMapping(value = "/stringPattern/{name}", method = { RequestMethod.DELETE })
+    public ResponseEntity deleteStringPattern(@PathVariable("tenant") final String tenant,
```
LosD: It's never called compile-time, so the generics doesn't really apply. Of course it still serves some documentation purpose, so I've added it (it's always `ResponseEntity<Void>` in our case).

monitor/extensions/referencedata-wizards/src/main/java/org/datacleaner/monitor/server/wizard/shared/file/FileWizardSession.java
LosD: I called it from the subclasses instead. Not exactly perfect design, though (and the wizard definitely needs to be tested thoroughly).

arjansh:
When I create "Reference data" through one of the wizards in the Monitor, I can give it a name which may for example contains white spaces, like the "vendor whitelist" for example. I can then not interact with such reference data entities through the REST interface, because I cannot use its name (which contains space) in a URL. We have to figure out some kind of manner to make this work.

arjansh:
Each time I try to execute one of the PUT methods without the necessary authorization (e.g. with "viewer" credentials instead of "admin" credentials), I get a nice big stacktrace in the servlet container's logfile. 

I think this should either not be logged at all, because the user is presented with a nice "HTTP Status 403 - Access is denied", or just as a one liner.

LosD:
Use '+' or %20

LosD:
For the stack trace, that's a problem which should probably be handled in another issue. At least I didn't do anything special here, so I expect it is the same in for other endpoints as well.

arjansh:
Regarding the stack trace, I understand.

Regarding the white space. The problem is, that when I just invoke the "referencedata" endpoint I get a json string containing references to all reference data, but there the href's value contains a whitespace:

`http://localhost:8080/DataCleaner-monitor/repository/demo/referencedata/dictionary/test test`

This endpoint should the return a reference which contains the %20 instead of the white space in the URL. (and the same should go for the referencedata/dictionaries, referencedata/stringPatterns and referencedata/synonymCatalogs endpoints).

LosD:
True about the +

I'll see if I can update the given URL, but it shouldn't be necessary, all clients works with space automatocally, at least all mine (Postman, Chorme, Firefox, Internet Explorer, Edge).

LosD:
Okay, it seems at least curl has a problem there. Will fix

LosD:
Fixed! :)

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ReferenceDataController.java
```diff
@@ -19,59 +19,384 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.multipart.MultipartFile;
```
arjansh: This import is no longer used after the last commit.

monitor/services/src/test/java/org/datacleaner/monitor/server/controllers/ReferenceDataControllerTest.java
```diff
@@ -0,0 +1,217 @@
...
+                .andExpect(status().isCreated()).andExpect(header().string("Location",
+                        is("http://localhost/test/referencedata/dictionary/testDictionary"))).andReturn();
+
+        final String location = mvcResult.getResponse().getHeader("Location");
+        _mockMvc.perform(get(location)).andExpect(status().isOk()).andExpect(jsonPath("$.name", is("testDictionary")))
+                .andExpect(jsonPath("$.entries", hasSize(3))).andExpect(jsonPath("$.entries[2]", is("baz")))
+                .andExpect(jsonPath("$.caseSensitive", is(false)));
+    }
+
+    private RequestPostProcessor withPut() {
```
arjansh: This method is no longer used after the last commit.

monitor/services/src/test/java/org/datacleaner/monitor/server/controllers/ReferenceDataControllerTest.java
```diff
@@ -0,0 +1,217 @@
...
+        _mockMvc.perform(get("/test/referencedata/synonymCatalogs")).andExpect(status().isOk())
+                .andExpect(content().contentType(MediaType.APPLICATION_JSON)).andExpect(jsonPath("$", hasSize(1)))
+                .andExpect(jsonPath("$[?(@.referenceDataType!='SYNONYM_CATALOG')]", hasSize(0)))
+                .andExpect(jsonPath("$[?(@.referenceDataType=='SYNONYM_CATALOG')]", hasSize(1))).andExpect(
+                jsonPath("$[0].href", is("http://localhost/test/referencedata/synonymCatalog/job%20title%20synonyms")));
+    }
+
+    @Test
+    public void testPutAndDeleteSynonymCatalog() throws Exception {
+        final SynonymCatalogModel dictionaryModel = new SynonymCatalogModel("synonymCatalog",
```
arjansh: The variable is called `dictionaryModel`, but it contains a `SynonymCatalogModel` so preferably rename it to `synonymCatalogModel`.

PR #1672, opened 2017-01-24
Add validation.

### LINKED ISSUE #1671: Missing validation for dictionaries
The improved text case transformed (#1654) does not validate its dictionaries. It should not accept non-case sensitive dictionaries.

LosD:
Also removes the default dictionaries, which wont work unless we start doing constructor injection.

Fixes #1671

(I'm not sure what that commit message is about. Apparently my brain auto-corrected validation to serialization... Please change as doing squash merge)

### DIFF
--- a/components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextCaseTransformer.java
+++ b/components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextCaseTransformer.java
@@ -40,6 +40,7 @@ import org.datacleaner.api.InputRow;
 import org.datacleaner.api.OutputColumns;
 import org.datacleaner.api.Provided;
 import org.datacleaner.api.Transformer;
+import org.datacleaner.api.Validate;
 import org.datacleaner.components.categories.TextCategory;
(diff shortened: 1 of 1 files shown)


components/basic-transformers/src/main/java/org/datacleaner/beans/transform/TextCaseTransformer.java
```diff
@@ -153,6 +124,20 @@ public void init() {
         wordEndDictionaryConnections = openConnections(wordEndDictionaries);
     }
 
+    @Validate
+    public void validate() {
+        validateDictionaries(allWordsDictionaries);
+        validateDictionaries(wordEndDictionaries);
+        validateDictionaries(wordEndDictionaries);
```
ClaudiaPHI: Looks like a copy paste issue here. 3 times "wordEndDictionaries"

LosD: Oh damn. Right you are!

ClaudiaPHI:
Tested. Looks good to me!

PR #1676, opened 2017-02-02
Docker tomcat port change

jhorcicka:
(No issue) 

Our Jenkins environment already uses port 8080, let's change it to something more original.


jhorcicka:
Hmm, okay. But then I am really wondering why the build was successful after the port has been changed.

PR #1678, opened 2017-02-03
REST endpoint to schedule job on Monitor with Hot Folder

### LINKED ISSUE #1677: Provide REST endpoint to schedule job on Monitor with Hot Folder
We want to be able to define a schedule where a job is trigger through a hot folder dependencies through a REST endpoint.

arjansh:
Resolves #1677 

Added an option to the JobsFolderController, so you can directly define a hot folder for a job which is posted to it. Also added a scheduling endpoint which you can use to schedule an existing job so it uses a hot folder.

I didn't add a DELETE option to the new REST endpoint, because deleting a schedule implicitly deletes the scheduled job, which can already be done through a "jobs" REST endpoint.

Next to that I updated the dependency on mockito to the latest version to be able to make use of java 8 lambda expressions in the unit tests.


jhorcicka:
Tested and LGTM.

PR #1680, opened 2017-02-03
#1679 Job execution progress always visible

### LINKED ISSUE #1679: Job execution log output and hidden progress
Job execution log output (on job execution progress page), when long enough, can make the blue progress bar disappear. The view is continuously updated to the last line so that the progress line goes up -- out of the visible screen. Let's add a scroll bar to the execution log text area.

jhorcicka:
Resolves #1679 

Job execution page: the progress bar is not any more moved from visible screen by long log output.

Before:
before

After:
after

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/util/WidgetFactory.java
+++ b/desktop/api/src/main/java/org/datacleaner/util/WidgetFactory.java
@@ -63,6 +63,7 @@ import org.jdesktop.swingx.JXStatusBar;
 import org.jdesktop.swingx.JXTaskPane;
 import org.jdesktop.swingx.JXTextArea;
 import org.jdesktop.swingx.JXTextField;
+import org.jdesktop.swingx.JXTitledPanel;
 import org.jdesktop.swingx.plaf.basic.BasicStatusBarUI;
 import org.jdesktop.swingx.plaf.metal.MetalStatusBarUI;
 import org.jdesktop.swingx.prompt.PromptSupport.FocusBehavior;
@@ -348,6 +349,17 @@ public final class WidgetFactory {
             taskPane.setIcon(icon);
         }
         return taskPane;
+    }
+
+    public static JXTitledPanel createTitledPanel(final String title, final JComponent content) {
+        final JXTitledPanel titlePanel = new JXTitledPanel(title, content);
+        titlePanel.setFocusable(false);
+        titlePanel.setTitlePainter((graphics2D, panel, width, height) -> {
+            graphics2D.setColor(WidgetUtils.BG_COLOR_LESS_DARK);
+            graphics2D.fillRect(0, 0, width, height);
+        });
+
+        return titlePanel;
     }
 
     public static JXTextField createTextField() {
--- a/desktop/ui/src/main/java/org/datacleaner/panels/result/ProgressInformationPanel.java
+++ b/desktop/ui/src/main/java/org/datacleaner/panels/result/ProgressInformationPanel.java
@@ -20,7 +20,6 @@
 package org.datacleaner.panels.result;
 
 import java.awt.BorderLayout;
-import java.awt.Component;
 import java.awt.Dimension;
 import java.io.PrintWriter;
 import java.io.StringWriter;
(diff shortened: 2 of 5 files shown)


desktop/ui/src/main/java/org/datacleaner/panels/result/ProgressInformationPanel.java
```diff
@@ -96,6 +100,17 @@ public ProgressInformationPanel(final boolean running) {
...
+        final JScrollPane scrollBar = new JScrollPane(component);
+        scrollBar.setVerticalScrollBarPolicy(ScrollPaneConstants.VERTICAL_SCROLLBAR_ALWAYS);
+        final Dimension size = new Dimension(800, 600);
+        scrollBar.setMinimumSize(size);
+        scrollBar.setPreferredSize(size);
+        scrollBar.setMaximumSize(size);
+
+        return scrollBar;
+    }
+
```
ClaudiaPHI: Isn't there a utility method in WidgetUtils for scroolbars ?

jhorcicka: Yes, there is. Thanks for the tip.

LosD:
I agree with @arjansh, I can't really see any way fixed-size can work well.

Isn't it possible to "just" stack the panels with the logger autoexpanding to 100% of the remaining height?

jhorcicka:
Maybe it is, I just don't know how, yet. I also don't want to leave it with fixed-width (if possible) so I am going to give it more time.

jhorcicka:
I found a hack/workaround that looks a bit like what we want. Notice missing "Execution Log" header that used to be another clickable/collapsable JXTaskPane. I had a problem with proper spaces and hiding functionality. I tried `executionLogTaskPane.setBorder(BorderFactory.createMatteBorder(0, 0, 10, 0, (Color) null));` which cut the upper part (header containing the switch button). 

What do you say, @LosD, @arjansh?

execution-log-scrollbar

LosD:
Hmmm... It's not optimal, but might be better than the alternatives. How well does it cope with multiple progress bars (e.g. from output data streams)?

jhorcicka:
Today I tried to use JXTitledPanel instead of JXTaskPane. I think it may be better if I find out how to set look&feel. 

titled-panel

jhorcicka:
I think I have. 

execution-log

desktop/api/src/main/java/org/datacleaner/util/WidgetFactory.java
```diff
@@ -350,6 +351,17 @@ public static JXTaskPane createTaskPane(final String title, final Icon icon) {
         return taskPane;
     }
 
+    public static JXTitledPanel createTitledPanel(final String title, final JComponent content) {
+        final JXTitledPanel titlePanel = new JXTitledPanel(title, content);
+        titlePanel.setFocusable(false);
+        titlePanel.setTitlePainter((graphics2D, panel, width, height) -> {
```
LosD: Can we move the titlePainter configuration into the LookAndFeelManager, so it's a little better separated, or does the UIManager also work terribly for "JXTitledPanel.titlePainer"?

Otherwise it's no biggie, we're doing plenty of horrible things in here anyway, but it would be nice not to do even more. :)

desktop/ui/src/main/java/org/datacleaner/panels/result/ProgressInformationPanel.java
```diff
@@ -84,16 +86,20 @@ public ProgressInformationPanel(final boolean running) {
...
         progressTaskPane.add(_progressBarPanel);
 
-        final JXTaskPane executionLogTaskPane = WidgetFactory.createTaskPane("Execution log", IconUtils.ACTION_LOG);
-        executionLogTaskPane.add(_executionLogTextArea);
-
+        final JXTitledPanel executionLogPanel =
+                WidgetFactory.createTitledPanel("Execution log", WidgetUtils.scrolleable(_executionLogTextArea));
+        executionLogPanel.setBorder(new MatteBorder(1, 1, 1, 1, WidgetUtils.COLOR_ALTERNATIVE_BACKGROUND));
         final DCTaskPaneContainer taskPaneContainer = WidgetFactory.createTaskPaneContainer();
+        final int margin = 10;
```
LosD: It would probably be better to use a static final field for this.

LosD:
Works beautifully!

jhorcicka:
I made the margin a constant, but specification of the painter through LookAndFeelManager does not work.

LosD:
Okay, let's merge when Travis is happy then! :)

PR #1685, opened 2017-02-16
#1684 PlainSearchReplaceTransformer case-insensitive mode

### LINKED ISSUE #1684: Ignore case sensitivity in PlainSearch&Replace
PlainSearchReplaceTransformer is always case-sensitive. In some cases it would be a good idea to have case-INsensitive variant. 

Let's add a checkbox switching this mode of behaviour.

jhorcicka:
Resolves #1684 

PlainSearchReplaceTransformer can work in case-insensitive mode.

case-insensitive

### DIFF
--- a/components/basic-transformers/src/main/java/org/datacleaner/beans/transform/PlainSearchReplaceTransformer.java
+++ b/components/basic-transformers/src/main/java/org/datacleaner/beans/transform/PlainSearchReplaceTransformer.java
@@ -22,6 +22,7 @@ package org.datacleaner.beans.transform;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.regex.Pattern;
 
 import javax.inject.Named;
 
@@ -65,6 +66,10 @@ public class PlainSearchReplaceTransformer implements Transformer {
     @Configured(order = 3)
     @Description("Replace the entire string when the search string is found.")
     boolean replaceEntireString = false;
+
+    @Configured(order = 4)
+    @Description("Case sensitivity will be ignored during searching and replacing. ")
+    boolean ignoreCaseSensitivity = false;
 
     public static void processRemovedProperties(final ComponentBuilder builder, final StringConverter stringConverter,
             final ComponentDescriptor<?> descriptor, final Map<String, String> removedProperties) {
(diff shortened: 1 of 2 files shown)


components/basic-transformers/src/main/java/org/datacleaner/beans/transform/PlainSearchReplaceTransformer.java
```diff
@@ -66,6 +67,10 @@
     @Description("Replace the entire string when the search string is found.")
     boolean replaceEntireString = false;
 
+    @Configured(order = 4)
+    @Description("Case sensitivity will be ignored during searching and replacing. ")
+    boolean ignoreCaseSensitivity = false;
```
LosD: I'm not really sure if this is consistent with the rest of our settings, but I don't really like negative preferences.

It's not-not case sensitive... So it's case sensitive :)

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/PlainSearchReplaceTransformer.java
```diff
@@ -114,28 +119,42 @@ public OutputColumns getOutputColumns() {
...
+        return result;
+    }
+
+    private String replaceEntireString(final String value) {
+        for (final Entry<String, String> entry : replacements.entrySet()) {
+            final String replacementString = entry.getValue();
+            final String searchString = entry.getKey();
+
+            if (ignoreCaseSensitivity == true) {
+                if (value.toLowerCase().indexOf(searchString.toLowerCase()) != -1) {
```
LosD: Can be replaced with `contains()`

components/basic-transformers/src/main/java/org/datacleaner/beans/transform/PlainSearchReplaceTransformer.java
```diff
@@ -114,28 +119,42 @@ public OutputColumns getOutputColumns() {
...
+                    return replacementString;
                 }
+            } else if (value.indexOf(searchString) != -1) {
```
LosD: Same here

jhorcicka:
@LosD: updated.

LosD:
LGTM, and test went fine. Let's merge when Travis is happy

PR #1687, opened 2017-02-21
fixes open jobs as template with outputstreams

### LINKED ISSUE #1686: Template loading of jobs with datastreams broken
## Description
If a user tries to template-load a job with connected output datastreams, it will fail, unless the column names are _exactly_ the same.

## Expected Behavior
Job should be loaded correctly after mapping the columns

## Observed Behavior
Column mapping exception. Job does not load

## Replication Steps
1. Create a job with some components mapped on an output data stream
1. Try to load that job with another datastore, containing columns with other names.

## Hypothesis
It seems that when we template-load the paths in the job, we will change the input columns names in the out job, but not the input column names in the output data stream job.

ClaudiaPHI:
fixes #1686. fixes open jobs as template with outputstreams

### DIFF
--- a/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
+++ b/engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
@@ -35,6 +35,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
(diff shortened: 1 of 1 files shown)


engine/xml-config/src/main/java/org/datacleaner/job/JaxbJobReader.java
```diff
@@ -533,13 +534,22 @@ private void wireOutputDataStreams(final Map<ComponentType, ComponentBuilder> co
...
+                    final List<MetaModelInputColumn> findSourceColumn = sourceColumns.stream().filter(
+                            sourceCol -> sourceCol.getName().equals(sourceColumnPath.getPath())).collect(Collectors
+                                    .toList());
+                    final String outputStreamColumnPathName;
+                    if (findSourceColumn.size() > 0) {
+                        outputStreamColumnPathName = getOutputStreamColumnPath(findSourceColumn.get(0).getName(),
+                                componentType, componentBuilder, i);
+                    } else {
+                        outputStreamColumnPathName = getOutputStreamColumnPath(sourceColumns.get(i).getName(),
+                                componentType, componentBuilder, i);
```
LosD: This seems to work in testing, but it's unclear to me... If you find the column from `sourceColumns` by name, you take that one, otherwise you take the same from sourceColumns, but by index instead. But if the index always matches, why search for the name?

ClaudiaPHI: Some of the tests will fails, because the order of the source columns and the columns type is not the same. In the end we cannot guarantee that the order will be the same.

LosD: Okay, but then we can't trust this at all?

LosD: Just having it less probable by using the name as the primary isn't going to cut it. Then we're setting ourselves up for hard-to-find bugs.

LosD: I think I've actually seen it happening: I was using the job we played with yesterday (country standardizer->completeness->csv for 3 columns) to go forward and back from orderdb to customers.csv a few times, and all of a sudden columns were missing.

There _is_ a small chance that it was human error, though. When using a CSV file, _all_ columns will be added to the job, so they need to be removed again to make the mapping easier. I don't think so, though, since they're still in the mapping.

LosD: Hmmm... OTOH, it's also missing from the completeness _inputs_ and I don't it this part of the code should affect that at all. Probably human error, or something else.

LosD:
Code LGTM now, but Travis is angry

ClaudiaPHI:
@LosD Can I merge this ?

LosD:
No, cause I did! ;)

PR #1691, opened 2017-02-24
Bug/1689 grouper counts

### LINKED ISSUE #1689: Grouper misses counts
Description:
Around 1-3 times per 10000 rows, it will miss a count

Expected Behavior:
All rows in all groups should be counted

Observed Behavior:
Missed counts.

Replication Steps:
1. Create a job with a lot of columns.
2. Run it.
3. Check that group counts are correct

Hypothesis:
Seems to be a race condition when run multithreaded.

jhorcicka:
Fixes #1689 

Synchronized section in GrouperTransformer::run was extended and the lock object was changed to property (instead of local variable). 

There is a new GrouperIT test to check the behaviour. Its input data file contains 40000 lines which takes around 5 seconds to run the test. With less (even 30000) lines the test using code with no synchronized section still succeeds in 10% of runs (1/10).


jhorcicka:
I had some difficulties with "datacleaner-home" and "conf.xml" in test resources (there already is "datacleaner-home" and "conf.xml" in src resources). So I did some copying/removing and adding of stuff. In src it should be the same as it was before (only one datastore was added) but github shows many new files.

desktop/ui/src/main/resources/test/conf.xml
```diff
@@ -0,0 +1,62 @@
...
+			environment.
+			This includes example datastores and example reference
+			data.
+		</configuration-description>
+		<configuration-version>4.0</configuration-version>
+		<author>DataCleaner.org</author>
+	</configuration-metadata>
+
+	<datastore-catalog>
+		<csv-datastore name="grouper-job"
```
jhorcicka: The only real change in this file, added "grouper-job" CSV.

desktop/ui/src/test/java/org/datacleaner/cli/GrouperIT.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.commons.io.Charsets;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.StringUtils;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import com.google.common.io.Files;
+
+public class GrouperIT {
```
jhorcicka: This is the new "integration" test.

desktop/ui/src/test/resources/datacleaner-home-test/conf.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jhorcicka: New conf.xml for test.

desktop/ui/src/test/resources/datacleaner-home-test/expected-results/grouper-job.txt
```diff
@@ -0,0 +1,9 @@
+﻿INFO  10:46:37 DataCleanerHome - Initializing DATACLEANER_HOME
```
jhorcicka: Expected result file.

desktop/ui/src/test/resources/datacleaner-home-test/jobs/grouper-job.analysis.xml
```diff
@@ -0,0 +1,109 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
jhorcicka: Grouper test job.

jhorcicka:
@LosD: I added a comparison against expected value(s). But what do you mean by "only one group"? Is it that input data contains only one value for the grouped column?

LosD:
@jhorcicka Exactly. It should end up with multiple groups.

jhorcicka:
@LosD: Updated. 

Thanks to "expected count comparison" I could reduce the number of input data rows. See, there are two types of problems. The first one is that group-size number provided by GrouperTransformer does not match with the length of the comma-separated list of values. For this test I needed at least 40,000 lines to break the test with old code. The second problem was that some items were missed completely; meaning that they were not even in the comma-separated list. For this test current 20,000 lines is enough.

LosD:
What's the current execution time? If it's negligible, leave it at the high count.

LosD:
Oh, you already did it. That's also fine

LosD:
Travis is angry, though

LosD:
Travis error _seems_ unrelated. Trying a restart

LosD:
Travis happy. Merging! :)

PR #1695, opened 2017-03-02
Execute hot folder jobs without an existing data store

### LINKED ISSUE #1690: Make hot folder jobs executable without existing datastore
Hot folder jobs are run as template jobs by definition, so they do not really require any existing datastore.

However, the current Job engines very much requires that, which means that hot folder jobs needs to a dummy datastore definition, which can only be overridden on filename, so hot folder jobs are limited to the datastore type they were registeret with.

Let's do something about that.

LosD:
This makes hot folder jobs to execute, even if there is no existing data store.

It also remove the pre-flight testing of hot folder jobs to make sure that errors shows up in the execution log of jobs.

Fixes #1690

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/job/DataCleanerJobContext.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/job/DataCleanerJobContext.java
@@ -22,6 +22,7 @@ package org.datacleaner.monitor.server.job;
 import java.util.List;
 import java.util.Map;
 
+import org.datacleaner.connection.Datastore;
 import org.datacleaner.job.AnalysisJob;
 import org.datacleaner.monitor.job.JobContext;
 import org.datacleaner.monitor.job.MetricJobContext;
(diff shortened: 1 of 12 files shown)


monitor/services/src/main/java/org/datacleaner/monitor/server/job/DataCleanerJobContext.java
```diff
@@ -39,6 +40,9 @@
 
     AnalysisJob getAnalysisJob(Map<String, String> variableOverrides, Map<String, String> overrideProperties);
 
+    AnalysisJob getAnalysisJob(Map<String, String> variableOverrides, Map<String, String> overrideProperties,
+            Datastore datastore);
```
jhorcicka: In theory we could add "finals" here (and above as well).

jhorcicka:
One small code style issue, otherwise tested & LGTM.

monitor/services/src/main/java/org/datacleaner/monitor/server/job/DataCleanerJobContext.java
LosD: This is an interface. `final` doesn't make any sense here? Our CheckStyle rules would even fail a final because it is redundant.

jhorcicka: OK

Issue #1694, opened 2017-02-27
DC monitor: Improve HotFolder integration tests


ClaudiaPHI:
The tests "HotFolderIT" and "RestEndpointsIT" test hot folder scheduling.
There is a 20 seconds waiting time to check the results of the job after a file has been copied to the hot folder. Try to improve the time of these 2 tests
We would like something like a "results" endpoint which returns all results (e.g. for a job) and possibly make it so we can filter on timestamp.

PR #1696, opened 2017-03-02
Feature/1694 improve it tests

ClaudiaPHI:
fixes part of #1694 

I added an api for getting all the job results after a specific timestamp. I added tests for the controller.

### DIFF
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
@@ -71,7 +71,6 @@ public class ResultsFolderController {
 
         return result;
     }
-
     @RolesAllowed(SecurityRoles.VIEWER)
     @RequestMapping(method = RequestMethod.GET, produces = "application/json")
     @ResponseBody
@@ -90,7 +89,7 @@ public class ResultsFolderController {
             for (final RepositoryFile file : files) {
                 final Map<String, String> map = new HashMap<>();
                 final String name = file.getName();
-                // get the timestamp of the job
+                //get the timestamp of the job
                 final String timestampString = name.substring(name.lastIndexOf("-") + 1, name.indexOf("."));
                 final Timestamp jobTimestamp = new Timestamp(Long.valueOf(timestampString));
                 if (jobTimestamp.after(searchedTimestamp)) {
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
@@ -71,24 +71,32 @@ public class ResultsFolderController {
             for (final RepositoryFile file : files) {
                 final Map<String, String> map = new HashMap<>();
                 final String name = file.getName();
+                // if the timestamp is not null then we check if we can add the job,
+                // otherwise we continue with loop
                 if (searchedTimestamp != null) {
-                    // get the timestamp of the job
-                    final String timestampString = name.substring(name.lastIndexOf("-") + 1, name.indexOf("."));
-                    final Date jobDate = new Date(Long.valueOf(timestampString));
-                    if (jobDate.after(searchedTimestamp)) {
-                        map.put("filename", name);
-                        map.put("repository_path", file.getQualifiedPath());
+                    if (!isEligibleJob(searchedTimestamp, name)) {
+                        continue;
                     }
-                } else {
-                    map.put("filename", name);
-                    map.put("repository_path", file.getQualifiedPath());
                 }
-                if (map.size() > 0) {
-                    result.add(map);
-                }
+                map.put("filename", name);
+                map.put("repository_path", file.getQualifiedPath());
+
+                result.add(map);
             }
         }
 
         return result;
+    }
+
+    public boolean isEligibleJob(final Date searchedTimestamp, final String name) {
+        if (searchedTimestamp != null) {
+            // get the timestamp of the job
+            final String timestampString = name.substring(name.lastIndexOf("-") + 1, name.indexOf("."));
+            final Date jobDate = new Date(Long.valueOf(timestampString));
+            if (jobDate.after(searchedTimestamp)) {
+                return true;
+            }
+        }
+        return false;
     }
 }
--- a/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
+++ b/monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
@@ -53,52 +53,39 @@ public class ResultsFolderController {
     @ResponseBody
     public List<Map<String, String>> resultsFolderJson(@PathVariable("tenant") final String tenant,
             @RequestParam(value = "not_before", required = false) final String timestamp) {
-        
-        
-        final TenantContext context = _tenantContextFactory.getContext(tenant);
-        if (timestamp == null || Long.valueOf(timestamp) < 0) {
-            return resultsFolderJsonHelper(tenant, context);
-        }
 
-        final Date searchedTimestamp = new Date(Long.valueOf(timestamp));
+        final TenantContext context = _tenantContextFactory.getContext(tenant);
+        final Date searchedTimestamp;
+        if (timestamp == null || Long.valueOf(timestamp) < 0) {
+            searchedTimestamp = null;
+        } else {
+            searchedTimestamp = new Date(Long.valueOf(timestamp));
+        }
         final RepositoryFolder resultsFolder = context.getResultFolder();
 
         final List<Map<String, String>> result = new ArrayList<>();
 
         {
-            final List<RepositoryFile> files = resultsFolder.getFiles(null, 
-                    FileFilters.ANALYSIS_RESULT_SER.getExtension());
+            final List<RepositoryFile> files =
+                    resultsFolder.getFiles(null, FileFilters.ANALYSIS_RESULT_SER.getExtension());
             for (final RepositoryFile file : files) {
                 final Map<String, String> map = new HashMap<>();
                 final String name = file.getName();
-                // get the timestamp of the job
-                final String timestampString = name.substring(name.lastIndexOf("-") + 1, name.indexOf("."));
-                final Date jobDate = new Date(Long.valueOf(timestampString));
-                if (jobDate.after(searchedTimestamp)) {
+                if (searchedTimestamp != null) {
+                    // get the timestamp of the job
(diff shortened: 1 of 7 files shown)


LosD:
Travis got angry

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -49,16 +51,52 @@
...
+                final Map<String, String> map = new HashMap<>();
+                final String name = file.getName();
+                // get the timestamp of the job
+                final String timestampString = name.substring(name.lastIndexOf("-") + 1, name.indexOf("."));
+                final Timestamp jobTimestamp = new Timestamp(Long.valueOf(timestampString));
+                if (jobTimestamp.after(searchedTimestamp)) {
+                    map.put("filename", name);
+                    map.put("repository_path", file.getQualifiedPath());
+                }
+                if (map.size() > 0) {
```
LosD: This is really weird. What's the difference from including it in the previous `if`?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -49,16 +51,52 @@
...
-    public List<Map<String, String>> resultsFolderJson(@PathVariable("tenant") final String tenant) {
+    public List<Map<String, String>> resultsFolderJson(@PathVariable("tenant") final String tenant,
+            @RequestParam(value = "time", required = false) final String timestamp) {
+        
+        
+        if (timestamp == null || Long.valueOf(timestamp) < 0) {
+            return resultsFolderJsonHelper(tenant);
+        }
+
+        final TenantContext context = _tenantContextFactory.getContext(tenant);
```
LosD: Duplicated in helper. Why not just pass it along?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -49,16 +51,52 @@
...
+        final Timestamp searchedTimestamp = new Timestamp(Long.valueOf(timestamp));
+        final RepositoryFolder resultsFolder = context.getResultFolder();
```
LosD: Duplicated in helper. Why not just pass it along?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -19,6 +19,7 @@
  */
 package org.datacleaner.monitor.server.controllers;
 
+import java.sql.Timestamp;
```
LosD: No SQL packages, please. If you something a little more modern that Date and Calendar, there's the Java 8 `java.time` package, which also has good tools for checking if one temporal value is after another.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -49,16 +51,52 @@
     @RolesAllowed(SecurityRoles.VIEWER)
     @RequestMapping(method = RequestMethod.GET, produces = "application/json")
     @ResponseBody
-    public List<Map<String, String>> resultsFolderJson(@PathVariable("tenant") final String tenant) {
+    public List<Map<String, String>> resultsFolderJson(@PathVariable("tenant") final String tenant,
+            @RequestParam(value = "time", required = false) final String timestamp) {
```
LosD: Name here is a bit weird. Wouldn't "not_before", or something to that effect, better match the purpose?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
ClaudiaPHI: There would be an empty array value in the json

jhorcicka:
Tested and LGTM. After @LosD approves the changes I believe we can merge it.

LosD:
I see line-comments addressed, but it doesn't look like the overall review comment was read at all.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
LosD: Errr, isn't map size only > 0 if the the previous `if` is true? So why not move it into the previous `if`?

ClaudiaPHI:
I did not want to change original the method, but I was thinking someone else might be using it and expecting a specific type of results, as in the one getting the results of a timestamp doesn't add the empty maps to the results.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -61,9 +70,22 @@
...
                 final Map<String, String> map = new HashMap<>();
-                map.put("filename", file.getName());
-                map.put("repository_path", file.getQualifiedPath());
-                result.add(map);
+                final String name = file.getName();
+                if (searchedTimestamp != null) {
+                    // get the timestamp of the job
+                    final String timestampString = name.substring(name.lastIndexOf("-") + 1, name.indexOf("."));
+                    final Date jobDate = new Date(Long.valueOf(timestampString));
+                    if (jobDate.after(searchedTimestamp)) {
```
LosD: These five lines should really all be in their own method, along with the parsing of the timestamp in the beginning.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -61,9 +70,22 @@
...
+                        map.put("filename", name);
+                        map.put("repository_path", file.getQualifiedPath());
+                    }
+                } else {
+                    map.put("filename", name);
+                    map.put("repository_path", file.getQualifiedPath());
```
LosD: Then you can nuke this nasty duplicate code

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -61,9 +70,22 @@
...
+                }
+                if (map.size() > 0) {
```
LosD: And this can just go into the if above

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -49,9 +51,16 @@
...
+    public List<Map<String, String>> resultsFolderJson(@PathVariable("tenant") final String tenant,
+            @RequestParam(value = "not_before", required = false) final String timestamp) {
 
+        final TenantContext context = _tenantContextFactory.getContext(tenant);
+        final Date searchedTimestamp;
+        if (timestamp == null || Long.valueOf(timestamp) < 0) {
+            searchedTimestamp = null;
+        } else {
+            searchedTimestamp = new Date(Long.valueOf(timestamp));
+        }
```
LosD: This can all go into the `isEligibleJob` method

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -61,12 +70,33 @@
                     resultsFolder.getFiles(null, FileFilters.ANALYSIS_RESULT_SER.getExtension());
             for (final RepositoryFile file : files) {
                 final Map<String, String> map = new HashMap<>();
-                map.put("filename", file.getName());
+                final String name = file.getName();
+                // if the timestamp is not null then we check if we can add the job,
+                // otherwise we continue with loop
+                if (searchedTimestamp != null) {
```
LosD: This seems redundant

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -61,12 +70,33 @@
...
+                    if (!isEligibleJob(searchedTimestamp, name)) {
```
LosD: Not really important, but this is pretty weird to me. Why the reverse logic instead of adding the result when isEligibleJob is true?

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
ClaudiaPHI: that timestamp is null if the user wants all the results. Therefore I do not want to go the eligible method if I do not care about the timestamp

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
ClaudiaPHI: I use it to avoid adding twice to the map. and nuke the second else

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -49,24 +51,46 @@
...
+        if (timestamp == null || Long.valueOf(timestamp) < 0) {
+            // we do not care about the timestamp, therefore we add the job to the result
+            return true;
+        } else {
+            searchedTimestamp = new Date(Long.valueOf(timestamp));
+        }
+        // get the timestamp of the job
+        final String timestampString = name.substring(name.lastIndexOf("-") + 1, name.indexOf("."));
+        final Date jobDate = new Date(Long.valueOf(timestampString));
+        if (jobDate.after(searchedTimestamp)) {
```
LosD: Not important at all, but you can just return `jobDate.after(searchedTimestamp)` directly.

monitor/services/src/main/java/org/datacleaner/monitor/server/controllers/ResultsFolderController.java
```diff
@@ -49,24 +51,46 @@
...
         final List<Map<String, String>> result = new ArrayList<>();
 
         {
             final List<RepositoryFile> files =
                     resultsFolder.getFiles(null, FileFilters.ANALYSIS_RESULT_SER.getExtension());
             for (final RepositoryFile file : files) {
                 final Map<String, String> map = new HashMap<>();
-                map.put("filename", file.getName());
+                final String name = file.getName();
+                if (!isEligibleJob(timestamp, name)) {
```
LosD: Still not important, but this could easily be reverted to a normal flow now:
```java
if (isEligibleJob(timestamp, name)) {
    map.put("filename", name);
    map.put("repository_path", file.getQualifiedPath());
    result.add(map)
}
```

PR #1698, opened 2017-03-08
Feature/1693 monitor removal api remains

jhorcicka:
Resolves #1693 

Everything but "api" was removed from "monitor" sub-module. 
Check/consider putting "widgets" and "pentaho" back. 

PlaceholderComponentJob class was put back, but to /engine/core/...

PR #1699, opened 2017-03-08
Feature/1693 unused dependencies removal

jhorcicka:
Related to #1698 

This change contains the removal of "unused" (according to maven) dependencies. It needs to be tested if they are all really unneeded.


jhorcicka:
This change introduced more problems then potential benefits. Closing.

PR #1701, opened 2017-03-09
Update multimapped combobox visibility

### LINKED ISSUE #1700: MultipleMappedComboBoxPropertyWidget does not set combox visibility on batch update
The `MultipleMappedComboBoxPropertyWidget` is missing functionality to update visibility when a batch update happens.

The issue manifest itself when the user of the widget adds available columns after first initialization, e.g. when there's multiple tables, and the user needs to choose one before columns can be shown.

LosD:
This overrides the `onValuesBatchSelected` method, ensuring that combobox visibility is properly updated on batch updates.

Fixes #1700


jhorcicka:
Tested and LGTM. 

sheet-2

PR #1703, opened 2017-03-22
Ordering of items in MultipleMappedComboBoxPropertyWidget

### LINKED ISSUE #1702: Items in MultipleMappedComboBoxPropertyWidget are in random order
Items are added to the combo-box using an unsorted set, as a result of which the order of the items is random and unpredictable.

arjansh:
Fixes #1702.

The column meanings were retrieved from the ColumnMeaningCollection in an ordered manner and after that assigned in order to an unordered Set implementation. This is changed, so now the are no longer retrieved in an ordered manner, but they are assigned to an ordered Set implementation which is instanctiated with a Comparator which ensures the ordering within the Set.


jhorcicka:
Tested and LGTM.

PR #1706, opened 2017-03-24
Override variables inside output data streams

### LINKED ISSUE #1705: Variables in output datastreams are not overridden
`JaxbJobReader` does does not propagate variables overrides to output datastreams. That means that it is impossible to use variables or templates properly with output data streams.

LosD:
This will pass the overrides along to the output data stream jobs, and redo the override for those.

It would be better to only have the variables global, but that would also require changes to the writer, so for now this it good enough, I think.

Fixes #1705

### DIFF
--- a/engine/xml-config/src/test/java/org/datacleaner/job/JaxbJobReaderTest.java
+++ b/engine/xml-config/src/test/java/org/datacleaner/job/JaxbJobReaderTest.java
@@ -393,7 +393,7 @@ public class JaxbJobReaderTest extends TestCase {
         assertTrue(file.exists());
         final AnalysisJobBuilder ajb = reader.create(file);
 
-        AnalysisJobBuilder odsjb =
+        final AnalysisJobBuilder odsjb =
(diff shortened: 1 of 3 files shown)


engine/xml-config/src/test/java/org/datacleaner/job/JaxbJobReaderTest.java
```diff
@@ -389,11 +389,14 @@ public void testReadVariables() throws Exception {
         final DataCleanerConfiguration configuration = new DataCleanerConfigurationImpl().withDatastores(datastore)
                 .withEnvironment(new DataCleanerEnvironmentImpl().withDescriptorProvider(descriptorProvider));
         final JaxbJobReader reader = new JaxbJobReader(configuration);
-        final File file = new File("src/test/resources/example-job-variables.xml");
+        final File file = new File("src/test/resources/example-job-variables-ods.analysis.xml");
```
jhorcicka: What does "ods" mean?

jhorcicka:
Tested and LGTM!

PR #1710, opened 2017-03-29
Feature/addresslines not mapped easily

### LINKED ISSUE #1709: Make Adddressline map easily
Improve the default mapping.When I have a datastore with fields named "Addressline 1" and "Addressline 2" and I use this, they aren't given the "Addressline" meaning, but the "Other" meaning. I would expect that with those names, they would have been recognized as address lines. 
The same with fields prefixed with starting with 'FLD'

ClaudiaPHI:
fixes #1709 
Removed the numbers at the end of the fields, as well as the fields prefixed by 'FLD'

### DIFF
--- a/api/src/main/java/org/datacleaner/metadata/DefaultColumnMeaningCollection.java
+++ b/api/src/main/java/org/datacleaner/metadata/DefaultColumnMeaningCollection.java
@@ -56,7 +56,7 @@ public class DefaultColumnMeaningCollection implements ColumnMeaningCollection {
     }
 
     private static String standardizeForMatching(String key) {
-        key = key.trim();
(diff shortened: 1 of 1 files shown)


LosD:
To me, this is way too limited. Focusing specifically on "FLD" or numbers is not really fixing the main issue with the algorithm not recognizing partial words. This fixes the test cases instead of the issue.

IMHO the best way is to divide along word boundaries, and check individual words (maybe as well as combinations of thing, to avoid words like "address line" being missed.

OTOH, the is a small step in the right direction, so we might want to make a new issue to be fixed later and merge this for now

LosD:
(code LGTM)

PR #1716, opened 2017-04-06
Add option to configure metrics export option when scheduling job

arjansh:
Add the option to active metrics export when scheduling jobs from DataCleaner desktop.

### DIFF
--- a/monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinitionModel.java
+++ b/monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinitionModel.java
@@ -19,24 +19,23 @@
  */
 package org.datacleaner.monitor.scheduling.model;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class ScheduleDefinitionModel {
-    private final String _hotFolder;
+    @JsonProperty("hotFolder")
+    private String _hotFolder;
     
     @JsonProperty("exportMetrics")
     private boolean _exportMetrics;
-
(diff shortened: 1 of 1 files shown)


monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinitionModel.java
```diff
@@ -24,6 +24,9 @@
 
 public class ScheduleDefinitionModel {
     private final String _hotFolder;
+    
+    @JsonProperty("exportMetrics")
+    private boolean _exportMetrics;
 
     @JsonCreator
     public ScheduleDefinitionModel(@JsonProperty("hotFolder") final String hotFolder) {
```
LosD: Maybe we should also move this `@JsonProperty` to a setter? I'd prefer this an immutable object, but as long as we're consistent, both is fine with me.

LosD:
Single comment on consistency, otherwise LGTM.

I don't think there's anything to test.

monitor/api/src/main/java/org/datacleaner/monitor/scheduling/model/ScheduleDefinitionModel.java
arjansh: I was considering the immutability too, but I'm afraid we may want to add more options later and that will result in a constructor with a lot of arguments. Just say it, if you'd rather have it immutable, I'll change it.

LosD: No, it's fine. But make them consistent so they're both using setters, then.

ClaudiaPHI:
LGTM!

PR #1718, opened 2017-04-18
Datahub 9.0 url update

jakubneubauer:


LosD:
LGTM, though I have no idea if the paths are correct :)

LosD:
If you're sur about the paths,, merge when Travis is happy.

engine/core/src/main/java/org/datacleaner/metamodel/datahub/DataHubUpdateConnection.java
```diff
@@ -31,7 +31,7 @@
  * Note: Some REST controllers do not need the tenant info. Others do.
  */
 public class DataHubUpdateConnection {
-    public final static String CONTEXT_PATH = "/service/v1";
+    public final static String CONTEXT_PATH = "/api/v1";
```
reinierdejong: should be /datahub/api/v1

jakubneubauer:
Now it is correct.

LosD:
Cool. Merging!

Issue #1719, opened 2017-05-02
Move all images and colors to own module


LosD:
To make it easier to use specialized images and colors, we should move all of these into its own module, which can then be replaced or extended.

PR #1723, opened 2017-05-11
#1719 Colors were put into a properties file

jhorcicka:
Related to already closed issue #1719. 

Colors were put into a properties file (in the same manner as image paths).

### DIFF
--- a/desktop/ui/pom.xml
+++ b/desktop/ui/pom.xml
@@ -279,5 +279,10 @@
 			<version>1.4</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.eobjects.datacleaner</groupId>
+			<artifactId>DataCleaner-oss-branding</artifactId>
+			<version>5.1.6-SNAPSHOT</version>
+		</dependency>
 	</dependencies>
 </project>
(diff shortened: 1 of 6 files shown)


LosD:
Wow, that was a lot more that I expected! :)

I really like the rgba one, but I'm a bit worried about the `color.someothercolor` one. There's potential for endless loops here.

Since the end result (though probably not going to happen right now) is preferably to nuke any keys that has a specific color, I might be better to not have these references, and just use the same color (maybe with a comment like "same as `color.red`").

LosD:
Hmmm... As we're not really going to use the ResourceBundles, as we talked about yesterday, you might instead want to make it into a utility class, probably somewhere outside oss-branding (or we'd have to copy it to any other branding class)

LosD:
By the way, in CSS, they've deprecated rgba for rgb. You might (or might not) want to follow them.

> Note: In the CSS Color Module Level 4 spec, rgba() has been defined as a legacy function with identical grammar and behaviour to rgb(); in effect, an alias. Going forward, both can accept exactly the same parameters.

https://developer.mozilla.org/en/docs/Web/CSS/color_value#rgb

jhorcicka:
I removed variables functionality in the properties file. Let's stop with additional improvements right now. Movement to a separate utility class can be done later and `rgba` => `rgb` as well. 

@LosD: Any other comments/remarks before merging?

LosD:
Nope! Merging :)

Issue #1631, opened 2016-11-30
Create a hashing transformer


jhorcicka:
The component should include all of the standard hashing algorithms and produce a hash and a length of the original data.

jhorcicka:
Apart from (easily available at MessageDigest)
```
MD2, MD5, SHA, SHA-224, SHA-256, SHA-384, SHA-512
```
what algorithms do we want (@LosD, @arjansh)? 

```
CRC (16, 32), MD4, MD6, ... ?
```

arjansh:
@jhorcicka Unless they're also easily available, I wouldn't add them up front.

jhorcicka:
I added CRC32 (it was simple), the rest can be added when needed.

LosD:
Having a CRC in there doesn't really make sense to me. It's not a hashing function, and not very widely used above protocol levels.

We don't really want to make the list the user has to choose between longer, just because we can.

jhorcicka:
OK, I'll remove it then.

Issue #1621, opened 2016-11-20
Salesforce Datatore:  Security Token Should be OPTIONAL


d416:
Salesforce.com can be configured to NOT use security tokens - this is an administrative option and can be invoked for various reasons (e.g. if the org is locked down by IP address or Single Sign-on is set up)

Given that a) tokens are optional, and b) to generate a token requires a password reset,  the "Salesforce Security Token:" field on the Salesforce.com Datastore setup dialog box should be OPTIONAL.  Currently this field is required and prevents anyone without a token from authenticating.

kaspersorensen:
Interesting point of view. We are leveraging Apache MetaModel for the connectivity to Salesforce. Could you maybe echo your request on the bugtracker for that? https://issues.apache.org/jira/browse/METAMODEL

d416:
This issue is with the authentication layer, not the metadata, therefore the issue is with the DataCleaner code.... like basically this file line 116:
https://github.com/datacleaner/DataCleaner/blob/70e7978fe45dbbdd0cd307252d05a127ec1fdc51/desktop/ui/src/main/java/org/datacleaner/windows/SalesforceDatastoreDialog.java

davkrause:
Noticed this from my emails today and started a pull reqest here: https://github.com/apache/metamodel/pull/143

LosD:
Partially fixed by #1621; Allows the empty string as token.

Still needs a release (and integration, of course) containing apache/metamodel#143 for a full fix, where null tokens are also allowed.

kaspersorensen:
Fixed as of DC CE 5.3.0

PR #1725, opened 2017-05-12
Issue #1631: Made security token optional in dialog

davkrause:


LosD:
Blocked by release of MetaModel containing apache/metamodel#143 (@davkrause please correct me if I'm wrong)

davkrause:
I think this is good enough on it's own since empty string is already working even before that change in MetaModel (which handles null also).

davkrause:
The issue does not seem to be properly linked. Here is a link: #1621
@d416 provided the idea to fix this just in the dialog

PR #1727, opened 2017-05-19
[PROPOSAL] Fixed width column names

davkrause:
Hi all,

This is a proposal - not yet fully working. But maybe you could provide youre thoughts.

I think the fixed width dialog in DataCleaner today is missing the ability to put column names. I see that some work was done, but even that is also very unintuitive. Most ETL tools that I have tried do this with a simple table with column names and size. So I tried implementing that. To make it work well, I decided to first make a more general purpose "DCEditableTable" class which shares a new base-class, "DCBaseTable" with "DCTable".

Please provide your feedback. Do you also think this is better? So I should continue to make it work.

I made a main method myself to test it, like this:
```
    public static void main(String[] args) {
        LookAndFeelManager.get().init();
        
        DataCleanerConfiguration configuration = new DataCleanerConfigurationImpl();
        DomConfigurationWriter configurationWriter = new DomConfigurationWriter();
        UserPreferences userPreferences = new UserPreferencesImpl(null);
        MutableDatastoreCatalog mutableDatastoreCatalog =
                new MutableDatastoreCatalog(configuration.getDatastoreCatalog(), configurationWriter, userPreferences);
        WindowContext windowContext = new DCWindowContext(configuration, userPreferences, null);

        int[] valueWidths = { 6, 4, 12, 12 };
        List<String> customColumnNames = Arrays.asList("Entry ID", "Period", "Post date", "Account number");
        
        FixedWidthDatastore originalDatastore = new FixedWidthDatastore("My datastore",
                new FileResource("src/test/resources/example_fixed_width_file.txt"), "example_fixed_width_file.txt",
                "UTF8", valueWidths, true, true, true, 0, customColumnNames);

        FixedWidthDatastoreDialog dialog = new FixedWidthDatastoreDialog(originalDatastore, mutableDatastoreCatalog,
                windowContext, configuration, userPreferences);
        dialog.open();
    }
```


davkrause:
I've updated the dialog some more. It is now quite functional. Attached is a screenshot with example file:

image

I still see an issue before I consider this really working. Maybe you can help: The data source is saved nicely when I click the button. And I can open it quite fine again too. But for a reason, if I restart DataCleaner, then the column names disappear. The column widths are saved, but the column names go back to A, B, C etc... Maybe that part is not persisted - where is the persistance of data source done?

kaspersorensen:
Hi @davkrause, awesome contribution if you ask me. Would give it my +1.

Gave it a quick check and it works quite well. A small observation from my side is that deleting a cell's contents is rather cumbersome. I had to double-click the cell to get the characters selected, and then hit DEL. It would be (another) good improvement if you added the capability clear a cell's content upon just clicking DEL when the cell is selected (and not editing).

You have a few "final" variable declaration omissions that you should fix or else our checkstyle rules are breaking (hence the Travis CI build failure).

The issue you are mentioning can be fixed if you take a look at `DomConfigurationWriter` in the `xml-config` module. Shouldn't be too difficult to fix - I see the configuration XML schema already supports the column names (`configuration.xsd` in the same module). I'll be happy to help with this stuff if you need it.

davkrause:
thx for the help. I'll try that

kaspersorensen:
I've tried the latest revision of this PR and it works well IMO.

kaspersorensen:
Hi @davkrause, thank you for the contribution and sorry for the wait. We had to cut a release first.

Merging now! :-)

PR #1736, opened 2017-06-13
Upgraded MetaModel dependency to 5.0-RC2. Fix lots of compiler warnings.

kaspersorensen:
This PR exercises the release candidate of MetaModel (5.0-RC2) which is currently being votes on. I believe it to be a good release though, and think it's time we get DataCleaner's codebase updated to take the new 5.x codebase of MetaModel. With this PR the first part of that is done.

__Note__: While this PR works right now, it won't work when the release-staging repository of Apache MetaModel goes away in 3-4 days. We can either merge it early and fix the repo later, or wait.


kaspersorensen:
OMG I see that my CRLF settings aren't compatible with what you guys use nowadays. Closing this PR so that I can get that fixed first ...

PR #1739, opened 2017-06-17
Update README modules list

davkrause:
The module list in README is not showing correctly in github. I updated indents and removed modules I could not find.


kaspersorensen:
LGTM

PR #1740, opened 2017-06-19
fixed URLs in DataHub connector according DataHub 9.0

jakubneubauer:
To be sure I compared the path engine/core/src/main/java/org/datacleaner/metamodel/datahub - master vs. release/5.1 and it seems to be aligned now.


LosD:
Changes LGTM. Let's merge when/if Travis is happy.

PR #1741, opened 2017-07-11
Workspace chooser at startup

jakubneubauer:
When started, DataCleaner desktop asks for workspace. By default it offers $HOME/.datacleaner.
And if there are some sub-folders matching pattern "[0-9\.]*" - it means workspace locations from previous versions, they are offered too.
screenshot from 2017-07-11 11-51-35

The configuration of this dialog is in $HOME/.datacleaner/global_config.xml:
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<datacleanerConfiguration>
    <workspaceConfiguration>
        <defaultWorkspace>/home/jakub/.datacleaner</defaultWorkspace>
        <showDialog>true</showDialog>
        <workspaces>
            <workspace>/home/jakub/.datacleaner</workspace>
        </workspaces>
    </workspaceConfiguration>
</datacleanerConfiguration>
```
If the option "Don't show again" is checked, the dialog will be not shown on next application start. However, the user can force it from Options dialog:

screenshot from 2017-07-11 11-53-43

Clicking the button exits the application with exit code 3 and the startup script will be responsible to restart with the new option:

There is a new commandline option `-ws` which will enforce showing the workspace chooser dialog.

### DIFF
--- a/desktop/api/src/main/java/org/datacleaner/bootstrap/WindowContext.java
+++ b/desktop/api/src/main/java/org/datacleaner/bootstrap/WindowContext.java
@@ -99,7 +99,17 @@ public interface WindowContext {
     boolean showExitDialog();
 
     /**
-     * Requests the application to exit.
+     * Requests the application to exit with exit code 0.
      */
     void exit();
+
+    /**
+     * Requests the application to exit.
+     */
+    void exit(int exitCode);
+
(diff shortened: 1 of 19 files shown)


LosD:
Not valid for Github

PR #1742, opened 2017-07-19
Removal of "datahub" artifacts in OSS version of DataCleaner.

kaspersorensen:
As part of detaching the OSS edition of DataCleaner from the commercial edition, I suggest via this PR to remove the DataHub artifacts from the OSS edition.


kaspersorensen:
Merging as per lazy concensus.

PR #1746, opened 2017-07-31
Updated MetaModel version to 5.0-RC3

kaspersorensen:


kaspersorensen:
Merging as per lazy concensus

PR #1750, opened 2017-08-20
#1748: Skipping cobertura (and other unnecesary stuff) on -DskipTests

### LINKED ISSUE #1748: Building with -DskipTests fails in xml-config project
For some reason I am getting compile errors when running a `mvn clean install -DskipTests`. The error indicates that it's trying to compile the test-src of the `xml-config` before code-generation has taken place. Must be some ordering of mvn plugins issue I think.

kaspersorensen:
Fixes #1748 

Seems that cobertura was the culprit here. And in addition to cobertura I felt that we should probably skip checkstyle when `-DskipTests` is set so that it really becomes more of a compile-only style build.


LosD:
LGTM!

PR #1758, opened 2017-10-20
Added a mechanism to adjust the UI according to screen resolution

kaspersorensen:


LosD:
Full-file reformatting everywhere. Many of them to what looks like possibly a wrong style? Did you import the formatter rules first?

components/html-rendering/src/main/java/org/datacleaner/result/html/HtmlAnalysisResultWriter.java
```diff
@@ -237,7 +236,7 @@ protected void writeBody(final Writer writer, final Map<ComponentJob, HtmlFragme
                     final ComponentDescriptor<?> descriptor = componentJob.getDescriptor();
                     if (!descriptor.equals(lastDescriptor)) {
                         final ComponentDocumentationWrapper wrapper = new ComponentDocumentationWrapper(descriptor);
-                        final String iconSrc = wrapper.getIconSrc(IconUtils.ICON_SIZE_MEDIUM);
+                        final String iconSrc = wrapper.getIconSrc(22);
```
LosD: Magic number

components/html-rendering/src/main/resources/org/datacleaner/documentation/template_component.html
```diff
@@ -81,7 +81,7 @@
 				</#if>
 				<div class="jumbotron">
 					<h1>
-						<img src="${component.iconSrc(component.iconSizeLarge)}" alt="" /> <span>${component.name}</span>
+						<img src="${component.iconSrc(32)}" alt="" /> <span>${component.name}</span>
```
LosD: Magic number

desktop/api/src/main/java/org/datacleaner/util/WidgetUtils.java
```diff
@@ -75,34 +75,24 @@
...
+    private static final WidgetScreenResolutionAdjuster adjuster = WidgetScreenResolutionAdjuster.get();
+
+    private static final int PIX_1 = adjuster.adjust(1);
+    private static final int PIX_2 = adjuster.adjust(2);
+    private static final int PIX_3 = adjuster.adjust(3);
+    private static final int PIX_4 = adjuster.adjust(4);
+    private static final int PIX_6 = adjuster.adjust(6);
+    private static final int PIX_9 = adjuster.adjust(9);
+    private static final int PIX_10 = adjuster.adjust(10);
+    private static final int PIX_17 = adjuster.adjust(17);
```
LosD: This whole block is pretty nasty (and quite inconsistently used)

If you really want the indirection, how about a call instead?

```java
int pix(int in) {
    return adjuster.adjust(in);
}

public static final Object BORDER_MENU_ITEM = new EmptyBorder(pix(2), pix(2), pix(2), pix(2));
```

The JIT compiler should optimize frequent calls pretty well.

components/html-rendering/src/main/java/org/datacleaner/result/html/HtmlAnalysisResultWriter.java
kaspersorensen: Yes but I was worried about using non-magic numbers here because I saw that depending on the screen size, the generated HTML would come out different. And I don't think we should change anything related to HTML generation in this round.

components/html-rendering/src/main/resources/org/datacleaner/documentation/template_component.html
kaspersorensen: Same as above.

desktop/api/src/main/java/org/datacleaner/util/WidgetUtils.java
kaspersorensen: I almost feel like adding that to the adjuster then, as a static method that one can do a static import of.

kaspersorensen:
I should add that I would also applaude anyone building a UI based on new technology. But it's one of those crazy big projects that I don't think can take place of improvements here and now. The approach here is hacky for sure, and if you DO try it, you'll see that I haven't added the adjuster everywhere yet, so it's even an inconsistent fix. But it takes the first couple of steps and I think that's ... well, a good first step ;-) I would be overwhelmed if I tried to implement the adjuster everywhere in one PR.

desktop/api/src/main/java/org/datacleaner/util/WidgetUtils.java
kaspersorensen: Oh well, the name `pix` is a bit too internal for that. So I did what you suggested :-)

components/html-rendering/src/main/java/org/datacleaner/result/html/HtmlAnalysisResultWriter.java
LosD: Fair enough

LosD:
I'm happy, but Travis/CheckStyle isn't (one-character parameter name).

desktop/api/src/main/java/org/datacleaner/util/WidgetUtils.java
```diff
@@ -728,4 +692,7 @@ public static void setDefaultButtonStyle(final AbstractButton button) {
         }
     }
 
+    private static int pix(int p) {
```
LosD: Okay, not 100% happy, though very minor. Missing final (and of course the one-char name from Travis) :)

kaspersorensen:
Fixed the things! :-)

LosD:
LGTM!

PR #1759, opened 2017-10-24
Issue #1753 remove non CE features

### LINKED ISSUE #1753: [PROPOSAL] Delete unused, non-CE and deprecated components and interfaces
I would like to propose that the community edition of DataCleaner move forward in simplifying life for the community contributors by removing stuff from the project that anyways does not apply to community users:

* Backwards compatibility of deserializing legacy result files.
* Backwards compatibility for legacy components.
* Integration with the monitor application
* Integration with DataCloud

Maybe there are other things?

By getting rid of the components, we can make a proper cut in the code base and we would also be able to attract contributors more easily since there will be less "baggage" for them to carry around and worry about.

kaspersorensen:
Fixes #1753

This is a rather aggressive cleanup in our codebase to remove non-CE relevant code. Here's a listing of stuff I removed (got my hands proper dirty this time so I hope I remember all):

 * DataCloud and remote components
 * DC Monitor
 * ExtensionSwap
 * UsageLogger
 * Old org.eobjects component interfaces
 * A lot of "various" deprecated classes
 * Some tests that would fail after cleaning up because they tried to deserialize deleted/deprecated items


LosD:
Whoah, that's a LOOOOT of code. I'll start digging :)

LosD:
Just curious: Any reason for the fork instead of a branch (same with the other PR), just habit? :)

kaspersorensen:
With OSS projects I like to use forks so as to not needlessly exercise my administrative privileges, but rather go through the "contributor" experience ;-)

LosD:
Hmmmm... My internal pedant wants to point out that it's not administrative privileges, but makes sense :)

LosD:
Do we really want to get rid of the ExtensionSwap? I realize that it'll need to be reworked with an internal UI (e.g. using something like a simple GitHub-pages hosted JSON that links to the extensions, which could in most cases simply be release files in GitHub), but it seems like a bit of a shame to completely lose it.

OTOH, if anyone does that work, s/he'd probably have to start all over anyway, and we still have the base extension loading, so it's probably all good.

BTW, you left in the "Browse the ExtensionSwap" menu item in _options_/_Extensions_, was that on purpose?

desktop/ui/src/main/java/org/datacleaner/windows/FileTransferProgressWindow.java
```diff
@@ -32,7 +32,6 @@
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.metamodel.util.Action;
 import org.datacleaner.actions.DownloadFilesActionListener;
-import org.datacleaner.actions.PublishResultToMonitorActionListener;
```
LosD: There's a reference to this class in the JavaDoc below

engine/core/src/main/java/org/datacleaner/util/CollectionUtils2.java
```diff
@@ -125,20 +125,6 @@ private CollectionUtils2() {
...
     /**
-     *
-     * @return
-     *
-     * @deprecated use Google Guava's {@link CacheBuilder},
-     *             {@link #createCache(int, long)} or something similar if
-     *             needed.
-     */
-    @Deprecated
-    public static <K, V> ConcurrentMap<K, V> createCacheMap() {
```
LosD: ConcurrentMap is now an unused import

desktop/ui/src/main/java/org/datacleaner/windows/FileTransferProgressWindow.java
kaspersorensen: Fixed

engine/core/src/main/java/org/datacleaner/util/CollectionUtils2.java
kaspersorensen: Fixed

LosD:
LGTM!

PR #1761, opened 2017-10-29
Convert2java: HtmlRenderer, CompositeBodyElement and BooleanAnalyzer res

davkrause:
These conversions mean that I can build much more of DC without Scala
and that I can exclude scala from my dependencies when depending on
datacleaner-html-rendering


kaspersorensen:
Generally this change looks fine with me but please check the build failure.

kaspersorensen:
Looking into html-rendering module a bit. It would be nice to get rid of scala entirely here. As you mention, it adds a transitive dependency to something that is rather core.

davkrause:
OK, I work on it

kaspersorensen:
Nice work and thank you @davkrause.

PR #1762, opened 2017-10-30
Adjusted High DPI thresholds after testing on an actual High DPI screen!

kaspersorensen:
Eh, some corrections to my previous update for High DPI screens. This one it actually looks rather OK on a high DPI screen ;-)

Biggest oversight was the max vs min thing. It meant that a factor of 3 would more or less always be applied to high DPI screen rendering.


LosD:
LGTM. Since I can't test, I'll trust your testing :)

PR #1780, opened 2018-01-10
MetaModel 5.0.1 and sanitizing of other dependencies

kaspersorensen:
I actually just wanted to update the MM dependency to 5.0.1 ;-)

But then I came across several dependencies that aren't used anymore. I think these where used in the DC monitor codebase maybe. But I figured it'd be good to weed them out.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
 		<httpcomponents.version>4.4.1</httpcomponents.version>
 		<guava.version>16.0.1</guava.version>
 		<curator.version>2.6.0</curator.version>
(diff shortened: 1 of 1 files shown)


LosD:
Is there a way to not only remove GWT, but to take it out back and shoot it? :)

LGTM, though I haven't tested it. I guess it's pretty much a matter of "works or doesn't", so I'm fine with just merging if you tried assembling and running it. Otherwise I'll try quickly running it tonight.

LosD:
Works for me. Merging

PR #1787, opened 2018-05-21
Issue #1786: Improved file textfield's resizing capabilities

### LINKED ISSUE #1786: Component config dialog requires resize to display file path on High-DPI screens
I've been using DataCleaner quite a bit recently on a high-dpi screen, and in general the tweaks we've applied works. Or at least they're mostly more good than not.

But there is one thing that really gets annoying: The component config dialogs for components such as "Write CSV file", or any other component that has a file property, don't display correctly. The file path text box is not shown unless you resize (enlarge) the window slightly. This should be fixed.

kaspersorensen:
Now resizing (specially making the window smaller) does not hide the text box used in a file selector:

image

Fixes #1786

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/widgets/AbstractResourceTextField.java
+++ b/desktop/ui/src/main/java/org/datacleaner/widgets/AbstractResourceTextField.java
@@ -19,10 +19,11 @@
  */
 package org.datacleaner.widgets;
 
-import java.awt.GridBagLayout;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.swing.Box;
+import javax.swing.BoxLayout;
 import javax.swing.JButton;
 import javax.swing.JComponent;
 import javax.swing.JFileChooser;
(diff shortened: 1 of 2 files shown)


LosD:
That's... Pretty horrible. Really no way to doing this without going to a much more complex layout?

kaspersorensen:
Maybe I could try with BoxLayout. But I'm sure the solution is to use another layout than FlowLayout. And then I just preferred GridBag because it has weights, so I could adjust how much to let the individual pieces influence the width of the thing.

The reason FlowLayout doesn't work for this type of widget is that it "wraps" components to the next line/flow vertically. So when you made the window smaller the parts started just disappearing because they where moved to an invisible line below.

LosD:
Oh sure, flow layout is wrong, GridBagLayout is just such a nasty complicated layout for simple stuff. Not a blocker, though.

kaspersorensen:
Heh, I tried with a `BoxLayout` and it works even better actually. So now it's simpler and if Travis doesn't complain, I suggest merging and then I'll open a vote for releasing.

LosD:
Great! Travis happy, merging! :)

zhangdw115:
When the explorer was quoted, it was found that the Chinese code could not be displayed.

PR #1788, opened 2018-05-21
Fix bitrot exposed by JDK 10

kaspersorensen:
Tried building DataCleaner using JDK 10 and the compiler (1) doesn't allow one of our casts because of a generalization in the `DefaultMutableTreeNode` class, and (2) made me aware of a wrong key we're using when looking up a key (of wrong type) in a map.


LosD:
LGTM

PR #1790, opened 2018-05-21
1722 better component search

### LINKED ISSUE #1722: Component search based on keywords/alias - "Table lookup" not shown when inputting "join"
I was demonstrating DC to a colleague this morning and a question came up: "How do I join two tables then?". I was forgetting that the name of the component is "Table lookup", so for a moment I was struggling to find it. In the component search/filter box I put in "join" but nothing showed up.

So I could formulate this issue / request two ways:

 * In particular: The "Table lookup" should somehow show up when searching for "join"
 * In general: Probably the filter should be expanded to look for aliases or keywords (I assume "join" is somewhere in the "Table lookup"s metadata.

kaspersorensen:
Fixes #1722 

Tested it on both "split" (displays "Tokenizer") and "join" (displays "Table lookup")

### DIFF
--- a/desktop/ui/src/main/java/org/datacleaner/connection/DatastoreDescriptors.java
+++ b/desktop/ui/src/main/java/org/datacleaner/connection/DatastoreDescriptors.java
@@ -276,7 +276,7 @@ public class DatastoreDescriptors {
         }
 
         // composite datastore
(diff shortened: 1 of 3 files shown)


LosD:
LGTM (even though an irrelevant branch was merged)

PR #1793, opened 2018-07-31
Upgraded MetaModel to 5.1.0. Disabled elasticsearch native protocols. And bumped Jackson to latest.

kaspersorensen:


LosD:
LGTM, but since I'm reviewing on my phone (I'm on vacation without a computer), I haven't been super thorough. Since it is mostly deletions, the actual places where the remaining code is touched is pretty slim, so maybe that's good enough. 

Only comment is what seems to be an unrelated change to the logging.

kaspersorensen:
Initially the Travis builds where failing because logging data succeeded 4 mb (which is apparently their maximum). So I disabled some debug and info logging that was enabled for test runs. I suspect that the reason for the increased logging is due to changes in dependencies and maybe some of the new versions of the dependencies has increased their logs. I noticed a lot of logs coming out of Apache Spark and other libraries.

LosD:
Ouch! Fair play, then.

Issue #1794, opened 2018-07-31
Kafka datastore


kaspersorensen:
With the release of MetaModel 5.1.0, a new DataContext implementation is made available for querying Kafka streams. We should add a datastore in DataCleaner to take advantage of this functionality. That will enable us to be one of the few (only?) open source data quality tools that work with Apache Kafka!

davkrause:
I just now notice this issue is here so I want to tell you that I am working on this feature. Currently have a branch ("kafka") in my mirror of Datacleaner. With the new apache metamodel version out, this branch should be working soon so you can inspect it.

PR #1796, opened 2018-10-04
Kafka datastore

davkrause:
Reference to #1794: A new datastore based on Apache Kafka.

Notes and questions:

* Datastore is read only (based on capability from KafkaDataContext).
* For the datastore icon in GUI I download the official kafka logo. Is it ok?
* Changed the icon size in GUI for datastore icons because source image is not scaled.
* The "topics" field in the datastore dialog GUI is just a CSV field. Most of this class is just copied from other dialogs.
* Local mvn install: I ran into test issues in the engine/env/cluster module. I think this is local issue, so posting this PR to see if the CI build works. I'm seeing something related to "forked JVM crashed".


kaspersorensen:
Well the build broke on Travis CI too :-) But not with the env-cluster module. Here's your very minor omission:

```
[ERROR] /home/travis/build/datacleaner/DataCleaner/desktop/ui/src/main/java/org/datacleaner/windows/KafkaDatastoreDialog.java:144: Line is longer than 130 characters (found 132). [LineLength]
```

Other than that ... Looks very solid!

kaspersorensen:
I haven't tested this thoroughly myself, but I'm gonna say LGTM/+1 from my side. Don't know if anyone feels like testing it out a bit more (I only had a tiny self-built topic available to me to try it on). If not, let's merge in a few days.

kaspersorensen:
Thank you for the contribution @davkrause. I'm merging this now after testing it a bit more. It seems very solid.

PR #1798, opened 2019-01-04
Update hive-jdbc to version 2.3.4, addressing CVE-2018-1314

kaspersorensen:
See https://nvd.nist.gov/vuln/detail/CVE-2018-1314 for details on the vulnerability in the previous versions.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/util/ws/SimpleServiceSession.java
+++ b/engine/core/src/main/java/org/datacleaner/util/ws/SimpleServiceSession.java
@@ -22,8 +22,6 @@ package org.datacleaner.util.ws;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import javax.xml.ws.WebServiceException;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


LosD:
Might not really matter in the face of security issues like this, but how is the compatibility with other Hive versions? I remember previous Hive-JDBC versions being pretty terrible in that regard (though it's been a while, so I might remember wrong).

LosD:
Failures are due to banned dependencies

kaspersorensen:
Hmm I probably need to look closer into it then. We also have a MetaModel upgrade pending so I'll try and consolidate a dependency upgrades PR soon :-)

LosD:
A single banned dependency left:
> Found Banned Dependency: org.eclipse.jetty.orbit:javax.servlet:jar:3.0.0.v201112011016

LosD:
I remember trying to work with it earlier (I even made a PR for them for not following the JDBC spec, dumping not implemented exceptions where its not allowed. No one ever reacted to it, so I put in catches where it made MM fail), it's quite terrible.

I do believe that they also support other ways, so  you're probably right that looking into that is a good idea.

LosD:
Hmmmm, I might actually be thinking of Impala. Not sure. :-)

kaspersorensen:
Yaay, this one finally worked.

LosD:
That took a few tries... :-)

kaspersorensen:
OK to merge?

LosD:
LGTM, so sure!

PR #1803, opened 2019-02-11
Fixed issue where JTextField's value is null (iso. empty) on some JVMs

kaspersorensen:
Running on Java 11 I see null pointer exceptions when I search the schema tree. It seems that the `getText()` call on JTextField returns null sometimes. This PR fixes our handling of that situation.


LosD:
Kinda LGTM, but any idea why it happens, and if there's a risk of losing a legitimate value? Would suck if edits dissappeared sometimes.

kaspersorensen:
The way I understand it, we shouldn't be losing any real values because it just returns null when the text field is empty.

LosD:
That sounds fair. LGTM!

PR #1806, opened 2019-02-12
Fixes #1801: Created a "String contains filter"

### LINKED ISSUE #1801: String "contains" filter
I'm working with DC to filter out some noisy stuff, and I'm baffled that I cannot find a filter to ignore all records that contains (or doesn't contain) specific strings.

kaspersorensen:


kaspersorensen:
Hmm one of the builds failed in Travis because of log size. So I've also added some logging moderation logic. I think this logic is good anyhow, but I guess you could say it's unrelated to the string contains filter.

components/basic-filters/src/main/java/org/datacleaner/beans/filter/StringContainsFilter.java
```diff
@@ -0,0 +1,79 @@
...
+    @Configured(order = 3)
+    boolean caseSensitive = false;
+
+    @Override
+    public Category categorize(final InputRow inputRow) {
+        final String value = inputRow.getValue(column);
+        return categorize(value);
+    }
+
+    protected Category categorize(String value) {
```
arjansh: Why is this method protected and not just private? Do you expect this class to be extended and the method to be overridden?

Also I would like the `value` parameter to be final (which implies that some of the internal logic of the method has to be changed.)

components/basic-filters/src/main/java/org/datacleaner/beans/filter/StringContainsFilter.java
```diff
@@ -0,0 +1,79 @@
...
+        if (value == null) {
+            return Category.UNMATCHED;
+        }
+        if (!caseSensitive) {
+            value = value.toLowerCase();
+        }
+
+        for (String substring : substrings) {
```
arjansh: I would like the `substring` variable to be final, which also implies some of the logic inside the for loop has to be changed.

components/basic-filters/src/main/java/org/datacleaner/beans/filter/StringContainsFilter.java
kaspersorensen: I made it protected actually because I was thinking of testing it without having to make MockInputRows and such. But I'm gonna change it to private and test it via the public interface.

components/basic-filters/src/main/java/org/datacleaner/beans/filter/StringContainsFilter.java
```diff
@@ -0,0 +1,75 @@
...
+    public Category categorize(final InputRow inputRow) {
+        final String value = inputRow.getValue(column);
+        return categorize(value);
+    }
+
+    private Category categorize(final String value) {
+        if (value == null) {
+            return Category.UNMATCHED;
+        }
+        final String str = caseSensitive ? value : value.toLowerCase();
```
arjansh: Can you use a more descriptive name for this variable than an abbreviation?

kaspersorensen: renamed it to "normalizedValue"

arjansh:
LGTM! I will merge it when Travis agrees.

PR #1807, opened 2019-02-13
Issue #1596: Removed dependence on locale-specific DecimalFormatSymbols.

### LINKED ISSUE #1596: Pattern finder finds different patterns on different systems
Depending on what system (e.g. your computer, another computer, Travis CI) you run a job with a `Pattern finder` analyzer in it, its results may in practice vary.

E.g. when I run the sample `Customer age analysis` job, which is provided with DataCleaner, I run the job and I inspect the results of the Birthdate patterns, on my computer it finds three patterns: `####-##-##`, `####.##.##` and `a/a`. On someone else's computer it finds  `####-##-##`, `##########` and `a/a` instead.

I'm under the impression that this may be caused by the fact the the Pattern Finder uses the java.text.DecimalFormatSymbols class and instantiates it without explicitly passing a locale, as a result of which the default locale of the JVM is used, which may differ per system.

kaspersorensen:
Fixes #1596

### DIFF
--- a/components/pattern-finder/src/main/java/org/datacleaner/beans/stringpattern/PatternFinderAnalyzer.java
+++ b/components/pattern-finder/src/main/java/org/datacleaner/beans/stringpattern/PatternFinderAnalyzer.java
@@ -54,7 +54,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @Named("Pattern finder")
-@Description( "The Pattern Finder will inspect your String values and generate and match string patterns that suit "
+@Description("The Pattern Finder will inspect your String values and generate and match string patterns that suit "
         + "your data.\nIt can be used for a lot of purposes but is excellent for verifying or getting ideas about "
         + "the format of the string-values in a column.")
 @ExternalDocumentation(value = { @DocumentationLink(title = "Kasper's Source: Pattern Finder 2.0",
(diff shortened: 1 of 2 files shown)


components/pattern-finder/src/main/java/org/datacleaner/beans/stringpattern/TokenizerConfiguration.java
```diff
@@ -29,7 +28,7 @@
 public class TokenizerConfiguration implements Serializable {
 
     private static final long serialVersionUID = 1L;
-
+
```
arjansh: Why adding the whitespaces here?

components/pattern-finder/src/main/java/org/datacleaner/beans/stringpattern/PatternFinderAnalyzer.java
```diff
@@ -118,13 +118,14 @@
     String[] predefinedTokenPatterns;
 
     @Configured(required = false, order = 12)
-    Character decimalSeparator = DecimalFormatSymbols.getInstance().getDecimalSeparator();
+    Character decimalSeparator = '.';
```
arjansh: Isn't it better to store the '.', ',' and '-' in constants, taking into account that you reuse them in the `TokenizerConfiguration` class

components/pattern-finder/src/main/java/org/datacleaner/beans/stringpattern/TokenizerConfiguration.java
kaspersorensen: By mistake

components/pattern-finder/src/main/java/org/datacleaner/beans/stringpattern/PatternFinderAnalyzer.java
kaspersorensen: Added constants in TokenConfiguration

arjansh:
LGTM!

PR #1810, opened 2019-03-05
Machine learning components

kaspersorensen:
This PR adds some components for Machine Learning that I've been working on for a while over in my private repo: https://github.com/kaspersorensen/extension_machine_learning

Essentially the PR contains 3 new components (and I can imagine more to come under the "ML" umbrella):

* "Classifier training" analyzer which lets the user train and save a model.
* "Classifier cross-evaluation" analyzer which lets the user test a model against another data set.
* "Classifier transformer" which applies the model to individual records and returns the classification outcome.

In terms of model generation features, here's a summary:

* Currently there are (just) two algorithms: Random Forest and SVM. But it's pretty easy to add more.
* Feature generation can either be done in DataCleaner using existing transformers etc., but there is also a layer of feature generation within the column mapping to the analyzers. Specifically, you can choose to apply either of these strategies (see MLFeatureModifierType):
  * Direct (use the feature value as is)
  * Scaled (min-to-max value scaling from 0 to 1)
 * One hot vector encoding (create a feature per distinct value)
 * N-gram based vector encoding.

A few things that I'd like to improve, but aren't essential for a review:

- [x] Add icons
- [x] Add parameters to limit how many "One hot" or n-gram features to generate (for instance to avoid generating "One hot" features for completely distinct values or for n-grams that appear very rarely).
- [x] Add regression models and components


kaspersorensen:
Here's a few screenshots for reference:

A simple/silly example of trying to classify gender based on income and given name 3-gram:
image

And the results of the training:
image

kaspersorensen:
And here's a screenshot of the transformers output on a few records from the orderdb database:
image

desktop/api/src/main/java/org/datacleaner/util/IconUtils.java
```diff
@@ -98,8 +98,8 @@ private int brighter(final int rgb, final int percent) {
 
     private static final WidgetScreenResolutionAdjuster adjuster = WidgetScreenResolutionAdjuster.get();
     public static final int ICON_SIZE_LARGE = adjuster.adjust(32);
-    public static final int ICON_SIZE_MEDIUM = adjuster.adjust(22);
-    public static final int ICON_SIZE_SMALL = adjuster.adjust(16);
+    public static final int ICON_SIZE_MEDIUM = adjuster.adjust(24);
+    public static final int ICON_SIZE_SMALL = adjuster.adjust(18);
```
LosD: This seems unrelated, and a bit dangerous to just adjust all icon sizes like this?

kaspersorensen: Hmm yes I admit this one does not fit in... I'll revert.
The background is that the new icon for machine learning components is very hard to distinguish when overlayed with the "folder" icon. But maybe that just means that a better icon needs to be built.

kaspersorensen:
Made a lot of changes to this branch...

* Updated the icons :-)
* Instead of one big "Train classifier" analyzer, I broke it out per algorithm. This allows for much better configuration options that fit with the algorithm selected.
* Added a neural net algorithm.

kaspersorensen:
Also added support for regression based models, incl. an analyzer to train a random forest regressor, and a transformer to apply regressors.

kaspersorensen:
I've been using this branch quite a lot now and feel happy merging it.

Issue #1813, opened 2019-03-07
Create separate module(s) for datastore implementations


kaspersorensen:
Working with the `engine-core` module is becoming a bit painful IMO. It's very large and one of the things that it holds is all the concrete datastore implementations, which ideally it wouldn't even depend on. I think we should consider a new set of modules for datastores, similar to how our components are organized.

PR #1812, opened 2019-03-07
Dynamodb

kaspersorensen:
This makes DataCleaner capable of querying and writing to AWS DynamoDB.

Fixes #1811 

A screenshot for reference:

image

PR #1814, opened 2019-03-08
1813/datastore modules

kaspersorensen:
Extending (branching out of) the branch in #1812, this does the refactor described in #1813: Creating separate modules for the datastore implementations. This should take a bit of the module complexity out of `engine-core` and into it's own modules related ONLY to datastores.

To make it work I had to move a few utils (most notably the `ReflectionUtils` class) into the `engine-utils` module, but I think that is only appropriate anyway.


LosD:
LGTM, though I must admit I didn't look too deep. It's huge! :)

(And only in part due to the actual changes. Please keep your hands of the reformatting button, _especially_ for big PRs)

PR #1822, opened 2019-04-18
Added support for ARFF datastores (no UI elements though).

kaspersorensen:
I guess this is probably the first of two PRs ... First creating the code to support .arff datastores (support for which was recently added in Apache MetaModel). The obvious successor MR would add UI support.


kaspersorensen:
Merging based on lazy concensus :-)

Issue #1824, opened 2019-06-06
Unable to run data cleaner image on docker


isethy:
I have 5.1.5 data cleaner installed on my mac system. I am now trying to run the data cleaner image on docker but it is failing. I am following this link for doing this.

https://github.com/kaspersorensen/datacleaner-docker

Error Message:
hostname:~ userxyz$ docker run --rm kaspersor/datacleaner --usage
Unable to find image 'kaspersor/datacleaner:latest' locally
latest: Pulling from kaspersor/datacleaner
88286f41530e: Pull complete 
009f6e766a1b: Pull complete 
86ed68184682: Pull complete 
81f6e59b92a9: Pull complete 
6a25db359c8f: Pull complete 
Digest: sha256:243ca722931a1b9489798b3fdf6e2a220ef801080195f9c00109dee8d2ffcdae
Status: Downloaded newer image for kaspersor/datacleaner:latest
INFO  16:47:07 DataCleanerHome - Initializing DATACLEANER_HOME
INFO  16:47:07 DataCleanerHome - Running in standard mode.
Failed to load DataCleaner version from manifest: null
Failed to load DataCleaner version from manifest: null
INFO  16:47:07 DataCleanerHome - Attempting to build DATACLEANER_HOME in user.home: /root/.datacleaner/5.2.2 -> file:///root/.datacleaner/5.2.2
INFO  16:47:07 DataCleanerHome - Folder file:///root/.datacleaner/5.2.2 does not exist. Trying to create it.
INFO  16:47:07 DataCleanerHome - Folder file:///root/.datacleaner/5.2.2 created successfully. Attempting to build DATACLEANER_HOME here.
INFO  16:47:07 DataCleanerHomeUpgrader - Did not find a suitable upgrade candidate
Using default log configuration: jar:file:/opt/DataCleaner/DataCleaner.jar!/org/datacleaner/log4j-default.xml
 -conf (-configuration, --configuration-file) PATH          : Path to an XML file describing the configuration of
                                                              DataCleaner
 -ds (-datastore, --datastore-name) VAL                     : Name of datastore when printing a list of schemas, tables
                                                              or columns. Overrides datastore used when used with -job
 -job (--job-file) PATH                                     : Path to an analysis job XML file to execute
 -list [ANALYZERS | TRANSFORMERS | FILTERS | DATASTORES |   : Used to print a list of various elements available in the
 SCHEMAS | TABLES | COLUMNS]                                : configuration
[...]

PR #1825, opened 2019-06-08
Improved exception messages thrown by JdbcDatastore.

kaspersorensen:
See #1824 for discussion/reasoning behind this change.


LosD:
(I'll wait with merging until Travis is happy, but I don't see why it wouldn't be)

PR #1837, opened 2019-11-03
Upgraded commons-compress to fix CVE-2019-12402

kaspersorensen:


kaspersorensen:
Self-merging since I think it's pretty obvious we need it :-)

LosD:
I though dependabot handled these now? But that's maybe on a bit slower schedule? It's "only" a DOS, but yeah, good to fix ASAP.

LosD:
Or have I been seeing things? I'd swear I saw a lot of dependabot PRs, but now there's none?

Issue #1767, opened 2017-11-21
Bundle the old RegexSwap's regexes into the distribution


kaspersorensen:
Now that RegexSwap is no longer available, should we just put all those regexes into the application itself?

I've gone ahead and queried the regexes just to be able to preserve them for future use:

```
<?xml version="1.0" encoding="UTF-8"?>
<regexes>
	<regex>
		<name>Danish postal code</name>
		<expression>^(DK(-| )?)?[0-9]{4}$</expression>
		<description>This regex allows three different formats of danish
			postal codes. Examplified with the 2200 postal code:

			* DK-2200
			* DK2200
			* 2200
		</description>
		<positiveVotes>1</positiveVotes>
		<negativeVotes>0</negativeVotes>
		<categories>denmark,geographic,postal address</categories>
		<detailsUrl>https://datacleaner.org/ws/regex/Danish%20postal%20code
		</detailsUrl>
		<author>datacleaner.org</author>
		<timestamp>0</timestamp>
	</regex>
	<regex>
		<name>Danish phone number</name>
		<expression>^(\(?\Q+45\E\)?( )?)?[0-9]{8}$</expression>
		<description>This regex is useful for validating Danish (8 digit)
			telephone numbers that may or may not be prefixed by +45 or (+45) and
			an optional space.</description>
		<positiveVotes>1</positiveVotes>
		<negativeVotes>0</negativeVotes>
		<categories>denmark,geographic,phone</categories>
		<detailsUrl>https://datacleaner.org/ws/regex/Danish%20phone%20number
		</detailsUrl>
		<author>datacleaner.org</author>
		<timestamp>0</timestamp>
	</regex>
	<regex>
		<name>Website URL</name>
		<expression>^https?://[a-z0-9_-][\.[a-z0-9_-]]*\.(com|edu|org|net|int|info|eu|biz|mil|gov|aero|travel|pro|name|museum|coop|asia|[a-z][a-z])+(:[0-9]+)?[/[a-zA-Z0-9\._#-]]*/?$
		</expression>
		<description>Validates a HTTP or HTTPS url for a website. It works
			well for most uses, but there are a few corner-cases that it cannot
			handle.

			Known issues:
			* URL "GET" parameters are not supported.
			* Top-level domains are not literally validated.
		</description>
		<positiveVotes>0</positiveVotes>
		<negativeVotes>0</negativeVotes>
		<categories>internet</categories>
[...]

PR #1896, opened 2022-01-02
Feature #1767 - RegexSwap client reimplementation based on JSON endpoint

kaspersorensen:
As described and discussed in #1767 - this changes the RegexSwapClient class to use https://datacleaner.github.io/content/regexes.json for fetching regexes instead of the decomissioned old XML endpoints.

### DIFF
--- a/engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
+++ b/engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
@@ -34,6 +34,7 @@ import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.client.methods.RequestBuilder;
 import org.apache.http.util.EntityUtils;
+import org.datacleaner.util.SystemProperties;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 
@@ -42,7 +43,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  */
 public final class RegexSwapClient {
 
-    private static final URI REGEXES_URI = URI.create("https://datacleaner.github.io/content/regexes.json");
+    private static final URI REGEXES_URI = URI.create(SystemProperties.getString("org.datacleaner.regexswap.url",
+            "https://datacleaner.github.io/content/regexes.json"));
 
     private final Map<String, Category> _categories = new HashMap<>();
     private final Map<String, Regex> _regexes = new HashMap<>();
--- a/engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
+++ b/engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
@@ -38,82 +38,80 @@ import org.apache.http.util.EntityUtils;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
- * Client class for the RegexSwap, which allows for easy retrieval of shared
- * regular expressions.
+ * Client class for the RegexSwap, which allows for easy retrieval of shared regular expressions.
  */
 public final class RegexSwapClient {
 
-	private static final URI REGEXES_URI = URI.create("https://datacleaner.github.io/content/regexes.json");
+    private static final URI REGEXES_URI = URI.create("https://datacleaner.github.io/content/regexes.json");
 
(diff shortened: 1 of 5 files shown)


engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
```diff
@@ -19,150 +19,101 @@
...
+	public void refreshRegexes() {
+		final HttpUriRequest request = RequestBuilder.get(REGEXES_URI).build();
+		final Map<String, ?> root;
+		try {
+			final HttpResponse response = _httpClient.execute(request);
+			final String json = EntityUtils.toString(response.getEntity());
+			final ObjectMapper objectMapper = new ObjectMapper();
+			root = objectMapper.readValue(json, Map.class);
+		} catch (IOException e) {
+			throw new UncheckedIOException(e);
```
LosD, line 80: Shouldn't this warn the user, or is it handled somewhere else?

engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
```diff
@@ -19,150 +19,101 @@
...
-                }
-            }
-
-            _regexes.put(name, regex);
-
-            regexes.add(regex);
-        }
-        return regexes;
-    }
+	private static final URI REGEXES_URI = URI.create("https://datacleaner.github.io/content/regexes.json");
```
LosD, line 46: Leaving the URL as a hardcoded value may leave us open to a similar issue again later.

LosD:
Code itself looks mostly good to me, but I see some changes from spaces to tabs in `RegexSwapClient.java` and `Regex.java`. If I remember correctly, the code style is spaces.

kaspersorensen:
> Code itself looks mostly good to me, but I see some changes from spaces to tabs in `RegexSwapClient.java` and `Regex.java`. If I remember correctly, the code style is spaces.

OMG yes. This is what happens when you (I) don't code for a long time and go rogue with a fresh Eclipse installation :D Thanks for the feedback notes!

engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
kaspersorensen: Hmm yes but I find it hard to imagine a situation where we wouldn't have to pull a release to fix it anyway?
Is your suggestion that we use an environment variable or something like that?

LosD: Yeah, just some kind of fallback. I guess an environment variable that overrides the default would be perfect for that.

Just so that someone who doesn't want to update for compatibility or whatever reason (or who wants a custom set of regexes hosted locally), can update the URL.

But its no biggie.

engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
LosD: I didn't really mean "more specific". I meant more like a proper "we couldn't contact the server" local warning instead of just dumping it up the chain.

But I might misremember how we handle exceptions and this is all handled properly somewhere (though "someone will probably handle it" seems like a bad strategy).

LosD (commented):
No real blockers, though I'm not sure I agree with the exception strategy. But if I remember correctly, this is just using the new-ish Oracle approved way to strip checked exceptions that we (unfortunately) already did in several places, so I guess it's not really a step backwards.

engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
kaspersorensen: I see what you mean, but since this is a engine/backend component I think that throwing an exception is most appropriate. We could add more logging and/or make a specific exception for this scenario. But then again I do kinda think this is also just about wrapping the checked exception in an unchecked exception and deciding not to try and handle it, because we have no good way to recover from it here.

engine/core/src/main/java/org/datacleaner/reference/regexswap/RegexSwapClient.java
kaspersorensen: Good point, fixed.

